PR #22, opened 2017-10-27
adding synthetics monitoring

bridgetlane:


bridgetlane:
Tried a first pass at adding synthetics monitoring #6

tuannvm:
This is great! any timeline that we could get this merged?

vendor/vendor.json
```diff
@@ -226,12 +226,30 @@
 			"revision": "2ee87856327ba09384cabd113bc6b5d174e9ec0f",
 			"revisionTime": "2017-07-27T06:48:18Z"
 		},
+		{
+			"checksumSHA1": "jZpG8JA8KXXPB0ITK6ZWq1HUJNk=",
+			"path": "github.com/blinohod/terraform/builtin/providers/newrelic",
```
paultyng: Something is weird here, not sure why this is in the vendor on the PR

bridgetlane: I'm not sure on why that one is there, I'll remove it

smithclay:
Hi Bridget, new maintainer here. Learning this code/guidelines for resources myself as I'm working on adding dashboard API support. 

Could you also add a new documentation page for the new `newrelic_synthetics_alert_condition` resource in this PR?  (examples in the `/website/docs/r` directory, tried adding one myself today in this commit: https://github.com/terraform-providers/terraform-provider-newrelic/pull/26/commits/377119c73cb2b594b09c911375d4101ee5be6535)

bridgetlane:
@paultyng @smithclay Added docs, and I believe I fixed up any vendor problems for this one

newrelic/data_source_newrelic_synthetics_monitor.go
```diff
@@ -0,0 +1,83 @@
...
+		return err
+	}
+
+	if monitor == nil {
+		return fmt.Errorf("The name '%s' does not match any New Relic monitors.", name)
+	}
+
+	d.SetId(monitor.ID)
+	d.Set("name", monitor.Name)
+	d.Set("monitor_id", monitor.ID)
```
paultyng: This needs to be defined in the schema.

newrelic/data_source_newrelic_synthetics_monitor.go
```diff
@@ -0,0 +1,83 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"api_key": {
+				Type:        schema.TypeString,
+				Required:    true,
+				DefaultFunc: schema.EnvDefaultFunc("NEWRELIC_API_KEY", nil),
+				Sensitive:   true,
+			},
+			"max_check": {
```
paultyng: Is there a reason to limit how many it queries? Why not just query them all? If the sorting isn't deterministic this could be problematic as it would result in erroneous diffs.

newrelic/data_source_newrelic_synthetics_monitor.go
```diff
@@ -0,0 +1,83 @@
...
+				Optional: true,
+				Default:  1000,
+			},
+		},
+	}
+}
+
+func dataSourceNewRelicSyntheticsMonitorRead(d *schema.ResourceData, meta interface{}) error {
+
+	apiKey := d.Get("api_key").(string)
```
paultyng: Typically you would set this up in the provider.go file. You can have more than one client on the meta, that is commonly do in other providers, especially since this `api_key` is already defined at the provider level.

newrelic/data_source_newrelic_synthetics_monitor.go
bridgetlane: @paultyng You want people to provide the monitor id? Doesnt it make more sense to get that programmatically?

newrelic/data_source_newrelic_synthetics_monitor.go
bridgetlane: There was a reason for this but I dont recall what it was at this point, I'll look into this again

newrelic/data_source_newrelic_synthetics_monitor.go
paultyng: Sorry, should have been more clear, in the schema you can specify `Computed` (and not `Optional` or `Required`) for data that is only output.

paultyng: https://github.com/hashicorp/terraform/blob/0cc9e050ecd4a46ba6448758c2edc0b29bef5695/helper/schema/schema.go#L109-L110

paultyng: https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed

(these are some new docs we launched recently!)

bridgetlane: Ahh thanks that makes sense! Will do

newrelic/data_source_newrelic_synthetics_monitor.go
paultyng: This is what I mean by putting multiple clients on the meta object, unfortunately its a bit of a refactor: https://github.com/terraform-providers/terraform-provider-azurerm/blob/91e4562440f9a8a1bb5ea58947e28dd1134d9e63/azurerm/config.go#L73-L76

```go
type providerClient struct {
  Client *newrelic.Client
  Synthetics *synthetics.Client
}
```

You basically just create a new client struct that would have both this existing client plus the new synthetics client to use as the meta in your resource methods. Unfortunately it means you have to adjust the [type assertions](https://github.com/terraform-providers/terraform-provider-newrelic/blob/744ea7cf38878334670a690094ac6e0a278d771c/newrelic/resource_newrelic_alert_channel.go#L107) to be the new struct you create.

But this way you can use the provider config to setup both clients at the same time.

smithclay: The recent changes merged in v`1.0.1` introduced a `ProviderClient` as above.

bridgetlane:
Hey @paultyng @smithclay sorry for the delay on this. I think these are all the changes you guys asked for, and it looks a lot cleaner now that I've pulled the api_key/synthetics client creation into the provider. Please take another look and let me know if anything else is needed 😄

newrelic/data_source_newrelic_synthetics_monitor.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	synthetics "github.com/dollarshaveclub/new-relic-synthetics-go"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceNewRelicSyntheticsMonitor() *schema.Resource {
```
smithclay: Please add a docs page + link to sidebar for the new data source type (see: https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/website/docs/d/application.html.markdown for an example)

newrelic/resource_newrelic_synthetics_alert_condition.go
```diff
@@ -0,0 +1,168 @@
+package newrelic
+
+import (
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	newrelic "github.com/paultyng/go-newrelic/api"
+)
+
+func resourceNewRelicSyntheticsAlertCondition() *schema.Resource {
```
smithclay: Please add a test for the new resource. See https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/newrelic/resource_newrelic_nrql_alert_condition_test.go for a similar example. Thanks!

bridgetlane:
@paultyng @smithclay please give this another look - thank you!

oba11:
Will be great if this can be merged, well done @bridgetlane

RSaundersonDeloitte:
*Very* interested in this PR.  Once it's been approved and merged, how quickly will it show up in the version of the provider which is packaged in terraform itself?

Constantin07:
When this PR will be merged ?

bridgetlane:
@paultyng @smithclay any way I could get a review on this?

paultyng:
LGTM, I think there may be some redundancy in the data source between `id` and `monitor_id`, and you could potentially remove `monitor_id` as its not in the examples, but not a requirement.

PR #54, opened 2018-10-26
Switch to 1.11 and Go modules

paultyng:


paultyng:
@bridgetlane is this cool with you?

bridgetlane:
@paultyng yup looks great!

PR #70, opened 2018-12-12
#3 Make resource_newrelic_alert_condition enabled status configurable

sgsiebers:
This PR is depends on changes from https://github.com/paultyng/go-newrelic/pull/27 which has been merged, but not released and pulled into this project.  I've tested by manually pulling in the change and it works as expected.  Without the dependent change, you can't update a condition to switch from enabled to disabled.

@bridgetlane, I may need help from you and/or @paultyng to release and pull in the dependent change in the go-newrelic module. 

As a side note, your README is out of date and still references the use of govendor, but it seems you've switched over to go modules, so that might need an update.

### DIFF
--- a/newrelic/resource_newrelic_alert_condition.go
+++ b/newrelic/resource_newrelic_alert_condition.go
@@ -92,6 +92,11 @@ func resourceNewRelicAlertCondition() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
 			"type": {
 				Type:         schema.TypeString,
 				Required:     true,
@@ -200,7 +205,7 @@ func buildAlertConditionStruct(d *schema.ResourceData) *newrelic.AlertCondition 
 	condition := newrelic.AlertCondition{
 		Type:                d.Get("type").(string),
 		Name:                d.Get("name").(string),
-		Enabled:             true,
+		Enabled:             d.Get("enabled").(bool),
 		Entities:            entities,
 		Metric:              d.Get("metric").(string),
 		Terms:               terms,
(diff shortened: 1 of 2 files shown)


AshleyPoole:
Can this be merged now? I'd like to see this feature 👍

paultyng:
I cut a release of go-newrelic, should be able to update this PR now. If you aren't able to I can probably get to the dependency update later this week.

sgsiebers:
Thanks @paultyng!  It looks like there's already a PR out for the dependency bump.
https://github.com/terraform-providers/terraform-provider-newrelic/pull/78
As soon as that's closed, this should be ready to merge too.

paultyng:
I merged the PR (along with some other dependency updates), I did notice some acceptance test failures in it, not sure if that is something new or what, but haven't had time to dig in yet.

sgsiebers:
I fetched latest updates from master and rebased them onto this branch locally just to see what would happen.  I get the exact same test failures in master as when running on this branch, so whatever is breaking the tests is already in master and not introduced here.  Still, I understand if you want to be cautious and figure them out first before merging this.

paultyng:
Go ahead and push your rebase here and I think we can run with it. If this isn't introducing new failures, probably worth getting it in.

sgsiebers:
Rebase pushed.  The acceptance test failure I'm getting is in the synthetics monitor data resource which is unrelated to this PR:

FAIL: TestAccNewRelicSyntheticsAlertCondition_Basic (0.45s)
    testing.go:538: Step 0 error: Error refreshing: 1 error occurred:
                * data.newrelic_synthetics_monitor.bar: 1 error occurred:
                * data.newrelic_synthetics_monitor.bar: data.newrelic_synthetics_monitor.bar: The name 'tf_test_qrkeqrdhy7' does not match any New Relic monitors.
        

As far as I can tell, there's nothing in the test actually creating a test synthetics monitor to read into a data resource.  Seem like that could be fixed using new capabilities in https://github.com/terraform-providers/terraform-provider-newrelic/pull/67 to create a synthetics monitor resource then read it with the data resource.

sgsiebers:
Ok, branch should be in sync with upstream/master now...test failures on the branch are the same as observed in master.  Ready to merge when you are comfortable, or let me know what else you require.

PR #79, opened 2019-02-07
add alert condition name validation

dstokes:
The New Relic API enforces a 64 character limit on alert condition
names as described in [the documentation](https://docs.newrelic.com/docs/alerts/new-relic-alerts/defining-conditions/define-alert-conditions#rename-condition). This pull request validates the alert condition name before making a request to the API.


sgsiebers:
Alert policy names are similarly restricted in length, care to make a similar validation change there?
https://docs.newrelic.com/docs/alerts/new-relic-alerts/configuring-alert-policies/create-or-rename-alert-policy

Issue #66, opened 2018-11-14
Wrong Type used in newrelic_alert_channel webhook configuration


oba11:
Hi there,

Noticed that a wrong type was used for keys `payload` and `headers` in the web hook configuration

Got an error that payload and headers should be string which is not so according to the [API documentation](https://docs.newrelic.com/docs/alerts/rest-api-alerts/new-relic-alerts-rest-api/rest-api-calls-new-relic-alerts#channels-list) with example below

```
{
  "base_url": "http://test.com",
  "auth_username": "username",
  "auth_password": "password",
  "payload_type": "application/json",
  "payload": {"account_id": 1, "account_name": "account name" },
  "headers": {"header1": "test", "header2": "test"}
}
```

### Terraform Version
v0.11.10

### Affected Resource(s)
Please list the resources as a list, for example:
- newrelic_alert_channel

### Debug Output
```
2018-11-14T12:54:29.861+0400 [DEBUG] plugin.terraform-provider-newrelic_v1.2.0_x4: 2018/11/14 12:54:29 [INFO] Reading New Relic alert channel 2024412
2018/11/14 12:54:32 [ERROR] root: eval: *terraform.EvalRefresh, err: newrelic_alert_channel.this: [DEBUG] Error setting Alert Channel Configuration: &errors.errorString{s:"configuration.payload: '' expected type 'string', got unconvertible type 'map[string]interface {}'"}
2018/11/14 12:54:32 [ERROR] root: eval: *terraform.EvalSequence, err: newrelic_alert_channel.this: [DEBUG] Error setting Alert Channel Configuration: &errors.errorString{s:"configuration.payload: '' expected type 'string', got unconvertible type 'map[string]interface {}'"}
2018/11/14 12:54:32 [TRACE] [walkRefresh] Exiting eval tree: newrelic_alert_channel.this

2018/11/14 12:54:32 [DEBUG] plugin: waiting for all plugin processes to complete...
Error: Error refreshing state: 1 error(s) occurred:

2018-11-14T12:54:32.015+0400 [DEBUG] plugin.terraform-provider-newrelic_v1.2.0_x4: 2018/11/14 12:54:32 [ERR] plugin: plugin server: accept unix /var/folders/mr/6gwkzzjs4nldgm7gf2dbj6yw0000gn/T/plugin499091540: use of closed network connection
[...]

Issue #45, opened 2018-09-19
Unable to specify payload for webhook in json


sharpjl:
### Terraform Version
Terraform v0.10.8

### Affected Resource(s)

- newrelic_alert_channel with type of webhook

### Terraform Configuration Files
```hcl
resource "newrelic_alert_channel" "webhooktest" {
  name = "webhooktest"
  type = "webhook"

  configuration = {
    payload_type    = "application/json"
    auth_username   = "my_user",
    auth_password   = "my_password",
    base_url        = "thirdpartyurl",
    payload         = {"priority": 3, "status": 2, "description": "New Relic Notification"}
  }
}
```
### Expected Behavior
Create alert channel with payload in json format.

### Actual Behavior
Error: Error parsing main.tf: At 28:34: illegal char

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

bridgetlane:
@sharpjl What line is line 28? Can you either direct me to the line or include your entire .tf file so that I can try to reproduce?

ctrombley:
Duplicate of #66

ctrombley:
Please continue future discussion in #66.

Issue #18, opened 2017-10-04
Unable to set custom headers for webhook alert channel


mkubenka:
### Terraform Version

Terraform v0.10.6

### Affected Resource(s)
- newrelic_alert_channel

### Terraform Configuration Files
```hcl
resource "newrelic_alert_channel" "api" {
  name = "API"
  type = "webhook"

  configuration = {
    base_url = "${var.url}"

    headers = {
      x-api-key = "${var.api_key}"
    }
  }
}
```

### Expected Behavior
Create alert channel with custom header `x-api-key`.

### Actual Behavior
```
1 error(s) occurred:

* newrelic_alert_channel.api: configuration (headers): '' expected type 'string', got unconvertible type '[]map[string]interface {}'
```

### Steps to Reproduce
1. `terraform apply`

ctrombley:
Duplicate of #66.

ctrombley:
Duplicate of #66

ctrombley:
Please continue future discussion in #66.

PR #81, opened 2019-02-11
Modify alert_channel to use maps for headers, payload

aaronshaver:
A possible solution for #66 - this change:
- un/marshalls headers and payload into separate maps
- tests various configurations of these optional fields
- documents these fields with examples

### DIFF
--- a/newrelic/resource_newrelic_alert_channel.go
+++ b/newrelic/resource_newrelic_alert_channel.go
@@ -50,9 +50,7 @@ var alertChannelTypes = map[string][]string{
 		"auth_type",
 		"auth_username",
 		"base_url",
-		"headers",
 		"payload_type",
-		"payload",
 	},
 }
 
@@ -89,7 +87,33 @@ func resourceNewRelicAlertChannel() *schema.Resource {
 				//TODO: ValidateFunc: (use list of keys from map above)
 				Sensitive: true,
 			},
+			"headers": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				ForceNew: true,
+				Sensitive: true,
+			},
+			"payload": {
+				Type:         schema.TypeMap,
+				Optional:     true,
+				ForceNew:     true,
+				Sensitive:    true,
+				ValidateFunc: configBlockLengthGreaterThan(0),
+			},
 		},
+	}
+}
+
+// This function verifies that the number of keys within a configuration
+// map is greater than the provided parameter.
+func configBlockLengthGreaterThan(minLength int) schema.SchemaValidateFunc {
+	return func(i interface{}, st string) (s []string, errorSlice []error) {
+		length := len(i.(map[string]interface{}))
+		if length > minLength {
+			return
+		}
+		errorSlice = append(errorSlice, fmt.Errorf("expected %s not to be empty", st))
+		return
 	}
 }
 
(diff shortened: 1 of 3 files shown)


t0astbandit:
@paultyng when you have some time would you please review this PR?

newrelic/resource_newrelic_alert_channel.go
```diff
@@ -141,6 +173,19 @@ func resourceNewRelicAlertChannelRead(d *schema.ResourceData, meta interface{})
 
 	d.Set("name", channel.Name)
 	d.Set("type", channel.Type)
+
+	// extract headers from Configuration before we try and set it in the resource
+	if headers, ok := channel.Configuration["headers"]; ok {
+		d.Set("headers", headers)
+		delete(channel.Configuration, "headers")
```
paultyng: I'm a little hesitant to do this `delete` as its a breaking change for anyone already using this behavior.

newrelic/resource_newrelic_alert_channel.go
```diff
@@ -141,6 +173,19 @@ func resourceNewRelicAlertChannelRead(d *schema.ResourceData, meta interface{})
...
+	}
+
+	// extract payload from Configuration before we try and set it in the resource
+	if payload, ok := channel.Configuration["payload"]; ok {
+		d.Set("payload", payload)
+		delete(channel.Configuration, "payload")
```
paultyng: I'm a little hesitant to do this `delete` as its a breaking change for anyone already using this behavior.

newrelic/resource_newrelic_alert_channel.go
nathanael-ness: I do not believe that the `delete` could break anything because, as far as I can tell setting the headers or payload is broken currently

brasey: I think nathanael-ness is right, I don't see how either `headers` or `payload` could be working currently with the type mismatch.

We're having to use a null-resource to deploy a webhook alert channel, can't use the resource as it is.

brasey:
Also related: #45

brasey:
And #18

StephenWeber:
Having reviewed the associated issues and thinking about how the code operates, I wanted to re-summarize the design. For most of it I'll talk about `payload` but all commentary is applicable to `headers` as well.

This PR adds the ability to specify the payload and headers as maps. This PR's proposed HCL example:

	alert_channel mychannel {

	    configuration {
	    	... other things ...
	    }
	
	    headers = {
	    	api-key = "my-internal-api-key"
	    }
	    payload = {
	    	account_id = "test"
	    }

	}

The [comment](https://github.com/terraform-providers/terraform-provider-newrelic/pull/81#discussion_r260264600) raised by @paultyng seems like people may be doing this already to get through the API:

	alert_channel {

	    configuration {
	    	headers = "{ 'header': 'value' }"
	    	payload = "account_id=test"
	    }

	}

It's possible some users with simple headers are providing an explicit string in `alert_channel.configuration.payload`. If so, we shouldn't just clobber those in favor of the map-style. Reconciling these two methods raises two situations:

1. User specified payload (or header) in one place, there is no conflict. We use whichever field they provided (`alert_channel.configuration.payload` or `alert_channel.payload`) -  we should support both methods and do the right thing.
2. User specifies payload in _both_ `alert_channel.configuration.payload` or `alert_channel.payload` and there's a conflict


We should catch the situation where the user mistakenly sets both fields, that may be in conflict, and issue a warning that helps clarify what they should do. Perhaps we change the field names to be more clear about this, such as `payload_map`, and still set it in the API's config as we are here.

So I think the proposal for updating this PR is:

1. changing the (optional) map field to be named `payload_map`
2. checking if a string is set at `alert_channel.configuration.payload` and using that
3. checking if both fields are specified and issuing a warning about the conflict
4. probably fixing the lines that delete payload from the configuration block to respect which place in HCL the payload is stored 
5. updating / adding tests to cover the new design
6. updating documentation and examples to reflect config options

Anything I've missed or that should also be considered?

beardface:
This issue has been open for a while now, does anyone have a work around?

ctrombley:
Related to paultyng/go-newrelic#52

ctrombley:
The terraform docs [here](https://www.terraform.io/docs/extend/best-practices/deprecations.html) explain how to use the `ConflictsWith` schema attribute to implement the either/or checking @StephenWeber describes above.

ctrombley:
Closing this in favor of the implementation in #307.

PR #122, opened 2019-03-12
Clean up dependencies & finish go modules migration

radeksimko:
This is a result of `go mod tidy` and `go mod vendor` which makes the provider to compile under strict `-mod=vendor` mode in 1.11:

```
go test -i $(go list ./... |grep -v 'vendor') || exit 1
# github.com/aws/aws-sdk-go/aws/session
vendor/github.com/aws/aws-sdk-go/aws/session/session.go:413:7: undefined: getCABundleTransport
make: *** [test] Error 1
```

I'm also fixing Travis & Makefile (aligning with most other providers which were migrated) so we can notice these problems earlier.


appilon:
@radeksimko don't know how this got buried, LGTM I'll re-run the tests

appilon:
no new failures in TC

Issue #131, opened 2019-04-30
Support for newrelic_nrql_ alert_condition "since_value" values greater than 5


timonwalkley:
newrelic_nrql_ alert_condition validation for the "since_value" attribute accepts values between 1-5. This attribute (now) supports values up to 20.

This attribute controls the evaluation offset of the nrql alert condition.

### Terraform Version
Terraform v0.11.13
provider.newrelic v1.5.0

### Affected Resource(s)
- newrelic_nrql_ alert_condition

### Terraform Configuration Files
An example of a resource where this happens:

```
resource "newrelic_nrql_alert_condition" "test" {
  policy_id   = "${newrelic_alert_policy.warning_policy.id}"
  name        = "test"

  term {
    duration      = 5
    threshold     = 0
    priority      = "critical"
    operator      = "above"
    time_function = "any"
  }

  nrql {
    query       = "SELECT sum(provider.errors.Maximum) from ServerlessSample FACET provider.functionName"
    since_value = "15"
  }
}
```

### Expected Behavior
Terraform Plan should succeed

### Actual Behavior
Terraform Plan fails:
newrelic_nrql_alert_condition.lambda_error_warning: expected nrql.0.since_value to be one of [1 2 3 4 5], got 15

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- GH-1234

timonwalkley:
Can we just remove the validation? Happy to raise a PR.

PR #136, opened 2019-05-10
Relax NRQL condition limitations

monkeyherder:
NRQL condition limitations for `duration` and `since_value` terms are previously overly strict and limiting out values that were perfectly acceptable for the API.

`duration` is previously limiting to one of 1, 2, 3, 4, 5, 10, 15, 30, 60, or 120, but the API has no such restriction and only requires that it is a positive integer

`since_value` is limited to "1", "2", "3", "4",  or "5", but the API allows any value between 1 and 20.

This PR relaxes limitations on the term validation to more accurately reflect the limitations imposed by the API.

### DIFF
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -52,7 +52,7 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 						"since_value": {
 							Type:         schema.TypeString,
 							Required:     true,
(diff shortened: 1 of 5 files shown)


ianrayns:
Awesome, this addresses #131

paultyng:
`since_value` was addressed in #144, didn't see this PR prior to merging, so going to modify to just do the other attribute.

ctrombley:
This has been addressed in other PRs (#144, #178).

PR #170, opened 2019-09-10
Updated Readme

mcough2:
updated the readme and a few comments to make the provider easier to use


README.md
```diff
@@ -32,7 +32,10 @@ $ make build
 
 Using the provider
 ----------------------
-__TODO__: Fill in for each provider
+
+If you're building the provider, follow the instructions to [install it as a plugin.](https://www.terraform.io/docs/plugins/basics.html#installing-a-plugin) After placing it into your plugins directory,  run `terraform init` to initialize it.
```
binaryseed: Is this part aimed at folks developing the provider? I don't think that regular users will need anything other than `terraform init` since it's an official provider

README.md
```diff
@@ -72,3 +75,11 @@ To update a dependency:
 1. Ensure you have govendor installed: `go get -u github.com/kardianos/govendor`
 2. CD to the root of this repo (not into the vendor directory)
 3. Run `govendor fetch PATH/FOR/PACKAGE` where PATH/FOR/PACKAGE is the path property in `vendor.json` of the package you wish to update. Ex. `go get github.com/paultyng/go-newrelic/api`
+
+Finding Certain ID and Key
```
binaryseed: This section feels like it's content that should be part of the resource-specific docs that get rendered out at https://www.terraform.io/docs/providers/newrelic/index.html

sanderblue: Agreed. Thoughts on adding a hyperlink to this page here to help with discovery?

README.md
sanderblue: Agreed this feels like something that would occur during development of the provider itself. Should this be moved to the `Developing the Provider` section?

sanderblue:
Thank you for the contribution. We've addressed quite a few things in the README recently and more updates are on the way.

PR #178, opened 2019-10-17
fix nrql alert condition duration

jamesgoodhouse:
Acceptable values range from 2-120 (inclusive).

### DIFF
--- a/website/docs/r/nrql_alert_condition.html.markdown
+++ b/website/docs/r/nrql_alert_condition.html.markdown
@@ -76,8 +76,8 @@ The following attributes are exported:
 
 ## Import
 
-Alert conditions can be imported using the `id`, e.g.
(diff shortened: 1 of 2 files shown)


website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -76,8 +76,8 @@ The following attributes are exported:
 
 ## Import
 
-Alert conditions can be imported using the `id`, e.g.
+Alert conditions can be imported using the a composite ID of `<policy_id>:<condition_id> `id`, e.g.
```
coreyarnold: looks like you'll want to remove the leftover "the" here.

jamesgoodhouse: good catch. fixed!

PR #186, opened 2019-11-01
Destroy test apps after an acceptance test run

ctrombley:
This PR cleans up the fake applications created after a run of the acceptance tests.

It also adds Go 1.12.x and 1.13.x to the Travis build and resolves some linting issues.


ctrombley:
This approach won't work, since New Relic applications cannot be deleted for some period after they have been created.

PR #190, opened 2019-11-06
chore: update CHANGELOG and README

sanderblue:
Release notes for `1.6.0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,24 +11,6 @@ Requirements
 ------------
 
 -	[Terraform](https://www.terraform.io/downloads.html) 0.10.x
--	[Go](https://golang.org/doc/install) 1.13 (to build the provider plugin)
-
-Building The Provider
----------------------
-
-Clone repository to: `$GOPATH/src/github.com/terraform-providers/terraform-provider-newrelic`
-
-```sh
-$ mkdir -p $GOPATH/src/github.com/terraform-providers; cd $GOPATH/src/github.com/terraform-providers
-$ git clone git@github.com:terraform-providers/terraform-provider-newrelic.git
-```
-
-Enter the provider directory and build the provider
-
-```sh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,7 +11,7 @@ Requirements
 ------------
 
 -	[Terraform](https://www.terraform.io/downloads.html) 0.10.x
--	[Go](https://golang.org/doc/install) 1.8 (to build the provider plugin)
+-	[Go](https://golang.org/doc/install) 1.13 (to build the provider plugin)
```
ctrombley: I wouldn't mind getting rid of this line and keeping everything above the fold targeted toward users (not developers). We specify the 1.13 requirement down below.

README.md
```diff
@@ -11,7 +11,7 @@ Requirements
...
 
 Building The Provider
```
ctrombley: Likewise here - can we move this whole section down below "Using the provider"?  Maybe move it into the build section below?

README.md
```diff
@@ -50,27 +53,28 @@ $ $GOPATH/bin/terraform-provider-newrelic
...
+In order to test the provider, run `make test`. This will run the unit test suite.
 
 ```sh
 $ make test
 ```
 
 In order to run the full suite of Acceptance tests, run `make testacc`.
 
-*Note:* Acceptance tests create real resources, and often cost money to run.
+*Note:* Acceptance tests *create real resources*, and often cost money to run. The environment variable `NEWRELIC_API_KEY` must also be set with your key for acceptance tests to work properly.
```
ctrombley: `NEWRELIC_LICENSE_KEY` is also currently needed for acc testing.

README.md
sanderblue: Yeah, I can understand this change. I kind of pondered that a bit earlier as well.

README.md
sanderblue: Ah, right. Thanks.

Issue #165, opened 2019-08-29
Data source for newrelic_alert_policy does not match names strictly enough


pickgr:
### Terraform Version
Terraform v0.11.11
provider.newrelic v1.5.0

### Affected Resource(s)
data.newrelic_alert_policy.ps_notify_pagerduty

provider.newrelic v1.5.0
### Terraform Configuration Files
```hcl
data "newrelic_alert_policy" "ops_notify_pagerduty" {
  name = "ops-notify-pagerduty"
}

data "newrelic_alert_policy" "ps_notify_pagerduty" {
  name = "ps-notify-pagerduty"
}
```
.

### Expected Behavior
Each data source should return a unique id.

### Actual Behavior
The id for ops-notify-pagerduty is returned for data.newrelic_alert_policy.ps_notify_pagerduty vs. ps-notify-pagerduty

### References
This issue is where the names are iterated and the match is based on the string containing a subset.  ie: in this line

https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/newrelic/data_source_newrelic_alert_policy.go#L52

there needs to be a stricter match (Use Compare vs. Contains)

jthurman42:
Resolved in #197

PR #197, opened 2019-11-10
fix issue #165 with strict name matching for alert policies

nvaller:
A possible fix for #165 .

- Implements a stricter matching on alert policy data sources.


ctrombley:
This PR addresses the same issue for the alert channels data source as well.

PR #208, opened 2019-11-13
chore: update CHANGELOG - release 1.7.0

sanderblue:


ctrombley:
Let's see if we can get #206 merged and bump the provider to v1.7.0, since we already have a minor release ready to go.

PR #225, opened 2019-11-18
docs: add Getting Started section

sanderblue:


website/docs/guides/getting_started.html.markdown
```diff
@@ -0,0 +1,176 @@
...
+---
+
+# Getting Started with the New Relic Provider
+
+## Before You Begin
+
+* This guide assumes you already have a New Relic agent deployed. If you don't have New Relic integrated yet, check out [New Relic's introduction documentation](https://docs.newrelic.com/docs/using-new-relic/welcome-new-relic/get-started/introduction-new-relic) to get started there, then head back over here to get started with the New Relic Terraform Provider.
+* [Install Terraform](https://www.terraform.io/intro/getting-started/install.html)
+and read the Terraform getting started guide that follows. This guide will
+assume a basic understanding of Terraform.
```
ctrombley: nit: unnecessary line breaks?

website/docs/guides/getting_started.html.markdown
```diff
@@ -0,0 +1,176 @@
...
+* Locate your API key by following [New Relic's Admin API key docs](https://docs.newrelic.com/docs/apis/get-started/intro-apis/types-new-relic-api-keys#admin).
```
ctrombley: Let's use "Admin's API key" for clarity.

website/docs/guides/getting_started.html.markdown
sanderblue: Not sure which line break you're referring to here, but Terraform's markdown flavor has several subtle differences with how it parses the markdown - line breaks being a part of that.

website/docs/guides/getting_started.html.markdown
sanderblue: Good call. Updated.

PR #141, opened 2019-05-22
Add support for outlier NRQL alert conditions

meqif:
Hi! 👋 

As the title says, this pull request adds support for [outlier](https://docs.newrelic.com/docs/alerts/new-relic-alerts/defining-conditions/outlier-detection-nrql-alert) NRQL alert conditions, as per the REST API documentation ([1](https://docs.newrelic.com/docs/alerts/new-relic-alerts/defining-conditions/create-alert-conditions-nrql-queries), [2](https://docs.newrelic.com/docs/alerts/rest-api-alerts/new-relic-alerts-rest-api/alerts-conditions-api-field-names)). The outlier condition type requires the `expected_groups` and `ignore_overlap` fields, which have been added as well.

This PR also adds the static alert condition type, which is simply the name of the default type and is usually omitted.

In order to support this, an upgrade to the `go-newrelic` dependency is required, which I've included as a separate commit.

Since the 'baseline' alert condition type is not yet supported by the REST API ([1](https://docs.newrelic.com/docs/alerts/rest-api-alerts/new-relic-alerts-rest-api/rest-api-calls-new-relic-alerts#excluded)), it was left out of this PR.

Please let me know if there's anything you'd like me to improve.

Issue #50, opened 2018-10-08
Add outlier and baseline as threshold type to nrql alerts


hingstarne:
At the moment of writing the only thresholds that are configurable in the provider are fixed numbers.
Would be great to have the baseline like mentioned [here](https://docs.newrelic.com/docs/alerts/new-relic-alerts/defining-conditions/create-baseline-alert-conditions) as a threshold type.
Are there any plans to add this feature?

bridgetlane:
@hingstarne I do not know of any but feel free to PR the functionality to this repo! Otherwise mark this issue as 'help wanted' so one of our contributors can get to it at some point if able.

hingstarne:
@bridgetlane Thank you for the response. I cannot add labels to an issue, so if you would be so kind to at the help wanted flag?

bridgetlane:
@hingstarne yes, done!

rahulwa:
I don't think New Relic have API for that, as from this doc:
https://docs.newrelic.com/docs/alerts/rest-api-alerts/new-relic-alerts-rest-api/rest-api-calls-new-relic-alerts#conditions-nrql

ctrombley:
Outliers have been addressed in #141, and as @rahulwa mentions, there is a feature gap in New Relic's underlying APIs that don't allow us to create baseline alerts currently.  We will revisit this if the gap is filled in the future.

PR #231, opened 2019-11-21
Document type arg

eheydrick:
Support for different `type` args was added in #141 (Issue #50) but the docs need to be updated as well. This documents the possible values for `type`.


sanderblue:
Thank you for the contribution! Not a blocker by any means, but it looks like PR #141 also added support for arguments `expected_groups` and `ignore_overlap` [here](https://github.com/terraform-providers/terraform-provider-newrelic/pull/141/files#diff-aafc20410fcb6b08b9c712e11eec6587R42-R49).  If it's not too much trouble and you're comfortable adding those to the docs as well, it would be much appreciated.

eheydrick:
@sanderblue For sure no problem, I'll get those other args added.

eheydrick:
@sanderblue Added the `expected_groups` and `ignore_overlap` args. I marked them as optional - as I understand it you'd only pass those if you're using the `outlier` type and not for the more common static type. If you have a suggestion for conveying they are required only for that type I'm all ears.

Issue #47, opened 2018-10-01
Alerts for Plugin Conditions


Thadir:
Good day,,

We started to maintain our configuration of New Relic in terraform, and up to now with the plugin version 1.0.1 its working great. However, we created some new relic plugins that provide custom metrics and I want to create custom alerts on these metrics:

The api is available:
https://docs.newrelic.com/docs/alerts/rest-api-alerts/new-relic-alerts-rest-api/rest-api-calls-new-relic-alerts#plugins-conditions but I cannot find if this feature is in the documentation.

bridgetlane:
Hi @Thadir! Terraform is an open source project and as such providers like this are largely supported by the community and contributors. This means that not every single API call in the New Relic API is currently available. If you contribute/make a pull request for the additional functionality you want I would be happy to review it. Otherwise please make this issue more clear with the functionality you are requesting it and label it 'help wanted'.

atrepca:
I might have bumped into the same limitation, so I'll try to describe: when adding an alert condition you can select the 'Plugins' product, then the type of plugin, like below:
<img width="692" alt="screen shot 2018-10-22 at 11 57 55 pm" src="https://user-images.githubusercontent.com/2030087/47319184-afd05e00-d656-11e8-9925-0b7f3aa4781e.png">
Currently `type` can be one of: `apm_app_metric, apm_jvm_metric, apm_kt_metric, servers_metric, browser_metric, mobile_metric`, so no plugins.

bridgetlane:
@atrepca yup thats going to have to be a feature addition - added the 'help wanted' label to this.

maikelmertens:
I'm too eagerly wanting this feature request implemented as we are working with some NewRelic plugins and defined alerts on it.

Although we want to automate and define it within code, this unfortunately isn't possible yet.

ctrombley:
Resolved by #234 .

PR #234, opened 2019-11-22
Alert Plugins Conditions

mikery:
Hi,

I need to manage Plugin Alerts from Terraform so took a first pass at implementing #47. If this approach is acceptable and feasible to merge I'll update the PR with docs.

### DIFF
--- a/newrelic/resource_newrelic_alert_plugins_condition_test.go
+++ b/newrelic/resource_newrelic_alert_plugins_condition_test.go
@@ -1,0 +1,513 @@
+package newrelic
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccNewRelicAlertPluginsCondition_Basic(t *testing.T) {
+	rName := acctest.RandString(5)
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckNewRelicAlertPluginsConditionDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckNewRelicAlertPluginsConditionConfig(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNewRelicAlertPluginsConditionExists("newrelic_alert_plugins_condition.foo"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "name", fmt.Sprintf("tf-test-%s", rName)),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "type", "apm_app_metric"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "enabled", "false"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "runbook_url", "https://foo.example.com"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "entities.#", "1"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "term.#", "1"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "term.1025554152.duration", "5"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "term.1025554152.operator", "below"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "term.1025554152.priority", "critical"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "term.1025554152.threshold", "0.75"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "term.1025554152.time_function", "all"),
+				),
+			},
+			{
+				Config: testAccCheckNewRelicAlertPluginsConditionConfigUpdated(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNewRelicAlertPluginsConditionExists("newrelic_alert_plugins_condition.foo"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "name", fmt.Sprintf("tf-test-updated-%s", rName)),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "enabled", "true"),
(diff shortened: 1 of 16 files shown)


newrelic/resource_newrelic_alert_plugins_condition_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+		ids, err := parseIDs(rs.Primary.ID, 2)
+		if err != nil {
+			return err
+		}
+
+		policyID := ids[0]
+		id := ids[1]
+
+		found, err := client.GetAlertCondition(policyID, id)
```
ctrombley: This needs to call `client.GetAlertPluginsCondition`.

newrelic/resource_newrelic_alert_plugins_condition_test.go
```diff
@@ -0,0 +1,513 @@
...
+	for _, r := range s.RootModule().Resources {
+		ids, err := parseIDs(r.Primary.ID, 2)
+		if err != nil {
+			return err
+		}
+
+		policyID := ids[0]
+		id := ids[1]
+
+		_, err = client.GetAlertCondition(policyID, id)
```
ctrombley: This needs to call `client.GetAlertPluginsCondition`.

newrelic/resource_newrelic_alert_plugins_condition_test.go
```diff
@@ -0,0 +1,513 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckNewRelicAlertPluginsConditionDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckNewRelicAlertPluginsConditionConfig(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNewRelicAlertPluginsConditionExists("newrelic_alert_plugins_condition.foo"),
+					resource.TestCheckResourceAttr(
+						"newrelic_alert_plugins_condition.foo", "name", fmt.Sprintf("tf-test-%s", rName)),
+					resource.TestCheckResourceAttr(
```
ctrombley: This check for `type` should be deleted, since it is not a valid attribute for this resource.

newrelic/resource_newrelic_alert_plugins_condition_test.go
```diff
@@ -0,0 +1,513 @@
...
+}
+
+resource "newrelic_alert_policy" "foo" {
+  name = "tf-test-updated-%[1]s"
+}
+
+resource "newrelic_alert_plugins_condition" "foo" {
+  policy_id = "${newrelic_alert_policy.foo.id}"
+
+  name               = "tf-test-%[1]s"
```
ctrombley: This attribute should be set to `"tf-test-updated-%[1]s"` to match the assertion above.

newrelic/resource_newrelic_alert_plugins_condition_test.go
```diff
@@ -0,0 +1,513 @@
...
+  enabled            = false
```
ctrombley: This attribute should be set to `true` to match the assertion above.

newrelic/resource_newrelic_alert_plugins_condition_test.go
```diff
@@ -0,0 +1,513 @@
...
+  entities           = ["${data.newrelic_application.app.id}"]
+  runbook_url        = "https://foo.example.com"
```
ctrombley: This attribute should be set to `"https://bar.example.com"` to match the assertion above.

newrelic/resource_newrelic_alert_plugins_condition_test.go
```diff
@@ -0,0 +1,513 @@
...
+				Config:    testAccCheckNewRelicAlertPluginsConditionConfig(rName),
+				Check:     testAccCheckNewRelicAlertPluginsConditionExists("newrelic_alert_plugins_condition.foo"),
+			},
+		},
+	})
+}
+
+func testAccCheckNewRelicAlertPluginsConditionDestroy(s *terraform.State) error {
+	client := testAccProvider.Meta().(*ProviderConfig).Client
+	for _, r := range s.RootModule().Resources {
```
ctrombley: Required at the top of this loop:

```
if r.Type != "newrelic_plugins_alert_condition" {
   continue
}
```
This is since there is also an alert policy defined in state at this point.

mikery:
Thanks @ctrombley for the quick response and detailed review!

> * Can we change the name of the resource definition func and associated file for consistency with the others?  `resourceNewRelicPluginsAlertCondition` lines up a bit better.  All references in the test file's HCL to the `newrelic_alert_plugins_condition` label will need to be updated to match as well.

I originally named it `NewRelicPluginsAlertCondition`, but renamed to `..AlertPluginsCondition` for consistency with the NR API and Go newrelic library. `newrelic_plugins_alert_condition` is a much clearer name, so will happily change back :-) 

> * This seems to call for a data source for Plugins, but that may be out of the scope of this PR.

I would also find this useful. Would a minimal implementation that allows looking up the plugin ID via its GUID be sufficient?
```
data "newrelic_plugin" "my_plugin" {
  guid = "com.example.my_plugin"
}

resource "foo" "bar" {
  plugin_id = "${data.newrelic_plugin.my_plugin.plugin_id}"
}
```
If that would be ok I can include it with this PR. It solves my immediate use case (no need to hardcode plugin IDs in my TF manifests).

If we'd need more functionality (e.g. access to the list of summary metrics), I'd be happy to work on it if you can scope it out in a separate issue and give some example usages you'd like to see.

> * In the meantime, I encourage you to try and get the acceptance tests running against your own account.  You can do this with `NEWRELIC_API_KEY=<api_key> NEWRELIC_LICENSE_KEY=<license_key> make testacc`.  Ideally we will have those passing once the feedback I've provided here has been implemented.

Cool, I'll try that after making the requested changes.

mikery:
@ctrombley I've applied the requested changes, added docs, and added a basic data source for looking up plugins. I haven't yet tried the integration tests as I'd like to set up an isolated NR account for this.

ctrombley:
Thanks for all your work on this PR @mikery!  I rebased on master, added a plugin component data source, and refactored the acceptance tests so that they are passing on our internal account.  I'm going to have someone else on my team do a final review and then we'll merge this in.  

We are going to stick with "newrelic_plugins_alert_conditions" for consistency with the underlying API, even though I do agree that it sounds clearer in the singular when spoken.

I am curious to know if it's possible to create a plugins alert condition without referencing an entity (plugin component), as suggested in the docs you wrote for the plugin data source.  Do you have an example of a plugin that behaves this way?

We will eventually chase this up with another PR to fill in the detailed metric data for the two new data sources.

ctrombley:
Resolves #47.

mikery:
> Thanks for all your work on this PR @mikery!
My pleasure, thanks for getting this merged in and adding the other data source!

> I am curious to know if it's possible to create a plugins alert condition without referencing an entity (plugin component), as suggested in the docs you wrote for the plugin data source. Do you have an example of a plugin that behaves this way?

That was an oversight on my part. The entities list is required. 
> 
> We will eventually chase this up with another PR to fill in the detailed metric data for the two new data sources.

Please cc/@ me when you add an issue for this. I'll try to pick it up, schedule permitting.

PR #246, opened 2019-11-26
add resource for custom events

jamesgoodhouse:
This new resource implements https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/introduction-event-api for allowing the insertion of a custom metrics into Insights.


jthurman42:
@jamesgoodhouse before getting to far down the marshaling / client side, we do have a `go-insights` client already that could be leveraged that abstracts much of that already for the user: https://github.com/newrelic/go-insights

jamesgoodhouse:
@jthurman42 Thanks for the info. I didn't spend any time looking shamefully, but @ctrombley pointed me in the right direction earlier today.

website/docs/index.html.markdown
```diff
@@ -68,8 +68,13 @@ resource "newrelic_alert_policy_channel" "alert_email" {
 
 The following arguments are supported:
 
-- `api_key` - (Required) Your New Relic API key. Can also use `NEWRELIC_API_KEY` environment variable.
+- `api_key` - (Optional) Your New Relic API key. Can also use `NEWRELIC_API_KEY` environment variable.
```
ctrombley: Can we add here "Required for all resources except `newrelic_insights_event`"?

website/docs/index.html.markdown
```diff
@@ -68,8 +68,13 @@ resource "newrelic_alert_policy_channel" "alert_email" {
...
 - `api_url` - (Optional) Possibility to change the API url (default is https://api.newrelic.com/v2). This is for instance relevant if the New Relic account is in the EU. The API url must then be set to https://api.eu.newrelic.com/v2. Can also use `NEWRELIC_API_URL` environment variable.
 - `infra_api_url` - (Optional) Possibility to change the Infra API url (default is https://infra-api.newrelic.com/v2). This is for instance relevant if the New Relic account is in the EU. The Infra API url must then be set to https://infra-api.eu.newrelic.com/v2. Can also use `NEWRELIC_INFRA_API_URL` environment variable.
 - `insecure_skip_verify` - (Optional) Trust self-signed SSL certificates. If omitted, the `NEWRELIC_API_SKIP_VERIFY` environment variable is used.
+- `insights_account_id` - (Optional) Your New Relic Account ID used when inserting or querying Insights events. Can also use `NEWRELIC_INSIGHTS_ACCOUNT_ID` environment variable.
```
ctrombley: Can we change the wording of these to reference the actual resource name here (`newrelic_insights_event`)?

website/docs/index.html.markdown
```diff
@@ -68,8 +68,13 @@ resource "newrelic_alert_policy_channel" "alert_email" {
...
+- `insights_insert_key` - (Optional) You Insights insert key used when inserting Insights events. Can also use `NEWRELIC_INSIGHTS_INSERT_KEY` environment variable.
```
ctrombley: nit: s/You/Your

website/docs/index.html.markdown
```diff
@@ -68,8 +68,13 @@ resource "newrelic_alert_policy_channel" "alert_email" {
...
+- `insights_insert_url` - (Optional) Possibility to change the Insights  insert url (default is https://insights-collector.newrelic.com/v1/accounts). Can also use `NEWRELIC_INSIGHTS_INSERT_URL` environment variable.
```
ctrombley: - nit: extra space after "Insights"
- Can we copy from above? "This is for instance relevant if the New Relic account is in the EU. The Insights API url must then be set to ..."

website/docs/index.html.markdown
```diff
@@ -68,8 +68,13 @@ resource "newrelic_alert_policy_channel" "alert_email" {
...
+- `insights_query_key` - (Optional) You Insights query key used when querying Insights events. Can also use `NEWRELIC_INSIGHTS_QUERY_KEY` environment variable.
```
ctrombley: nit: s/You/Your

jamesgoodhouse:
@ctrombley should be gtg now on requested fixes. Only thing that jumps out to me is I wasn't sure what the EU collector endpoint was. And also, while reading through that chunk of docs I realized that it might not make sense to document the Insights query URL and key, as I don't believe it's used anywhere except in ACC tests. It could be used in the future if a data resource were created for querying via NRQL. I started that work, but decided it's best left for a different PR if it seems like it's needed by others.

ctrombley:
Hey @jamesgoodhouse, everything is looking good here but there are two more items I think we should address:
- I agree with you about the documentation for the insights_query_key.  let's pull it out.
- Acceptance tests will now fail for any dev who doesn't pass an account ID.  We can go two ways with this:
  - Skip the insights event tests unless they are running in New Relic's account.  See [here](https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/newrelic/data_source_newrelic_plugin_test.go#L12-L14) for an example of how we do this elsewhere, or
  - Update the [testing section of the readme](https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/README.md#testing) and let everyone know that it's required.

I personally am a fan of the first solution, because I'd like to make it as easy as possible to run most of that suite.  Thoughts?

jamesgoodhouse:
@ctrombley: I removed the insights query docs and added the check for nr-internal account during the ACC test

jamesgoodhouse:
Absolutely! Glad to have helped and added a new feature @ctrombley

PR #247, opened 2019-11-26
Backfill alert condition acceptance tests

### LINKED ISSUE #214: Backfill acceptance test coverage for alert conditions resource
### Acceptance Criteria
- Evaluate the test coverage of the alert conditions resource and backfill test coverage where it is missing.  Where applicable, the ideal acceptance test coverage should include scenarios for:
  - Basic creation and import
  - Asserting there is no diff on re-apply when changes have not been introduced
  - Asserting re-apply does not fail if underlying resource has disappeared
  - Updates as needed
  - Deletion
- Take note of any functionality that is not working as intended, feature gaps, or misrepresentations of the underlying API payload

ctrombley:
Resolves #214.
 
This PR backfills acceptance test coverage for the alert conditions resource and refactors the existing test code for simplicity.

### DIFF
--- a/newrelic/resource_newrelic_alert_condition_test.go
+++ b/newrelic/resource_newrelic_alert_condition_test.go
@@ -11,7 +11,8 @@ import (
 )
 
 func TestAccNewRelicAlertCondition_Basic(t *testing.T) {
-	rName := acctest.RandString(5)
+	rName := fmt.Sprintf("tf-test-%s", acctest.RandString(5))
+	rNameUpdated := fmt.Sprintf("tf-test-updated-%s", acctest.RandString(5))
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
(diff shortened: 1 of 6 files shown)


newrelic/resource_newrelic_alert_condition_test.go
```diff
@@ -45,126 +53,148 @@ func TestAccNewRelicAlertCondition_Basic(t *testing.T) {
...
-	resource.Test(t, resource.TestCase{
+	rName := fmt.Sprintf("tf-test-%s", acctest.RandString(5))
+	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckNewRelicAlertConditionDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccCheckNewRelicAlertConditionConfigZeroThreshold(rName),
+				Config: testAccCheckNewRelicAlertConditionConfigThreshold(rName, 0),
```
sanderblue: Not a blocker, but I noticed a pattern in testing examples that the `Config` functions didn't have the `Check` word as part of it. And the subsequent functions passed to `ComposeTestCheckFunc()` contained the `Check` word since it's doing a check vs returning a config.

sanderblue: Here's an [example](https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html) that I was going off of.

ctrombley: I will update - I had meant to put this in.

PR #255, opened 2019-12-03
test: fix failing data source test

ctrombley:
This PR fixes a broken test on master.

Import tests don't make sense for a data source, so we are removing this one here.


sanderblue:
I guess we also should remove the import test from [`data_source_newrelic_alert_policy_test.go`](https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/newrelic/data_source_newrelic_alert_policy_test.go#L29-L34), but that's for another PR :)

PR #292, opened 2020-01-02
changed alert threshold value type to float

prodjito:


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/terraform-providers/terraform-provider-newrelic
+module github.com/prodjito/terraform-provider-newrelic
```
sanderblue: Hi there, we shouldn't change this as it would point to your forked repo and break imports (tests are failing due that currently). 
```
module github.com/terraform-providers/terraform-provider-newrelic
```

newrelic/resource_newrelic_infra_alert_condition.go
```diff
@@ -30,7 +30,7 @@ func thresholdSchema() *schema.Resource {
 	return &schema.Resource{
 		Schema: map[string]*schema.Schema{
 			"value": {
-				Type:     schema.TypeInt,
+				Type:     schema.TypeFloat,
```
sanderblue: Unfortunately this has a bit of a ripple effect that breaks the build. This is due to the typings in the Go client (`go-newrelic`) requiring `Value` to be of type `int` [here](https://github.com/paultyng/go-newrelic/blob/master/api/types.go#L416).  

Although CI passed, it would fail on the `master` branch 😢

go.mod
prodjito: I didn't mean to include that in the PR.

newrelic/resource_newrelic_infra_alert_condition.go
prodjito: @sanderblue @jthurman42  Looks like the go-client has been updated https://github.com/newrelic/newrelic-client-go/pull/71. So this pr is no longer build breaking right?

jthurman42: The Terraform provider has not been fully ported to newrelic/newrelic-client-go yet.  That work is likely to start next week.  (Note: Synthetics was ported, just not the rest)

ctrombley:
Closing this in favor of #312 .

PR #325, opened 2020-01-17
docs: update alert-channel examples

donaldducky:
I noticed some conflicting examples (`config` vs `configuration`) when I was trying out the provider.


jthurman42:
Thanks for creating the PR! We recently deprecated `configuration` in favor of config.  One item to note is that `config` is now a block, so also need to remove the `=` sign.

```
  config {
      myconfigItem = "value"
  }
```

If you can update the PR, I'll get it merged in shortly.  Thanks again!

donaldducky:
Cool, I removed the `=` 👍 

I was wondering why I couldn't find `configuration {` when I was searching lol

jthurman42:

PR #358, opened 2020-01-23
feat: replace provider backend with newrelic-client-go

### LINKED ISSUE #323: Error unmarshaling `newrelic_alert_channel` configuration headers after release v1.12.0
Hi there,

It appears that the attempted forwards-compatible change for deprecating `configuration` in the `newrelic_alert_channel` has broken unmarshaling of configuration headers in the PR #307, release https://github.com/terraform-providers/terraform-provider-newrelic/releases/tag/v1.12.0

### Terraform Version
`0.12.6`

### Affected Resource(s)
- newrelic_alert_channel

#### Working on v1.11.0, broken on v1.12.0
```hcl
resource "newrelic_alert_channel" "slack_warning" {
  name = "Some slack channel"
  type = "slack"

  configuration = {
    channel = "#some-slack-channel"
    url     = var.slack_url
  }

  // https://github.com/terraform-providers/terraform-provider-newrelic/issues/5
  lifecycle {
    ignore_changes = ["configuration"]
  }
}
```

### Expected Behavior
It appears as though there was an attempt to add forwards compatible support to v1.12.0 for the configuration object in order to support migration from `configuration = {` to `config {` so I would expect that to work with a deprecation warning.

### Actual Behavior
While using the above Terraform configuration on v1.12.0 of the NR TF provider, I receive the following error running `terraform plan`. The deprecated warning is expected and we are working through the migration but the JSON unmarshal error is not expected:

```
Warning: "configuration": [DEPRECATED] use `config` block instead

  on .terraform/modules/newrelic-alerts/newrelic-alerts/alert_channels.tf line 1, in resource "newrelic_alert_channel" "slack_warning":
   1: resource "newrelic_alert_channel" "slack_warning" {

Error: json: cannot unmarshal string into Go struct field ChannelConfiguration.channels.configuration.headers of type map[string]string
```

### Steps to Reproduce
1. `terraform plan` with the above example Terraform configuration and v1.12.0 of the TF NR provider

### Important Factoids
It also appears that `config` is not backwards compatible before v1.12.0 which makes this a hard cutover (and potentially a breaking change for other consumers)

### References
- #307

### LINKED ISSUE #354: Migrate fully to newrelic-client-go
### Objective
Migrate remaining code to use `newrelic-client-go` client.

### Acceptance Criteria
- No references to go-newrelic should remain

ctrombley:
Resolves #354 .

This PR removes the reference to the go-newrelic module, and puts a single instance of the `NewRelic` client in place for the backend of the provider.


newrelic/resource_newrelic_alert_condition.go
```diff
@@ -118,7 +118,7 @@ func resourceNewRelicAlertCondition() *schema.Resource {
 			},
 			"condition_scope": {
 				Type:     schema.TypeString,
-				Optional: true,
+				Required: true,
```
sanderblue: Is this a breaking change?

newrelic/resource_newrelic_alert_policy_test.go
```diff
@@ -173,12 +174,16 @@ func testAccCheckNewRelicAlertPolicyExists(n string) resource.TestCheckFunc {
 
 func testAccDeleteAlertPolicy(name string) func() {
```
sanderblue: Not a blocker, but it looks like we have a duplicate of this function in [helpers_test.go](https://github.com/terraform-providers/terraform-provider-newrelic/pull/358/files#diff-8644a294a323fa7b442332ed8b0d2116R52).

newrelic/resource_newrelic_alert_condition.go
sanderblue: The API apparently throws an error and this is required per our docs, so this should be required so the update is correct.

Issue #269, opened 2019-12-05
Add documentation on how to prevent always re-creating alerts channels with secrets


ctrombley:
Related to https://github.com/terraform-providers/terraform-provider-newrelic/issues/130#issuecomment-562313261.

We should include some notes in the documentation on how to work around constantly re-provisioning alerts channels that contain secrets in their configuration.

zlesnr:
https://github.com/terraform-providers/terraform-provider-newrelic/pull/364

PR #364, opened 2020-01-27
docs: Make note about ignoring secrets

zlesnr:
resolves: #269 

Without this change, users could be left wondering why their resource keeps
getting marked as changed, though no change has been made out of band.  Here we
make a note about the reason a change gets marked, and include an example to
avoid this behavior.


website/docs/guides/getting_started.html.markdown
```diff
@@ -156,6 +156,56 @@ resource "newrelic_alert_policy_channel" "alert_policy_email" {
...
+for every run.
+
+Consider the following example.
+
+```hcl
+resource "newrelic_alert_channel" "slack" {
+  name = "slack"
+  type = "slack"
+
+  configuration = {
```
sanderblue: The `configuration` block has been deprecated. We want to encourage others to use the new `config` block instead. Could we update these docs to reflect that too potentially?

zlesnr: Indeed yes.  I've updated the doc to use `config` in place of `configuration` and checked for other references.

Issue #319, opened 2020-01-17
Add ability to provision multiple alert notification channels to an alert policy


sanderblue:
**Objective**
Provide the ability to apply multiple notification channels to an alert policy. Consider adding ability to add multiple channels to multiple policies as well (optional). 

**Notes**
This will help reduce code repetition and provider a slightly better API for developers to utilize when provisioning this resource.

We should provide a new method in `newrelic-client-go` to facilitate this.

PR #365, opened 2020-01-28
feat(alert_policy_channels): add ability to add multiple channels to a policy

sanderblue:
Resolves: #319 

This PR introduces the ability to provision multiple alert channels per a policy via the `/v2/alerts_policy_channels` endpoint. Prior to this, we could only provision a single channel per HCL block. This will greatly reduce the amount of HCL code required to accomplish adding multiple notification channels to a policy.

### DIFF
--- a/newrelic/helpers.go
+++ b/newrelic/helpers.go
@@ -28,6 +28,24 @@ func parseIDs(serializedID string, count int) ([]int, error) {
 	return ids, nil
 }
 
+// Converts a hash of IDs into an array.
+// Examples: "12345:54432:66564" -> []int{12345,54432,66564}
+func parseHashedIDs(serializedID string) ([]int, error) {
+	rawIDs := strings.Split(serializedID, ":")
+	ids := make([]int, len(rawIDs))
+
+	for i, rawID := range rawIDs {
+		id, err := strconv.ParseInt(rawID, 10, 32)
+		if err != nil {
+			return ids, err
+		}
+
+		ids[i] = int(id)
+	}
+
+	return ids, nil
+}
+
 func serializeIDs(ids []int) string {
 	idStrings := make([]string, len(ids))
 
--- a/newrelic/helpers.go
+++ b/newrelic/helpers.go
@@ -3,6 +3,7 @@ package newrelic
 import (
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strconv"
 	"strings"
 	"unicode"
@@ -73,4 +74,11 @@ func stripWhitespace(str string) string {
 		}
 	}
 	return b.String()
+}
+
+// Mutates original slice
+func sortIntegerSlice(integers []int) {
+	sort.Slice(integers, func(i, j int) bool {
+		return integers[i] < integers[j]
+	})
 }
--- a/newrelic/provider_test.go
+++ b/newrelic/provider_test.go
@@ -3,7 +3,6 @@ package newrelic
 import (
 	"fmt"
 	"os"
-	"strings"
 	"testing"
 	"time"
 
@@ -114,32 +113,3 @@ func testAccCreateApplication(t *testing.T) {
 
 // 	testAccCleanupComplete = true
 // }
-
-// testCheckResourceAttr validates the value in state for the given name/key combination.
-func testCheckResourceAttr(is *terraform.InstanceState, name string, key string, value string) error {
-	// Empty containers may be elided from the state.
-	// If the intent here is to check for an empty container, allow the key to
-	// also be non-existent.
-	emptyCheck := false
-	if value == "0" && (strings.HasSuffix(key, ".#") || strings.HasSuffix(key, ".%")) {
-		emptyCheck = true
-	}
-
-	if v, ok := is.Attributes[key]; !ok || v != value {
-		if emptyCheck && !ok {
-			return nil
-		}
-
-		if !ok {
-			return fmt.Errorf("%s: Attribute '%s' not found", name, key)
-		}
-
-		return fmt.Errorf(
-			"%s: Attribute '%s' expected %#v, got %#v",
-			name,
-			key,
-			value,
-			v)
-	}
-	return nil
-}
--- a/newrelic/provider_test.go
+++ b/newrelic/provider_test.go
@@ -3,6 +3,7 @@ package newrelic
 import (
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
@@ -113,3 +114,32 @@ func testAccCreateApplication(t *testing.T) {
 
 // 	testAccCleanupComplete = true
 // }
+
+// testCheckResourceAttr validates the value in state for the given name/key combination.
+func testCheckResourceAttr(is *terraform.InstanceState, name string, key string, value string) error {
+	// Empty containers may be elided from the state.
+	// If the intent here is to check for an empty container, allow the key to
+	// also be non-existent.
+	emptyCheck := false
+	if value == "0" && (strings.HasSuffix(key, ".#") || strings.HasSuffix(key, ".%")) {
+		emptyCheck = true
+	}
+
+	if v, ok := is.Attributes[key]; !ok || v != value {
+		if emptyCheck && !ok {
+			return nil
+		}
+
+		if !ok {
+			return fmt.Errorf("%s: Attribute '%s' not found", name, key)
+		}
+
+		return fmt.Errorf(
+			"%s: Attribute '%s' expected %#v, got %#v",
+			name,
+			key,
+			value,
+			v)
+	}
+	return nil
+}
--- a/newrelic/resource_newrelic_alert_policy_channel.go
+++ b/newrelic/resource_newrelic_alert_policy_channel.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/newrelic/newrelic-client-go/newrelic"
+	"github.com/newrelic/newrelic-client-go/pkg/alerts"
 	"github.com/newrelic/newrelic-client-go/pkg/errors"
 )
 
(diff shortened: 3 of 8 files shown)


newrelic/helpers.go
```diff
@@ -27,6 +27,24 @@ func parseIDs(serializedID string, count int) ([]int, error) {
 	return ids, nil
 }
 
+// Converts a hash of IDs into an array.
+// Examples: "12345:54432:66564" -> []int{12345,54432,66564}
+func parseHashedIDs(serializedID string) ([]int, error) {
```
zlesnr: Perhaps a test for this one would be good.

newrelic/resource_newrelic_alert_policy_channel.go
```diff
@@ -136,3 +166,19 @@ func policyChannelExists(client *newrelic.NewRelic, policyID int, channelID int)
 
 	return false, nil
 }
+
+func policyChannelsExist(client *newrelic.NewRelic, policyID int, channelIDs []int) (bool, error) {
```
zlesnr: Perhaps here also a test would be good.

newrelic/structures_newrelic_alert_policy_channel.go
```diff
@@ -0,0 +1,11 @@
+package newrelic
+
+func expandChannelIDs(channelIDs []interface{}) []int {
```
zlesnr: A test for this one seems good too.

newrelic/helpers.go
sanderblue: Agreed

newrelic/structures_newrelic_alert_policy_channel.go
sanderblue: Added

newrelic/resource_newrelic_alert_policy_channel.go
sanderblue: This is covered in integration tests 😎 Did you want a unit test for this?

zlesnr: Oh not if its covered.

newrelic/resource_newrelic_alert_policy_channel.go
```diff
@@ -83,35 +103,37 @@ func resourceNewRelicAlertPolicyChannelRead(d *schema.ResourceData, meta interfa
 	}
 
 	d.Set("policy_id", policyID)
-	d.Set("channel_id", channelID)
+	d.Set("channel_ids", channelIDs)
```
ctrombley: Do we need to set `channel_id` here?

newrelic/resource_newrelic_alert_policy_channel.go
```diff
@@ -136,3 +158,19 @@ func policyChannelExists(client *newrelic.NewRelic, policyID int, channelID int)
 
 	return false, nil
 }
+
+func policyChannelsExist(client *newrelic.NewRelic, policyID int, channelIDs []int) (bool, error) {
+	for _, id := range channelIDs {
+		channelExists, err := policyChannelExists(client, policyID, id)
```
ctrombley: Can we batch this so we don't have to make a request each time for each iteration?

newrelic/resource_newrelic_alert_policy_channel.go
sanderblue: This is where the "interesting" logic of whether or not to set `channel_id` or `channel_ids` comes into play.

newrelic/resource_newrelic_alert_policy_channel.go
sanderblue: Refactored to only make one request instead of multiple.

newrelic/resource_newrelic_alert_policy_channel.go
sanderblue: Logic has been updated and is ready for re-review.

newrelic/resource_newrelic_alert_policy_channel.go
```diff
@@ -83,56 +99,103 @@ func resourceNewRelicAlertPolicyChannelRead(d *schema.ResourceData, meta interfa
...
+		d.Set("channel_id", parsedChannelIDs[0])
+	}
+
+	if channelIDsOk && len(parsedChannelIDs) > 0 {
+		d.Set("channel_ids", parsedChannelIDs)
+	}
+
+	// Handle import (set `channel_ids` since this resource doesn't exist in state yet)
+	if !channelIDOk && !channelIDsOk {
+		d.Set("channel_ids", parsedChannelIDs)
```
zlesnr: Hmm.  In the case where a new resource has only `channel_id` set, does this tell them about an inaccurate create during the plan since we're setting the `channel_ids` plural?

newrelic/provider_test.go
```diff
@@ -113,3 +114,32 @@ func testAccCreateApplication(t *testing.T) {
 
 // 	testAccCleanupComplete = true
 // }
+
+// testCheckResourceAttr validates the value in state for the given name/key combination.
+func testCheckResourceAttr(is *terraform.InstanceState, name string, key string, value string) error {
```
ctrombley: Why was this necessary to copy in?  [`TestCheckResourceAttr`](https://github.com/hashicorp/terraform/blob/master/helper/resource/testing.go#L955) is exported by the testing package, does it not get the job done in this case?

website/docs/r/alert_policy_channel.html.markdown
```diff
@@ -38,4 +59,19 @@ resource "newrelic_alert_policy_channel" "foo" {
 The following arguments are supported:
 
   * `policy_id` - (Required) The ID of the policy.
-  * `channel_id` - (Required) The ID of the channel.
+  * `channel_ids` - (Optional*) Array of channel IDs to apply to the specified policy.
+  * `channel_id` - **Deprecated!** (Optional*) The ID of the channel. Please use the `channel_ids` argument instead.
+
+<sup>\*Note: Even though **channel_id** and **channel_ids** are optional, at least one of those arguments must be used for this resource to work.</sup>
+
+## Import
```
ctrombley: 👍

newrelic/resource_newrelic_alert_policy_channel.go
```diff
@@ -24,55 +25,70 @@ func resourceNewRelicAlertPolicyChannel() *schema.Resource {
...
-	policyID := d.Get("policy_id").(int)
-	channelID := d.Get("channel_id").(int)
+	if err != nil {
+		return err
+	}
 
-	serializedID := serializeIDs([]int{policyID, channelID})
+	serializedID := serializeIDs(append(
+		[]int{policyChannels.ID},
+		policyChannels.ChannelIDs...,
```
ctrombley: If we don't sort these, changing their order in HCL will result in drift.   Probably not a big deal, but something to consider?

newrelic/resource_newrelic_alert_policy_channel_test.go
```diff
@@ -32,6 +33,39 @@ func TestAccNewRelicAlertPolicyChannel_Basic(t *testing.T) {
 				),
 			},
 			// Test: Import
+			{
+				ResourceName:     resourceName,
+				ImportState:      true,
+				ImportStateCheck: testAccNewRelicAlertPolicyImportStateCheckFunc(),
```
ctrombley: Can this be replaced with a call to `TestCheckResourceAttr`?

newrelic/resource_newrelic_alert_policy_channel.go
```diff
@@ -83,56 +99,103 @@ func resourceNewRelicAlertPolicyChannelRead(d *schema.ResourceData, meta interfa
 	}
 
 	d.Set("policy_id", policyID)
-	d.Set("channel_id", channelID)
+
+	_, channelIDOk := d.GetOk("channel_id")
```
ctrombley: Can some of the `Set`s here be refactored to a `flatten` func?

newrelic/resource_newrelic_alert_policy_channel.go
sanderblue: Good call.

newrelic/resource_newrelic_alert_policy_channel_test.go
sanderblue: `ImportStateCheck` expects the type `ImportStateCheckFunc` which works with different types in the arguments that are passed to the pointer function.

newrelic/provider_test.go
sanderblue: Well I could've written my own version of this, but this seemed like something we could use in other custom import check functions.

sanderblue: I can try to see if I can call one of the functions that TF exports inside the custom check I created. I tried several ways to do that, but couldn't get it to work like we'd want.

newrelic/resource_newrelic_alert_policy_channel.go
sanderblue: Even if we sort the IDs and save them in order within the state, TF will still detect a change if the user changes the order of the IDs in the HCL in a way that doesn't match the state, which then creates a new resource due to `ForceNew: true`

sanderblue: But sorting does avoid some form of drift, but there's still a chance a human change will cause drift. We can add a note in the docs to say "Best practice: sort your channel IDs in ascending order within your HCL to avoid state drift"

newrelic/resource_newrelic_alert_policy_channel.go
sanderblue: Updated

newrelic/resource_newrelic_alert_policy_channel.go
sanderblue: After importing a subsequent `plan` will provide a deprecation warning message.

newrelic/provider_test.go
sanderblue: Or I could do this
```go
func testAccNewRelicAlertPolicyImportStateCheckFunc(resourceName string, t *testing.T) resource.ImportStateCheckFunc {
	return func(state []*terraform.InstanceState) error {
		expectedChannelsCount := "1"
		channelsCount := state[0].Attributes["channel_ids.#"]

		if channelsCount != expectedChannelsCount {
			return fmt.Errorf(
				"%s: Attribute '%s' expected %#v, got %#v",
				resourceName,
				"channel_ids.#",
				expectedChannelsCount,
				channelsCount,
			)
		}

		return nil
	}
}
```

newrelic/resource_newrelic_alert_policy_channel.go
ctrombley: 👍

Issue #156, opened 2019-07-13
newrelic_infra_alert_condition support for violation_close_timer


novas0x2a:
### Affected Resource(s)
Please list the resources as a list, for example:
- newrelic_infra_alert_condition 

Right now, newrelic_alert_condition supports violation_close_timer but newrelic_infra_alert_condition does not. This is most likely due to the [docs not mentioning the field](https://docs.newrelic.com/docs/infrastructure/new-relic-infrastructure/infrastructure-alert-conditions/rest-api-calls-new-relic-infrastructure-alerts#).

However, [go-newrelic](https://github.com/paultyng/go-newrelic/blob/6bd4f0fecd759bc18c4806cba51dcb97ecdffbbc/api/types.go#L365) supports it (merged in https://github.com/paultyng/go-newrelic/pull/54); so I think this is actually a doc bug.

PR #370, opened 2020-01-29
feat(infra_alert_condition): add violation_close_timer to newrelic_infra_alert_condition resource

sanderblue:
Resolves: #156

### DIFF
--- a/newrelic/resource_newrelic_infra_alert_condition.go
+++ b/newrelic/resource_newrelic_infra_alert_condition.go
@@ -1,12 +1,10 @@
 package newrelic
 
 import (
-	"fmt"
 	"log"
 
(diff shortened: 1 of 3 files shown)


zlesnr:
Looks good to me.  The interface casting bits look a bit strange, but the tests are passing.

samjgalbraith:
I'm trying to use this attribute in 1.16.0 and it's not working for violation_close_timer = 0. It just sets the value to default (24) on all newrelic_infra_alert_condition resources upon create. Running apply again, the resource realizes the mismatch and plans to change the violation_close_timer  24 => 0. Executing the plan, however, fails to actually change the value, so there's a perpetual state mismatch.

samjgalbraith:
^ I've opened an issue  #553

Issue #372, opened 2020-01-30
newrelic_alert_channel PagerDuty resource marked as changed rather than recreated


stuntbaboon:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
- 0.11.14
- provider "newrelic" (1.12.2)

### Affected Resource(s)
newrelic_alert_channel

### Terraform Configuration Files
```hcl
resource "newrelic_alert_channel" "alerts_channel_pagerduty" {
  name  = "PagerDuty"
  type  = "pagerduty"

  config {
    service_key = "${data.terraform_remote_state.pagerduty.newrelic_integration_key[0]}"
  }
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
What should have happened?

I expected the newrelic_alert_channel resource to be recreated as mentioned in issue - https://github.com/terraform-providers/terraform-provider-newrelic/issues/130 due to the service_key secret.

### Actual Behavior
What actually happened?

An attempt was made to change the newrelic_alert_channel resource and it failed.

Has this behaviour changed in version 1.12.2.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:

`terraform apply` of the above configuration results in the following terraform output:

```shell
~ newrelic_alert_channel.eabn_bxf_processing_alerts_channel_pagerduty
      config.0.service_key:         <sensitive> => <sensitive> (attribute changed)

Error: Error applying plan:

1 error occurred:
	* newrelic_alert_channel.alerts_channel_pagerduty: 1 error occurred:
	* newrelic_alert_channel.alerts_channel_pagerduty: doesn't support update
```

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- https://github.com/terraform-providers/terraform-provider-newrelic/issues/130
- https://github.com/terraform-providers/terraform-provider-newrelic/issues/5

[...]

PR #387, opened 2020-02-07
fix(newrelic_alert_channel): Force new resource for all config fields

zlesnr:
resolves: #372


rlmartin:
Amazing. When will a version of this be tagged/released? Thanks.

zlesnr:
Likely Wednesday @rlmartin.

PR #398, opened 2020-02-13
feat(alert_policy): add ability to add multiple channels to a policy

sanderblue:
Resolves: #319

The PR introduces the ability to add alert channels to a policy during resource creation by allowing the user to supply an array of channel IDs in their Terraform config for a `newrelic_alert_policy` resource.

Example:
```hcl
resource "newrelic_alert_policy" "foo" {
  name = "example-policy"
  channel_ids = [
    3156981,
    3103316
  ]
}
```

It should be noted that the user will _not_ be able to import the channel IDs. However, a user can still import the primary resource, add channel IDs to their HCL, and then use `terraform apply` to add the channels to the policy.

### DIFF
--- a/newrelic/resource_newrelic_alert_policy.go
+++ b/newrelic/resource_newrelic_alert_policy.go
@@ -66,7 +66,7 @@ func resourceNewRelicAlertPolicyCreate(d *schema.ResourceData, meta interface{})
 	channels := d.Get("channel_ids").([]interface{})
 
 	if len(channels) > 0 {
-		channelIDs := expandChannelIDs(channels)
+		channelIDs := expandAlertChannelIDs(channels)
 		matchedChannelIDs, err := findExistingChannelIDs(client, channelIDs)
 
 		if err != nil {
@@ -111,7 +111,7 @@ func resourceNewRelicAlertPolicyRead(d *schema.ResourceData, meta interface{}) e
 	channelIDs := d.Get("channel_ids").([]interface{})
 
 	if len(channelIDs) > 0 {
-		d.Set("channel_ids", expandChannelIDs(channelIDs))
+		d.Set("channel_ids", expandAlertChannelIDs(channelIDs))
 	}
 
 	return flattenAlertPolicy(policy, d)
--- a/newrelic/resource_newrelic_alert_policy.go
+++ b/newrelic/resource_newrelic_alert_policy.go
@@ -108,12 +108,6 @@ func resourceNewRelicAlertPolicyRead(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
-	channelIDs := d.Get("channel_ids").([]interface{})
-
-	if len(channelIDs) > 0 {
-		d.Set("channel_ids", expandAlertChannelIDs(channelIDs))
-	}
-
 	return flattenAlertPolicy(policy, d)
 }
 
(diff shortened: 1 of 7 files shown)


newrelic/resource_newrelic_alert_policy.go
```diff
@@ -78,6 +108,12 @@ func resourceNewRelicAlertPolicyRead(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
+	channelIDs := d.Get("channel_ids").([]interface{})
```
sanderblue: Fairly certain we don't need to do this here because state is still preserved due to the fact that if a user adds or removes any channel IDs, the original resource is destroyed and then recreated with the updated channel associations (via the `ForceNew: true` schema option). @ctrombley @zlesnr any thoughts on that?

newrelic/resource_newrelic_alert_policy.go
```diff
@@ -78,6 +108,12 @@ func resourceNewRelicAlertPolicyRead(d *schema.ResourceData, meta interface{}) e
...
+
+	if len(channelIDs) > 0 {
+		d.Set("channel_ids", expandChannelIDs(channelIDs))
```
sanderblue: Fairly certain we don't need to  this (lines 111-115) because state is still preserved due to the fact that if a user adds or removes any channel IDs, the original resource is destroyed and then recreated with the updated channel associations (via the `ForceNew: true` schema option). @ctrombley @zlesnr any thoughts on that?

newrelic/structures_newrelic_alert_policy.go
```diff
@@ -20,6 +20,16 @@ func expandAlertPolicy(d *schema.ResourceData) *alerts.Policy {
 	return &policy
 }
 
+func expandChannelIDs(channelIDs []interface{}) []int {
```
sanderblue: Any thoughts on where a better home for this might be? Feels like it could go in `structures_newrelic_alert_channel.go` or in a new `structures_newrelic_alert_policy_channels.go` (which is where I put it in the 1st attempt at this particular feature).

zlesnr: Hmm, is this sort of a `util.go` opportunity or something perhaps?

newrelic/resource_newrelic_alert_policy.go
sanderblue: Correct, however since any changes to channel IDs warrants a new resource, this is never needed because the state will be saved on create. I guess it doesn't hurt to keep this here. I know that if I remove this part, it doesn't affect anything, but again, it's probably fine to leave it, especially since it's a bit more explicit.

newrelic/structures_newrelic_alert_policy.go
sanderblue: I should provide more context - the `structures_*.go` file pattern is a Terraform prescribed pattern that involves "expanding" (usually associated with creation or updating) and "flattening" (usually associated with reads). Some more info can be read [here](https://www.terraform.io/docs/extend/writing-custom-providers.html#implementing-a-more-complex-read). I guess my questions was more around which domain is best for `expandChannelIDs()` - I don't think it should be in `alert_policy`, but can't decide between the two mentioned above :)

sanderblue: Since the `structures_newrelic_alert_policy_channels.go` file doesn't exist in this PR, I'm going to put it in `structures_newrelic_alert_channel.go`.

newrelic/resource_newrelic_alert_policy.go
ctrombley: I think you're right - this conditional only evals true when we already are tracking ids within a policy, and since we are forcing new on any change, it's not needed.  👍

sanderblue: Awesome, thanks for the affirmation

PR #407, opened 2020-02-19
feat(synthetics): add labels resource

### LINKED ISSUE #164: Feature request: Synthetic Test labels
Hi gang! I would like to request adding the ability to label our synthetic tests via Terraform. Currently we can only either do this manually or by sending a request to the NR API.

### Terraform Version
```
Terraform v0.12.7
+ provider.newrelic v1.5.1
```

### Affected Resource(s)
- newrelic_synthetics_monitor

### Expected Behavior
Would like to be able to apply labels to my synthetics monitors as per <https://docs.newrelic.com/docs/apis/synthetics-rest-api/label-examples/use-synthetics-label-apis>

### Actual Behavior
Ain't no labels

ctrombley:
Resolves #164.

This PR adds a new synthetics labels resource, following existing patterns.

### DIFF
--- a/newrelic/resource_newrelic_synthetics_label.go
+++ b/newrelic/resource_newrelic_synthetics_label.go
@@ -1,0 +1,149 @@
+package newrelic
+
+import (
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/newrelic/newrelic-client-go/pkg/errors"
+	"github.com/newrelic/newrelic-client-go/pkg/synthetics"
+)
+
+func resourceNewRelicSyntheticsLabel() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNewRelicSyntheticsLabelCreate,
+		Read:   resourceNewRelicSyntheticsLabelRead,
+		Delete: resourceNewRelicSyntheticsLabelDelete,
+		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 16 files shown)


newrelic/resource_newrelic_synthetics_label.go
```diff
@@ -0,0 +1,149 @@
...
+
+	err := client.Synthetics.DeleteMonitorLabel(monitorID, labelType, value)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func expandSyntheticsLabel(d *schema.ResourceData) *synthetics.MonitorLabel {
```
sanderblue: Not a blocker, but could potentially put these into the structures_{resource}.go pattern and add some a couple unit tests for expand and flatten.

sanderblue: however, I see that they are the top-level scenario that we talked about before, so maybe no big deal.

ctrombley: I left them here since there was no data source that needed to share them.

ctrombley: This makes sense though, I'll make it so

ctrombley: @sanderblue that's right, seems like there is no good way to unit test at the top level since the expand and flatten functions depend on an instace of `schema.ResourceData`.  I will leave this as-is after all.

PR #409, opened 2020-02-19
feat(synthetics): add secure credentials resource

### LINKED ISSUE #404: Support Synthetics Secure Credentials
Add support for New Relic Synthetics [Secure Credential](https://docs.newrelic.com/docs/apis/synthetics-rest-api/secure-credentials-examples/use-synthetics-secure-credentials-apis) management:

* Data Source
* Resource

Blocked on downstream newrelic/newrelic-client-go#134

ctrombley:
Resolves #404 .

This PR introduces a resource and data source for Synthetics secure credentials, following existing patterns.


newrelic/data_source_newrelic_synthetics_secure_credential.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceNewRelicSyntheticsSecureCredential() *schema.Resource {
```
zlesnr: For my own education, is there anything here that marks this a data source other than the file name?  This looks pretty similar to the resources below.

jthurman42: A data source only has a Read function, and the Schema excludes fields that are not returned (in this case `value` is not returned by the API, so it's not included in the schema).  I agree that there is a bit of duplication here, but not sure how to remove it without adding complexity to reading the code.

ctrombley: @zlesnr It's wired up in `newrelic/provider.go` in the collection of data sources, rather than in the collection of resources.

PR #411, opened 2020-02-19
fix(provider): deprecate and re-enable the use of infra_api_url

### LINKED ISSUE #402: Allow the `infra_api_url` attribute to be set
### Terraform Version
v1.13.1

### Affected Resource(s)
- `newrelic_infra_alert_condition`

### Description
v1.11.0 introduced an unintentional breaking change that keeps the `infra_api_url` attribute from working as expected.  This attribute was moved to `infrastructure_api_url`, but the old value was not deprecated properly.

ctrombley:
Resolves #402 .

This PR re-enables the use of the `infra_api_url` attribute, but deprecates it in favor of `infrastructure_api_url`.


newrelic/provider.go
```diff
@@ -184,3 +184,19 @@ func providerConfigure(data *schema.ResourceData, terraformVersion string) (inte
 
 	return &providerConfig, nil
 }
+
+func getInfraAPIURL(data *schema.ResourceData) string {
+	old, oldOk := data.GetOk("infra_api_url")
+	new, newOk := data.GetOk("infrastructure_api_url")
```
sanderblue: I think `new` is a keyword in Go. At a minimum it's a function: `new()`

Might want to use `newURL` maybe?

PR #417, opened 2020-02-21
Implementation of CRUD operations for NewRelic App Labels

### LINKED ISSUE #414: Add support for NewRelic App Labels
Please add support to creating/listing/deleting of application labels. 
For details see the following:
https://docs.newrelic.com/docs/apis/rest-api-v2/labels-examples-v2/

lukaszczechowski:
Resolves #414

### DIFF
--- a/newrelic/provider.go
+++ b/newrelic/provider.go
@@ -110,6 +110,7 @@ func Provider() terraform.ResourceProvider {
 			"newrelic_alert_condition":              resourceNewRelicAlertCondition(),
 			"newrelic_alert_policy_channel":         resourceNewRelicAlertPolicyChannel(),
 			"newrelic_alert_policy":                 resourceNewRelicAlertPolicy(),
+			"newrelic_application_label":            resourceNewRelicApplicationLabel(),
 			"newrelic_plugins_alert_condition":      resourceNewRelicPluginsAlertCondition(),
 			"newrelic_dashboard":                    resourceNewRelicDashboard(),
 			"newrelic_infra_alert_condition":        resourceNewRelicInfraAlertCondition(),
@@ -120,7 +121,6 @@ func Provider() terraform.ResourceProvider {
 			"newrelic_synthetics_monitor_script":    resourceNewRelicSyntheticsMonitorScript(),
 			"newrelic_synthetics_label":             resourceNewRelicSyntheticsLabel(),
 			"newrelic_synthetics_secure_credential": resourceNewRelicSyntheticsSecureCredential(),
-			"newrelic_app_label":                    resourceNewRelicAppLabel(),
 		},
 	}
 
--- a/newrelic/provider.go
+++ b/newrelic/provider.go
@@ -120,6 +120,7 @@ func Provider() terraform.ResourceProvider {
 			"newrelic_synthetics_monitor_script":    resourceNewRelicSyntheticsMonitorScript(),
 			"newrelic_synthetics_label":             resourceNewRelicSyntheticsLabel(),
 			"newrelic_synthetics_secure_credential": resourceNewRelicSyntheticsSecureCredential(),
+			"newrelic_app_label":                    resourceNewRelicAppLabel(),
 		},
 	}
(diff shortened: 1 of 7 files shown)


jthurman42:
Thank you so much for the contribution!  We would like to see some testing around this before merging, is that something that you could provide as well?  [Here](https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/newrelic/resource_newrelic_synthetics_label_test.go) is an example of the Synthetics Labels test.

newrelic/resource_newrelic_app_label.go
```diff
@@ -0,0 +1,172 @@
+package newrelic
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/newrelic/newrelic-client-go/pkg/apm"
+	"github.com/newrelic/newrelic-client-go/pkg/errors"
+	"log"
+	"strings"
```
jthurman42: We try to organize imports with core first, then white space, then third-parties for clarity:

```
import (
  "log"
  "strings"

  "github.com/hashicorp/terraform-plugin-sdk/helper/schema"
  "github.com/newrelic/newrelic-client-go/pkg/apm"
  "github.com/newrelic/newrelic-client-go/pkg/errors"
)
```

website/docs/guides/getting_started.html.markdown
```diff
@@ -205,11 +205,35 @@ resource "newrelic_alert_channel" "slack" {
 This should avoid any of the configuration items from causing a change to the
 resource.
 
+## Add a Label to the Application
```
jthurman42: Adding this to the getting started guide is great.  Can you also add a specific resource document? Good example is the [Synthetics Label](https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/website/docs/r/synthetics_label.html.markdown) doc.

newrelic/resource_newrelic_app_label.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+func resourceNewRelicAppLabel() *schema.Resource {
```
jthurman42: It's a bit long but changing `App` to `Application` fits with the rest of the provider.

newrelic/provider.go
```diff
@@ -120,6 +120,7 @@ func Provider() terraform.ResourceProvider {
 			"newrelic_synthetics_monitor_script":    resourceNewRelicSyntheticsMonitorScript(),
 			"newrelic_synthetics_label":             resourceNewRelicSyntheticsLabel(),
 			"newrelic_synthetics_secure_credential": resourceNewRelicSyntheticsSecureCredential(),
+			"newrelic_app_label":                    resourceNewRelicAppLabel(),
```
jthurman42: Please update to `newrelic_application_label` to be consistent with the other resources, and sort within the map instead of appending to the end.

newrelic/resource_newrelic_app_label.go
lukaszczechowski: Reordered as suggested

website/docs/guides/getting_started.html.markdown
lukaszczechowski: Thanks for the comment. I removed it from Getting started and moved to the dedicated resource as suggested.

newrelic/resource_newrelic_app_label.go
lukaszczechowski: Renamed as suggested.

newrelic/provider.go
lukaszczechowski: Renamed and reordered as suggested.

lukaszczechowski:
@jthurman42 thank you for getting in touch so quickly. Thank you for the review I really appreciate it.
I applied the suggested changes and added some tests for the new resource. Unfortunately, I didn't have a chance to run them in some Acceptance Test environment. Hopefully this is sth which can be run on your CI.
I'm waiting for the feedback.

PR #431, opened 2020-03-02
docs: list of potential metrics for newrelic_alert_condition

harrykimpel:
I updated the docs page for newrelic_alert_condition and added the potential metrics for the corresponding type of condition


zlesnr:
Great, thank you for the contribution.

PR #447, opened 2020-03-06
removed the 'servers_metric' and updated the priority item

harrykimpel:


ctrombley:
Thanks @harrykimpel !

Issue #436, opened 2020-03-02
Add an NR1 workloads resource and data source


ctrombley:
Blocked by newrelic/newrelic-client-go/issues/153.

Add a resource and data source for managing NR1 workloads with the provider, following existing patterns.

PR #474, opened 2020-03-20
feat(workloads): add a workloads resource

ctrombley:
Resolves: #436 

This PR adds support for New Relic One workloads.

### DIFF
--- a/newrelic/resource_newrelic_workload.go
+++ b/newrelic/resource_newrelic_workload.go
@@ -125,7 +125,7 @@ func resourceNewRelicWorkload() *schema.Resource {
 			"permalink": {
 				Type:        schema.TypeString,
 				Computed:    true,
-				Description: "The URL of the Workload.",
+				Description: "The URL of the workload.",
 			},
 			"scope_accounts": {
 				Type:        schema.TypeList,
@@ -185,10 +185,7 @@ func resourceNewRelicWorkloadCreate(d *schema.ResourceData, meta interface{}) er
 
 	if len(workload.ScopeAccounts.AccountIDs) > 0 {
 		var scopeAccounts workloads.ScopeAccountsInput
-		scopeAccounts.AccountIDs = []int{}
-		for _, a := range workload.ScopeAccounts.AccountIDs {
-			scopeAccounts.AccountIDs = append(scopeAccounts.AccountIDs, a)
-		}
+		scopeAccounts.AccountIDs = append(scopeAccounts.AccountIDs, workload.ScopeAccounts.AccountIDs...)
 
 		createInput.ScopeAccountsInput = &scopeAccounts
 	}
@@ -263,10 +260,7 @@ func resourceNewRelicWorkloadUpdate(d *schema.ResourceData, meta interface{}) er
 
 	if len(workload.ScopeAccounts.AccountIDs) > 0 {
 		var scopeAccounts workloads.ScopeAccountsInput
-		scopeAccounts.AccountIDs = []int{}
-		for _, a := range workload.ScopeAccounts.AccountIDs {
-			scopeAccounts.AccountIDs = append(scopeAccounts.AccountIDs, a)
-		}
+		scopeAccounts.AccountIDs = append(scopeAccounts.AccountIDs, workload.ScopeAccounts.AccountIDs...)
 
 		updateInput.ScopeAccountsInput = &scopeAccounts
 	}
@@ -276,7 +270,7 @@ func resourceNewRelicWorkloadUpdate(d *schema.ResourceData, meta interface{}) er
 		return err
 	}
 
-	return resourceNewRelicSyntheticsMonitorRead(d, meta)
+	return resourceNewRelicWorkloadRead(d, meta)
 }
 
 func resourceNewRelicWorkloadDelete(d *schema.ResourceData, meta interface{}) error {
--- a/newrelic/resource_newrelic_workload.go
+++ b/newrelic/resource_newrelic_workload.go
@@ -1,0 +1,327 @@
+package newrelic
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/newrelic/newrelic-client-go/pkg/errors"
+	"github.com/newrelic/newrelic-client-go/pkg/workloads"
+)
+
+func resourceNewRelicWorkload() *schema.Resource {
+	userReference := &schema.Schema{
+		Type:     schema.TypeList,
+		MaxItems: 1,
+		Computed: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"email": {
+					Type:     schema.TypeString,
+					Computed: true,
+				},
+				"gravatar": {
+					Type:     schema.TypeString,
+					Computed: true,
+				},
+				"id": {
+					Type:     schema.TypeInt,
+					Computed: true,
+				},
+				"name": {
+					Type:     schema.TypeString,
+					Computed: true,
+				},
+			},
+		},
+	}
+
+	return &schema.Resource{
+		Create: resourceNewRelicWorkloadCreate,
+		Read:   resourceNewRelicWorkloadRead,
+		Update: resourceNewRelicWorkloadUpdate,
+		Delete: resourceNewRelicWorkloadDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"account": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
+				Computed:    true,
+				Description: "The account the workload belongs to.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+			"created_at": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "The moment when the object was created, represented in milliseconds since the Unix epoch.",
+			},
+			"created_by": userReference,
+			"entity": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Computed:    true,
+				Description: "A list of entities manually assigned to this workload.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"guid": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Computed:    true,
+							Description: "The unique entity identifier in New Relic.",
+						},
+					},
+				},
+			},
+			"entity_search_query": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Description: "A list of search queries that define a dynamic workload.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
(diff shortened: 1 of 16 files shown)


newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+		Schema: map[string]*schema.Schema{
+			"account": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
+				Computed:    true,
+				Description: "The account the workload belongs to.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
```
mbfrahry: Is `id` or `name` required here? You could add `ExactlyOneOf: []string{"account.0.id", "account.0.name"}` both here and in the `name` declaration to force that behavior.

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+			},
+			"created_by": userReference,
+			"entity": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Computed:    true,
+				Description: "A list of entities manually assigned to this workload.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"guid": {
```
mbfrahry: Are there any other attributes that will be added under this Set? If not we can simply this whole block to look something like 

```
"entity_guids": {
  Type:     schema.TypeSet,
  Optional: true,
  Elem: &schema.Schema{
  Type: schema.TypeString,
},
```

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The workload's name.",
+			},
+			"permalink": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The URL of the Workload.",
+			},
+			"scope_accounts": {
```
mbfrahry: Similar to the above, we should be able to simplify this to look like 
```
"storage_account_ids": {
  Type:     schema.TypeSet,
  Optional: true,
  Elem: &schema.Schema{
  Type: schema.TypeString,
},
```

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+		Update: resourceNewRelicWorkloadUpdate,
+		Delete: resourceNewRelicWorkloadDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"account": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
```
mbfrahry: Should this be Optional/Computed? It seems like it's necessary to build this [resource](https://github.com/terraform-providers/terraform-provider-newrelic/pull/474/files#diff-d8b554046f06d5baab974aeb013d168aR155). If this isn't required, we'll have to tune the logic in that link to reflect that.

newrelic/structures_newrelic_workload.go
```diff
@@ -0,0 +1,163 @@
...
+	if e, ok := d.GetOk("entity"); ok {
+		workload.Entities = expandWorkloadEntities(e.(*schema.Set).List())
+	}
+
+	if e, ok := d.GetOk("entity_search_query"); ok {
+		workload.EntitySearchQueries = expandWorkloadEntitySearchQueries(e.(*schema.Set).List())
+	}
+
+	if e, ok := d.GetOk("scope_accounts"); ok {
+		workload.ScopeAccounts = expandWorkloadScopeAccounts(e.([]interface{})[0].(map[string]interface{}))
```
mbfrahry: This could be nil so we should do a couple checks here
```
  if len(e) > 0 && e[0] != nil {
    workload.ScopeAccounts = expandWorkloadScopeAccounts(e.([]interface{})[0].(map[string]interface{}))
  }
```

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+		Name: workload.Name,
+	}
+
+	if len(workload.Entities) > 0 {
+		var entityGUIDs []string
+		for _, e := range workload.Entities {
+			entityGUIDs = append(entityGUIDs, *e.GUID)
+		}
+
+		createInput.EntityGUIDs = entityGUIDs
```
mbfrahry: It feels like we're doubling up the work here by first reading it into the workload struct and then copying those values into the createInput struct. Do you think it'd be better to just read the config straight into the inputStruct rather than going through another struct?

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+
+		updateInput.ScopeAccountsInput = &scopeAccounts
+	}
+
+	_, err = client.Workloads.UpdateWorkload(ids.GUID, updateInput)
+	if err != nil {
+		return err
+	}
+
```
mbfrahry: You'll want to call another d.SetId() after you finish updating the resource.

newrelic/structures_newrelic_workload.go
```diff
@@ -0,0 +1,163 @@
+package newrelic
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/newrelic/newrelic-client-go/pkg/nerdgraph"
+	"github.com/newrelic/newrelic-client-go/pkg/workloads"
+)
+
```
mbfrahry: Are the methods in this file going to be used in other resource? If not, I'd argue keeping all these contained in the same file just for maintainability?

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+		Name: &workload.Name,
+	}
+
+	if len(workload.Entities) > 0 {
+		var entityGUIDs []string
+		for _, e := range workload.Entities {
+			entityGUIDs = append(entityGUIDs, *e.GUID)
+		}
+
+		updateInput.EntityGUIDs = entityGUIDs
```
mbfrahry: This is similar to the doubling of work in the createInput struct comment

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+
+	return &schema.Resource{
+		Create: resourceNewRelicWorkloadCreate,
+		Read:   resourceNewRelicWorkloadRead,
+		Update: resourceNewRelicWorkloadUpdate,
+		Delete: resourceNewRelicWorkloadDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
```
mbfrahry: In terms of maintainability of the schema, you should try order the attributes that are Required at the top, Optional in the middle, and Computed at the bottom. This makes it fairly easy to scan the schema and know how things are prioritized down when creating the various structs/setting the values in state.

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/newrelic/newrelic-client-go/pkg/errors"
+	"github.com/newrelic/newrelic-client-go/pkg/workloads"
+)
+
+func resourceNewRelicWorkload() *schema.Resource {
+	userReference := &schema.Schema{
```
mbfrahry: I'd argue that this provides a little bit of confusion because it's been pulled out this way. I could see pulling this schema out if it had more nested values but I'd suggest leaving it in the main schema otherwise. I think it's only used in an attribute that we can remove but you can take this info and apply it going forward.

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		if _, ok := err.(*errors.NotFound); ok {
+			d.SetId("")
+			return nil
+		}
+
+		return err
+	}
+
+	return flattenWorkload(workload, d)
```
mbfrahry: I don't think calling out to another method to set our attributes provides much value. It just forces us to go down another rabbithole to find any potential issues

newrelic/structures_newrelic_workload.go
```diff
@@ -0,0 +1,163 @@
...
+func expandWorkload(d *schema.ResourceData) workloads.Workload {
+	workload := workloads.Workload{
+		Name: d.Get("name").(string),
+		ID:   d.Get("workload_id").(int),
+		GUID: d.Get("guid").(string),
```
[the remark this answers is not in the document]
ctrombley: 👍

newrelic/structures_newrelic_workload.go
```diff
@@ -0,0 +1,163 @@
...
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/newrelic/newrelic-client-go/pkg/nerdgraph"
+	"github.com/newrelic/newrelic-client-go/pkg/workloads"
+)
+
+func expandWorkload(d *schema.ResourceData) workloads.Workload {
+	workload := workloads.Workload{
+		Name: d.Get("name").(string),
+		ID:   d.Get("workload_id").(int),
```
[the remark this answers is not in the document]
ctrombley: 👍

newrelic/resource_newrelic_workload.go
ctrombley: 👍

newrelic/resource_newrelic_workload.go
ctrombley: 👍

newrelic/resource_newrelic_workload.go
ctrombley: 👍

newrelic/resource_newrelic_workload.go
ctrombley: 👍

newrelic/resource_newrelic_workload.go
ctrombley: 👍

newrelic/resource_newrelic_workload.go
ctrombley: 👍

newrelic/resource_newrelic_workload.go
ctrombley: 👍

newrelic/resource_newrelic_workload.go
```diff
@@ -0,0 +1,327 @@
...
+							Optional: true,
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+			"created_at": {
```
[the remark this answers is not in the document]
ctrombley: 👍

newrelic/resource_newrelic_workload.go
ctrombley: I've flattened this down to `account_id` which is required.

newrelic/structures_newrelic_workload.go
ctrombley: This is following the existing patterns we have in place elsewhere.  We'll consider a cross-cutting refactoring to simplify, but not as part of this PR.

newrelic/resource_newrelic_workload.go
ctrombley: This is following the existing patterns we have in place elsewhere.  We'll consider a cross-cutting refactoring to simplify, but not as part of this PR.

ctrombley:
Holding off on merging this as we may see some more API changes from the Workloads team.

Issue #52, opened 2018-10-18
support to update application settings


bharathkumarna:
We would need support to update certain parameters of apm application.

reference: https://rpm.newrelic.com/api/explore/applications/update

ctrombley:
This will require changes to the [underlying SDK](https://github.com/paultyng/go-newrelic/blob/master/api/applications.go).

jthurman42:
Upstream library now supports updates, unblocking this: https://github.com/newrelic/newrelic-client-go/blob/master/pkg/apm/applications.go#L110

zlesnr:
I've started this work, but based on the new client.  https://github.com/terraform-providers/terraform-provider-newrelic/pull/449 adds support for the new client.  v0.15.0 of the new client is ready for update.

PR #497, opened 2020-03-30
feat(application): implement newrelic_application resource

zlesnr:
resolves: #52

### DIFF
--- a/newrelic/resource_newrelic_application.go
+++ b/newrelic/resource_newrelic_application.go
@@ -1,0 +1,119 @@
+package newrelic
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/newrelic/newrelic-client-go/pkg/apm"
+)
+
+func resourceNewRelicApplication() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNewRelicApplicationCreate,
+		Read:   resourceNewRelicApplicationRead,
+		Update: resourceNewRelicApplicationUpdate,
+		Delete: resourceNewRelicApplicationDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"app_apdex_threshold": {
+				Type:     schema.TypeFloat,
+				Required: true,
+			},
+			"end_user_apdex_threshold": {
+				Type:     schema.TypeFloat,
+				Required: true,
+			},
+			"enable_real_user_monitoring": {
+				Type:     schema.TypeBool,
+				Required: true,
+			},
+		},
+	}
+}
+
+func resourceNewRelicApplicationCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ProviderConfig).NewClient
+
+	userApp := expandApplication(d)
+
+	listParams := apm.ListApplicationsParams{
+		Name: userApp.Name,
+	}
+
+	result, err := client.APM.ListApplications(&listParams)
+	if err != nil {
+		return err
+	}
+
+	if len(result) != 1 {
+		return fmt.Errorf("more/less than one result from query for %s", userApp.Name)
+	}
+
+	app := *result[0]
+
+	if app.Name != userApp.Name {
+		return fmt.Errorf("the result name %s does not match requested name %s", app.Name, userApp.Name)
+	}
+
+	d.SetId(strconv.Itoa(app.ID))
+
+	log.Printf("[INFO] Importing New Relic application %v", userApp.Name)
+	return resourceNewRelicApplicationUpdate(d, meta)
+}
+
+func resourceNewRelicApplicationRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ProviderConfig).NewClient
+
+	userApp := expandApplication(d)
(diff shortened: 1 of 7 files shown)


newrelic/resource_newrelic_application.go
```diff
@@ -0,0 +1,114 @@
...
+	app, err := client.APM.GetApplication(userApp.ID)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[INFO] Read found New Relic application %+v\n\n\n", app)
+
+	flattenApplication(app, d)
+
+	return nil
```
sanderblue: Should we return result of `flattenApplication(app, d)`, such as when an error occurs and is returned in that function?

newrelic/resource_newrelic_application_test.go
```diff
@@ -0,0 +1,71 @@
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckNewRelicApplicationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNewRelicApplicationConfig(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNewRelicApplicationExists(testAccExpectedApplicationName),
+					resource.TestCheckResourceAttr(
```
sanderblue: Fairly certain the **import** test handles verifying these values so no need to include them.

cc: @ctrombley

sanderblue: We can still keep `testAccCheckNewRelicApplicationExists` though. That's a more custom check.

newrelic/resource_newrelic_application_test.go
```diff
@@ -0,0 +1,71 @@
...
+resource "newrelic_application" "app" {
+	name = "%s"
+	app_apdex_threshold = "0.9"
+	end_user_apdex_threshold = "0.8"
+	enable_real_user_monitoring = true
+}
+`, testAccExpectedApplicationName)
+}
+
+func testAccCheckNewRelicApplicationExists(n string) resource.TestCheckFunc {
```
sanderblue: I think we want to make a call to the API via the client to ensure it exists. At least that's what we're doing in the other resources I believe. Should be relatively easy copy and paste and replace a few things.

newrelic/resource_newrelic_application.go
zlesnr: I've adjusted the `flattenApplication()` to return an error, and adjusted here to return.

newrelic/resource_newrelic_application_test.go
zlesnr: Have done.

newrelic/resource_newrelic_application_test.go
zlesnr: Have done.

newrelic/resource_newrelic_application.go
```diff
@@ -0,0 +1,112 @@
...
+
+	_, err := client.APM.UpdateApplication(userApp.ID, updateParams)
+	if err != nil {
+		return err
+	}
+
+	return resourceNewRelicApplicationRead(d, meta)
+}
+
+func resourceNewRelicApplicationDelete(d *schema.ResourceData, meta interface{}) error {
```
ctrombley: I'm not familiar with this pattern either, allowing delete to remove the resource from the state file but not actually deleting anything.  Is there a motivation behind this?

What is stopping us from actually calling the delete endpoint?

newrelic/resource_newrelic_application_test.go
```diff
@@ -0,0 +1,77 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckNewRelicApplicationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNewRelicApplicationConfig(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNewRelicApplicationExists(testAccExpectedApplicationName),
+				),
+			},
```
ctrombley: Can we add an update step here?

website/docs/d/application.html.markdown
```diff
@@ -6,6 +6,44 @@ description: |-
   Looks up the information about an application in New Relic.
 ---
 
+# Resource: newrelic_application
```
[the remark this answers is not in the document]
zlesnr: Ohh, I missed that bit in the organization.

newrelic/resource_newrelic_application.go
```diff
@@ -0,0 +1,112 @@
...
+			},
+			"enable_real_user_monitoring": {
+				Type:     schema.TypeBool,
+				Required: true,
+			},
+		},
+	}
+}
+
+func resourceNewRelicApplicationCreate(d *schema.ResourceData, meta interface{}) error {
```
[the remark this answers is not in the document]
zlesnr: There is an AWS resource that behaves similarly, where its created out of band, but then managed as part of the TF workflow.  I think it was the VPC gateway or maybe NAT gateway.

newrelic/resource_newrelic_application_test.go
zlesnr: Added.

website/docs/d/application.html.markdown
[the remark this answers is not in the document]
zlesnr: Moved.

newrelic/resource_newrelic_application.go
zlesnr: Handled out of band.

PR #504, opened 2020-04-03
feat(alerts): convert Alerts Policies to nerdgraph

zlesnr:
resolves: #463

### DIFF
--- a/newrelic/resource_newrelic_alert_policy.go
+++ b/newrelic/resource_newrelic_alert_policy.go
@@ -1,13 +1,15 @@
 package newrelic
 
 import (
+	"errors"
+	"fmt"
 	"log"
 	"strconv"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/newrelic/newrelic-client-go/newrelic"
-	"github.com/newrelic/newrelic-client-go/pkg/errors"
+	"github.com/newrelic/newrelic-client-go/pkg/alerts"
 )
 
 func resourceNewRelicAlertPolicy() *schema.Resource {
@@ -17,7 +19,7 @@ func resourceNewRelicAlertPolicy() *schema.Resource {
 		Update: resourceNewRelicAlertPolicyUpdate,
 		Delete: resourceNewRelicAlertPolicyDelete,
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			State: resourceImportStateWithMetadata(1, "account_id"),
 		},
 		Schema: map[string]*schema.Schema{
 			"name": {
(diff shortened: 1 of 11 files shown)


zlesnr:
Will use a function from https://github.com/terraform-providers/terraform-provider-newrelic/pull/487

newrelic/resource_newrelic_alert_policy.go
```diff
@@ -26,23 +28,31 @@ func resourceNewRelicAlertPolicy() *schema.Resource {
...
+			// 	Computed:    true,
+			// 	Description: "The time the policy was created.",
+			// 	Deprecated:  "Unavailable attribute in NerdGraph.",
+			// },
+			// "updated_at": {
+			// 	Type:        schema.TypeString,
+			// 	Computed:    true,
+			// 	Description: "The time the policy was last updated.",
+			// 	Deprecated:  "Unavailable attribute in NerdGraph.",
+			// },
```
sanderblue, line 55: Should we go ahead and delete these?

newrelic/resource_newrelic_alert_policy.go
```diff
@@ -73,82 +104,118 @@ func resourceNewRelicAlertPolicyCreate(d *schema.ResourceData, meta interface{})
...
+		accountID = d.Get("account_id").(int)
 
-	if err != nil {
-		if _, ok := err.(*errors.NotFound); ok {
-			d.SetId("")
-			return nil
-		}
+	} else if len(ids) == 2 {
+		policyID = ids[0]
+		accountID = ids[1]
```
sanderblue, line 145: Wondering if you can use the `selectAccountID()` helper function located in `resource_helpers.go` to avoid these conditionals. Since `policyID` should always be `ids[0]`, could use the helper function to grab the account ID.

newrelic/resource_newrelic_alert_policy_test.go
```diff
@@ -15,7 +15,7 @@ func TestAccNewRelicAlertPolicy_Basic(t *testing.T) {
 	resourceName := "newrelic_alert_policy.foo"
 	rName := acctest.RandString(5)
 
-	resource.ParallelTest(t, resource.TestCase{
+	resource.Test(t, resource.TestCase{
```
sanderblue, line 18: I _think_ these can use `resource.ParallelTest` again now that we figured out that preliminary oddity with that other test.

zlesnr: Ah yes, this was for testing.

newrelic/resource_newrelic_alert_policy.go
zlesnr: I'll take a look.

newrelic/resource_newrelic_alert_policy.go
zlesnr: I think yes.  I'll put up a PR to deprecate them on the 1.x before I forget.

zlesnr: Removed.  https://github.com/terraform-providers/terraform-provider-newrelic/pull/508 to deprecate.

newrelic/resource_newrelic_alert_policy.go
zlesnr: Done.

PR #505, opened 2020-04-06
docs(nrql_alert_condition): update docs to reflect version 2.0 changes

sanderblue:
This PR updates the `resource_nrql_alert_condition` documentation to reflect NerdGraph integration. 

It also includes v2.x docs prep

### DIFF
--- a/website/docs/r/nrql_alert_condition.html.markdown
+++ b/website/docs/r/nrql_alert_condition.html.markdown
@@ -219,7 +219,7 @@ $ terraform import newrelic_nrql_alert_condition.foo 538291:6789035:static
 $ terraform import newrelic_nrql_alert_condition.foo 538291:6789035:outlier
 ```
 
-~> **NOTE:** The value of `conditionType` in the composite ID must a valid condition type - `static`, `baseline`, or `outlier.`
+~> **NOTE:** The value of `conditionType` in the import composite ID must be a valid condition type - `static`, `baseline`, or `outlier.`
 
 The actual values for `policy_id` and `condition_id` can be retrieved from the following New Relic URL when viewing the NRQL alert condition you want to import:
 
--- a/website/docs/r/nrql_alert_condition.html.markdown
+++ b/website/docs/r/nrql_alert_condition.html.markdown
@@ -10,6 +10,10 @@ description: |-
 
 Use this resource to create and manage NRQL alert conditions in New Relic.
 
+-> **IMPORTANT!** Version 2.0.0 of the New Relic Terraform Provider introduces some [additional requirements](/docs/providers/newrelic/index.html) for configuring the provider.
+<br><br>
+Before upgrading to version 2.0.0 or later, it is recommended to upgrade to the most recent 1.x version of the provider (version 1.16.0) and ensure that your environment successfully runs `terraform plan` without unexpected changes.
+
 ## Example Usage
 
 ##### Type: `static` (default)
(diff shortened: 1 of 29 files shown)


website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -10,6 +10,10 @@ description: |-
 
 Use this resource to create and manage NRQL alert conditions in New Relic.
 
+-> **IMPORTANT!** Version 2.0.0 of the New Relic Terraform Provider introduces some [additional requirements](/docs/providers/newrelic/index.html) for configuring the provider.
+<br><br>
+Before upgrading to version 2.0.0 or later, it is recommended to upgrade to the most recent 1.x version of the provider (version 1.16.0) and ensure that your environment successfully runs `terraform plan` without unexpected changes.
```
sanderblue, line 15: 🤔💭 Should this live here? How do we want to handle this in resource documentation.

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -48,81 +60,166 @@ See additional [examples](#additional-examples).
 
 The following arguments are supported:
 
+- `account_id` - (Optional) The New Relic account ID of the account you wish to create the condition. Defaults to the account ID set in your environment variable `NEWRELIC_ACCOUNT_ID`.
+- `baseline_direction` - (Optional) The baseline direction of a _baseline_ NRQL alert condition. Valid values are: `lower_only`, `upper_and_lower`, `upper_only` (case insensitive).
```
sanderblue, line 64: 🤔💭 This is a case-insensitive, however, should we provide examples with the casing that matches NerdGraph (uppercase)? Or should we show examples with everything as lowercase. 

Note: We can only make this decision for new attributes. The old attributes must stay the same as before due to the `outlier` condition type.

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -48,81 +60,166 @@ See additional [examples](#additional-examples).
...
 - `nrql` - (Required) A NRQL query. See [NRQL](#nrql) below for details.
-- `value_function` - (Optional) Possible values are `single_value`, `sum`.
+- `term` - (Required) A list of terms for this condition. See [Terms](#terms) below for details.
+- `value_function` - (Optional) Possible values are `single_value`, `sum` (case insensitive). Defaults to `single_value`.
 - `expected_groups` - (Optional) Number of expected groups when using `outlier` detection.
 - `ignore_overlap` - (Optional) Whether to look for a convergence of groups when using `outlier` detection.
-- `violation_time_limit_seconds` - (Optional) Sets a time limit, in seconds, that will automatically force-close a long-lasting violation after the time limit you select.  Possible values are `3600`, `7200`, `14400`, `28800`, `43200`, and `86400`.
+- `violation_time_limit` - (Optional) Sets a time limit, in hours, that will automatically force-close a long-lasting violation after the time limit you select. Possible values are `ONE_HOUR`, `TWO_HOURS`, `FOUR_HOURS`, `EIGHT_HOURS`, `TWELVE_HOURS`, `TWENTY_FOUR_HOURS` (case insensitive).
```
sanderblue, line 76: See [comment](https://github.com/terraform-providers/terraform-provider-newrelic/pull/505/files#r405160195) above

website/docs/r/nrql_alert_condition.html.markdown
ctrombley: I like the idea of this showing up in the resources.

website/docs/r/nrql_alert_condition.html.markdown
ctrombley: I think this is fine.

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -48,81 +60,166 @@ See additional [examples](#additional-examples).
...
+    threshold_occurrences = "all"
+  }
+}
+```
+
 ##### Type: `outlier`
+
+Example outlier detection: Get notified if members of a group deviate by a key metric.
+
+-> **NOTE:** The `outlier` NQRL alert condition type currently does not support new schema attributes introduced in v2.0.0 of the New Relic Terraform provider.
```
ctrombley, line 161: This is a good example of why I think the new attributes should say something like "introduced in v2" or something similar. 

Either that or we should split off v1 docs into their own section of the docs site as discussed today.

sanderblue: I guess in v2.0 we wouldn't include an `outlier` example then?

sanderblue: Removed the HCL for the `outlier` example and provided a short message that directly links out to **v1.x** version of the `outlier` example.

PR #546, opened 2020-05-04
docs(getting started): fix resource naming

DarthVanger:
Rename "my_alert_policy_name" to "alert_policy_name" in the Getting Started guide.
Previously both names were used, resulting in errors
when trying to run terraform after pasting the code from docs.

`alert_policy_name` was used on line 105, while other lines used `my_alert_policy_name`.

Since no other resource names start with `my_`, I've decided to keep it as `alert_policy_name`.


jthurman42:

Issue #538, opened 2020-05-01
Provide (or release?) a way to set apdex_t for an existing new relic application


ryangardner:
I was about to pull down this provider and start fumbling my way through trying to implement this myself, when I came across: 

https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/newrelic/resource_newrelic_application.go

Which looks like it has the ability to do what I was needing. Is there a rough idea when the next release might go out that will include this feature? (If you accept bribes, what's your preferred way to be bribed?)

jthurman42:
There is a bit of work to get this into the provider, but there is an API endpoint to do so.  We're currently prepping to start a 2.x release cycle with the change required to enable this ability.

The specific endpoint in question is: https://api.newrelic.com/graphiql

with the following graphql query:

```
mutation {
  apmApplicationEntityChangeSettings(guid: "<MY APPLICATION ENTITY GUID>", settings: {apdexTarget: 0.5}) {
    entity {
      guid
      name
      settings {
        apdexTarget
      }
    }
  }
}
```

ryangardner:
Great, thanks for the reply!

We did see that graphql endpoint - (that currently only lets you set apdexTarget - but maybe someday will let you set other settings)

Is the current master branch getting ready for the 2.x release? We did build the current master version and what's inside there that lets you set the appdex and a couple other settings does seem to work (at least to set apdex): https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/newrelic/resource_newrelic_application.go#L22-L36

(and one piece of feedback on that specific resource, since it hasn't been released yet...  you might want to consider naming it `applicationSettings` or something since the version there almost makes it seem like it will _create_ a new `application` in new relic rather than just _modify an existing one_ )

jthurman42:
Looks like this was merged into master just after the split, but appears to be backwards compatible.  @zlesnr any reason not to port the application resource to v1.x?

Good point on the naming, as this won't create an application.  We overrode the 'create' function to act as an import.  Read/Update/Delete should all function (note the API requires apps to stop reporting for 12 hours before a delete can occur, so this is also slightly unexpected behavior)

zlesnr:
Not that I see @jthurman42.  Looks like its all REST there.

ctrombley:
[...]

PR #558, opened 2020-05-07
feat(application): implement newrelic_application resource

zlesnr:
resolves: #538


website/docs/r/application.html.markdown
```diff
@@ -0,0 +1,49 @@
...
+layout: "newrelic"
+page_title: "New Relic: newrelic_application"
+sidebar_current: "docs-newrelic-resource-application"
+description: |-
+  Manage configuration for an existing application in New Relic.
+---
+
+# Resource: newrelic_application
+
+Use this resource to manage configuration for an application that already exists in New Relic.
```
jthurman42, line 11: Should make a larger call-out that you can not Create a new application this way.

sanderblue: Could use one of the markdown callouts that TF supports...

`~>` for a beige background callout
`->` for a blue background callout 

There's also one for green, but can't remember what it is off the top of head.

zlesnr: Updated to include a brighter note.

zlesnr:
Fixes were made here: https://github.com/terraform-providers/terraform-provider-newrelic/pull/559. Will cherry pick after merge.

website/docs/r/application.html.markdown
```diff
@@ -0,0 +1,49 @@
...
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The ID of the application.
+
+## Import
+
+Applications can be imported using notation `application_id`, e.g.
```
sanderblue, line 41: 💭 Is `application_id` mentioned somewhere else? Thoughts on wording this something like "Applications can be imported using the application ID, e.g."

sanderblue: Oh I bet this is a NerdGraph field

jthurman42:
Changes went into master for the updated name, those need to be cherry picked back onto this release branch @zlesnr

Issue #476, opened 2020-03-23
add support for account_id in dashboard widget


echoAlexey:
Hi!

I want to build cross-account dashboard, this requires me to pass "account_id" to my widgets.
At the moment this is not supported. 

From New Relic doc it is possible to pass account_id:

Widget data element | Description
-- | --
account_id  			 | Source account to fetch data from, if not the current account.

https://docs.newrelic.com/docs/insights/insights-api/manage-dashboards/insights-dashboard-api#widget-data

sanderblue:
Hey there! Thanks for the reporting. Unfortunately this is not supported via the Dashboards API which has some limitations regarding cross-account widgets. The [requirements](https://docs.newrelic.com/docs/insights/insights-api/manage-dashboards/insights-dashboard-api#requirements) section of the docs provides a little more insight there. Essentially those types of widgets are not exported by the API.

Related: #154 (for a little more context 🙂)

echoAlexey:
@sanderblue  Hi! I understand that exporting cross-account widget is not supported, but having them created by passing "account_id" as param should be possible, isn't it? 
I see limitation which is preventing us to "view" it, but not making correct call to API with account id.

echoAlexey:
I don't see problem if terraform will have state with "visualization": "inaccessible" and run dashboard re-creation every time, but it will re-create dashboard with correct setup by passing "account_id" in the calls. 
So far terraform resource implementation is limiting us, not the API as you say. So it is not right to close this issue.

sanderblue:
Thanks for the followup. We were discussing the best coarse of action before reopening the ticket. Your points are valid and we'll explore this further 🙂

sanderblue:
Resolved in: #903 

**Update:** With some recent updates to the Dashboards API, we were able to implement this feature. It will go out with the next release. 😎

PR #561, opened 2020-05-08
Add support for cross account widgets

ianrayns:
Adding support for `account_id` attribute in dashboard widgets (#476) 

Unfortunately due to the API being unable to retrieve cross account widgets, this does mean that changes are detected every time.

Acceptance tests are difficult as I'm not sure if the account used by the CI server has access to other New Relic accounts. Happy to add these if required 👍

### DIFF
--- a/newrelic/resource_newrelic_dashboard.go
+++ b/newrelic/resource_newrelic_dashboard.go
@@ -263,6 +263,11 @@ func resourceNewRelicDashboard() *schema.Resource {
 							Elem:        &schema.Schema{Type: schema.TypeInt},
 							Description: "A collection of entity ids to display data for. These are typically application IDs.",
 						},
+						"account_id": {
+							Type:        schema.TypeInt,
+							Optional:    true,
+							Description: "Source account to fetch data from, if not the current account.",
+						},
 						"metric": {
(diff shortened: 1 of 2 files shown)


zlesnr:
Thank you for the PR.  We'll certainly want to test this behavior, and we can help with the CI if you want to send tests.  We can start there and see where we get.  There may be upstream complications.

emeshuris:
hi, are there any updates on this work?

zlesnr:
Not to my knowledge.  This functionality will need tests to confirm the behavior.

emeshuris:
is there any way to help? my company has a large amount of sub accounts and this is somewhat of an issue for us.

mason-jones-ck:
I'm glad to see this work, but it's been a while without any activity -- this is something that my company requires. We need to be able to specify account-id for dashboards, widgets, and eventually multiple queries within the same widget. Is someone working on testing and integrating this change (and perhaps figuring out how to specify multiple API keys in order to avoid seeing changes each time)?

zlesnr:
https://github.com/newrelic/terraform-provider-newrelic/pull/903 is in process now, and details are being worked out over there.

Issue #631, opened 2020-06-02
newrelic_nrql_alert_condition output; term as map instead of list


skyzyx:
### Feature Description

I'm writing a Terraform module, then writing integration tests for that module using [Terratest](https://terratest.gruntwork.io).

#### Module

I have some module code wherein I output an _object_ representing a resource.

```hcl
resource "newrelic_nrql_alert_condition" "cond_alb_5xx_errors" {
  nrql {
    query       = "FROM LoadBalancerSample SELECT average(provider.httpCodeTarget5XXCount.Sum) WHERE label.Service = '${local.tag_service}'"
    since_value = "3" # Evaluation offset
  }

  # WARN
  term {
    priority      = "warning"
    operator      = "above"
    threshold     = var.cond_alb_5xx_errors_warn
    time_function = "any" # at least once in
    duration      = var.cond_alb_5xx_errors_over_minutes
  }

  # CRIT
  term {
    priority      = "critical"
    operator      = "above"
    threshold     = var.cond_alb_5xx_errors_crit
    time_function = "any" # at least once in
    duration      = var.cond_alb_5xx_errors_over_minutes
  }

  # Snip...
}

output "cond_alb_5xx_errors" {
  value = newrelic_nrql_alert_condition.cond_alb_5xx_errors
}
```

#### Test + calling Terraform

I have a `test.tf` file and a `test.go` file. In the `test.tf` file, I call that module and run it. I then have to output data from this file to make it available to the Terratest code in `test.go` (essentially, one level of abstraction).

```hcl
module "nri_alerts" {
  source = "..."

  slack_channel_name = "..."
  slack_webhook_url  = "..."
  runbook_url        = "..."
  tags               = [...]
}

output "cond_alb_5xx_errors_nrql" {
  value = newrelic_nrql_alert_condition.cond_alb_5xx_errors.nrql
}

output "cond_alb_5xx_errors_term" {
  value = newrelic_nrql_alert_condition.cond_alb_5xx_errors.term
}
```

#### Reading the output

The data from the `nrql` node is an array of objects:

```
[
  {...}
]
```

The data from the `term` node is expected to be an array of objects, but Terraform calls it a "set":

```
[
  {...},
  {...}
]
```

Whereas I can access `nrql[0]`, I cannot access `term[0]`. Instead, I needed to convert the _set_ to a _list_.

```hcl
output "cond_alb_5xx_errors_term" {
[...]

PR #645, opened 2020-06-05
chore(alerts): deprecation notice for direct use of "terms"

zlesnr:
resolves: #631

The named condition priorities "warning" and "critical" are promoted to replace the use of "terms" on a `newrelic_nrql_alert_condition` resource.

### DIFF
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -13,6 +13,99 @@ import (
 	"github.com/newrelic/newrelic-client-go/pkg/errors"
 )
 
+// termSchema returns the schema used for a critial or warning term priority.
+func termSchema() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			// Maps to `thresholdDuration` in NerdGraph and values are in seconds, not minutes.
+			// Validation is different in NerdGraph - Value must be within 120-3600 seconds (2-60 minutes) and a multiple of 60 for BASELINE conditions.
+			// Convert to seconds when using NerdGraph
+			"duration": {
+				Deprecated:    "use `threshold_duration` attribute instead",
+				Type:          schema.TypeInt,
+				Optional:      true,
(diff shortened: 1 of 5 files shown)


newrelic/resource_newrelic_nrql_alert_condition.go
```diff
@@ -13,6 +13,102 @@ import (
...
+					return
+				},
+			},
+		},
+	}
+}
+
+func namedPrioritySchema() *schema.Resource {
+	rec := termSchema()
+	delete(rec.Schema, "priority")
```
sanderblue, line 107: Thoughts on removing `priority` from the base term schema instead? And then adding the attribute in `termSchema`? Feels a little more future facing.

zlesnr: Oh, as in add `priority` only where it's needed, rather than delete where it's not?

sanderblue: Maybe functions like 

```go
termSchema() {
  // The new hotness (moved cheese)
}

termSchemaDeprecated() {
   // add priority here
}
```

newrelic/structures_newrelic_nrql_alert_condition.go
```diff
@@ -260,9 +345,37 @@ func flattenNrqlAlertCondition(accountID int, condition *alerts.NrqlAlertConditi
...
+		}
+	} else {
+		// Handle the named condition priorities.
+		conditionTerms := flattenNrqlTerms(condition.Terms, configuredTerms)
+
+		for _, term := range conditionTerms {
+			switch term["priority"].(string) {
+			case "critical":
+				t := term
+				delete(t, "priority")
```
sanderblue, line 363: Kind of have a similar feeling to this as the other comment. Wondering if this could added to the deprecated scenario vs deleting it from the future facing setup.

zlesnr: This one might get a little squirrelly unless we duplicate a considerable chunk code.

sanderblue: Fair enough

PR #659, opened 2020-06-09
fix(alerts): avoid drift in nrql_alert_condition acceptance tests

zlesnr:


newrelic/resource_newrelic_nrql_alert_condition_test.go
```diff
@@ -157,6 +157,8 @@ func TestAccNewRelicNrqlAlertCondition_NerdGraphBaseline(t *testing.T) {
 				// we don't set deprecated fields on import
 				ImportStateVerifyIgnore: []string{
 					"term",
+					"critical",
+					"warning",
```
sanderblue, line 161: Fairly certain we shouldn't ignore these fields on import. I think we also need to make a test that uses those attributes (if we haven't already). Not sure how big of a lift it would be to set the new attributes on import instead of `term`.

newrelic/resource_newrelic_nrql_alert_condition_test.go
```diff
@@ -464,19 +468,19 @@ resource "newrelic_nrql_alert_condition" "foo" {
...
 		duration      = 2
 		operator      = "above"
 		priority      = "warning"
-		threshold     = 1.1
+		threshold     = 1.12
 		time_function = "any"
 	}
 
-	violation_time_limit_seconds = 3600
+	violation_time_limit = "EIGHT_HOURS"
```
sanderblue, line 483: I think we should probably stick with the deprecated `violation_time_limit_seconds` attribute for this test config since it facilitates testing configurations that contain the deprecated fields.

zlesnr: Ah yes, good call.

newrelic/resource_newrelic_nrql_alert_condition_test.go
zlesnr: I can dig a bit on that, shouldn't be too complex.

zlesnr:
https://github.com/terraform-providers/terraform-provider-newrelic/pull/645

Issue #517, opened 2020-04-13
Provider: Auth configuration updates for v2.0


sanderblue:
### Objective
As part of v2.x preparation, we'll be introducing some breaking changes to the provider config listed below. We also need to update the associated documentation.

**Breaking changes:**

- Require Personal API key
- Update the `NEWRELIC_*` environment variables to match the New Relic standardized naming convention of `NEW_RELIC_*` 
- Require Account ID
- Remove `insights_account_id` config and rebase on provider-level `account_id` config
- Update provider block docs

<br>

**Deprecate:**

- Admin API key -- choosing to skip this due to lack of readiness
- Ensure that resources that still require admin API key (e.g. synthetics monitors) will require its use

PR #664, opened 2020-06-10
chore: adhere to environment variable naming standard

zlesnr:
Here we replace instances of NEWRELIC_* variables with NEW_RELIC_* to meet the
New Relic standard.

Resolves: #517


zlesnr:
Pretty close here.  Will need to update the tests configuration blocks to include the account id, or double check that they are getting loaded from the environment.

newrelic/provider_test.go
```diff
@@ -117,16 +117,16 @@ func TestAccNewRelicProvider_Region(t *testing.T) {
 }
 
 func testAccPreCheck(t *testing.T) {
-	if v := os.Getenv("NEWRELIC_API_KEY"); v == "" {
-		t.Fatal("NEWRELIC_API_KEY must be set for acceptance tests")
+	if v := os.Getenv("NEW_RELIC_API_KEY"); v == "" {
+		t.Log("[WARN] NEW_RELIC_API_KEY has not been set for acceptance tests")
```
ctrombley, line 121: Why not `t.Fatal` here?

website/docs/index.html.markdown
```diff
@@ -74,14 +74,38 @@ resource "newrelic_alert_policy_channel" "alert_email" {
 
 The following arguments are supported:
 
-- `api_key` - (Required except for `newrelic_insights_event` resource) Your New Relic API key. The `NEWRELIC_API_KEY` environment variable can also be used.
-- `personal_api_key` - (Required only for the `newrelic_workload` resource) Your New Relic Personal API key. The `NEWRELIC_PERSONAL_API_KEY` environment variable can also be used.
+- `api_key` - (Required) Your New Relic Personal API key. The `NEW_RELIC_API_KEY` environment variable can also be used.
+- `admin_api_key` - (Required for many resources) Your New Relic Admin API key. The `NEW_RELIC_ADMIN_API_KEY` environment variable can also be used.
```
ctrombley, line 78: Maybe link to the compatibility table here?

website/docs/index.html.markdown
```diff
@@ -74,14 +74,38 @@ resource "newrelic_alert_policy_channel" "alert_email" {
...
+
+Please see the [latest provider configuration docs](/docs/providers/newrelic/guides/provider_configuration.html) for the current recommended configuration settings.
+
+### Update the environment
+
+All environment variables in use by the provider have been renamed.
+
+Replace all instances of environment variables named `NEWRELIC_*` with `NEW_RELIC_*`.
+
+`NEW_RELIC_PERSONAL_API_KEY` becomes `NEW_RELIC_API_KEY`.  Please note, `NEW_RELIC_API_KEY` is now a *Personal API Key*.  Set the `NEW_RELIC_ADMIN_API_KEY` variable using an *Admin API Key*.
```
ctrombley, line 98: I think it might be less confusing to word this as: 

`NEW_RELIC_PERSONAL_API_KEY` *has been replaced by* `NEW_RELIC_API_KEY`.

website/docs/index.html.markdown
```diff
@@ -74,14 +74,38 @@ resource "newrelic_alert_policy_channel" "alert_email" {
...
+
+### Configuration
```
ctrombley, line 100: How about:
```
### Configuring the `provider` block
```

website/docs/index.html.markdown
zlesnr: See https://github.com/terraform-providers/terraform-provider-newrelic/pull/665

newrelic/provider_test.go
zlesnr: Good call, just an oversight when moving things around.

website/docs/index.html.markdown
zlesnr: Updated to be a little smoother.

website/docs/index.html.markdown
zlesnr: Updated.  Keep the note at the end that says what the contents are supposed to be?

PR #674, opened 2020-06-15
Feat/entities

### LINKED ISSUE #673: Support for NR1 Entities

ctrombley:
Resolves #673 .

Adds a `newrelic_entity` data source. 

`newrelic_application` has been deprecated and any examples making use of `newrelic_application` have been updated to use the new data source.

### DIFF
--- a/newrelic/data_source_newrelic_entity.go
+++ b/newrelic/data_source_newrelic_entity.go
@@ -3,6 +3,7 @@ package newrelic
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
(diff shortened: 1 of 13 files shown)


newrelic/data_source_newrelic_entity.go
```diff
@@ -0,0 +1,166 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name of the entity in New Relic One.  The first entity matching this name for the given search parameters will be returned.",
+			},
+			"type": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				Description:  "The entity's type. Valid values are APPLICATION, DASHBOARD, HOST, MONITOR, and WORRKLOAD.",
+				ValidateFunc: validation.StringInSlice([]string{"APPLICATION", "DASHBOARD", "HOST", "MONITOR", "WORKLOAD"}, false),
```
sanderblue, line 26: Not necessarily a blocker, but for a more user-friendly approach, I think we've been trying to avoid case sensitivity and then handling the uppercase transformation for GraphQL types in the associated expand/flatten functions. That being said, case-sensitive validation up front does somewhat make things simpler to handle in the code.

newrelic/data_source_newrelic_entity.go
```diff
@@ -0,0 +1,166 @@
...
+
+	err = d.Set("application_id", e.ApplicationID)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func expandEntityTag(cfg []interface{}) *entities.TagValue {
```
sanderblue, line 148: Should we add unit tests for the expand/flatten functions?

website/docs/d/entity.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+resource "newrelic_alert_condition" "foo" {
+  policy_id = newrelic_alert_policy.foo.id
+
+  name        = "foo"
+  type        = "apm_app_metric"
+  entities    = [data.newrelic_application.app.application_id]
+  metric      = "apdex"
+  runbook_url = "https://www.example.com"
+
+  term {
```
sanderblue, line 39: Just a reminder that we will probably want to update the regular alert condition to use the `critical` and `warning` arguments like we did for the `nrql_alert_condition` resource.

newrelic/data_source_newrelic_entity.go
ctrombley: 👍

newrelic/data_source_newrelic_entity.go
ctrombley: 👍

website/docs/d/entity.html.markdown
ctrombley: 👍 but nothing to do here at the moment correct?

ctrombley: 👍 but nothing to do here at the moment correct?

sanderblue: Correct. I just wanted to mention it - partly so I better retain this as a reminder 😉

PR #685, opened 2020-06-17
chore(ci): enable integration tests on push

sanderblue:
This PR enables running our test suite, including integration tests, with GitHub Actions.


.github/workflows/test.yml
```diff
@@ -62,13 +62,13 @@ jobs:
           ${{ runner.os }}-go-
 
     - name: Unit Tests
-      run: make test-unit
+      run: make test
       env:
```
zlesnr, line 66: For the `make test` case, I think it would be good to remove all the environment variables.  This would simulate what happens in hashicorp's release process, so we don't get stuck at the last mile.

sanderblue: Interesting, I'll take a look.

PR #687, opened 2020-06-17
chore(tests): avoid empty NEW_RELIC_ACCOUNT_ID so make test passes

zlesnr:


zlesnr:
To reproduce what this is solving for, `unset NEW_RELIC_ACCOUNT_ID; make test`

zlesnr:
Refactored a bit.  Hows that look @ctrombley?

PR #695, opened 2020-06-19
docs: include information on pinning a version

zlesnr:


website/docs/index.html.markdown
```diff
@@ -96,6 +96,14 @@ directed at the newer provider version.
 -> <small>**Deprecation notice:** 2020-06-12<br> 
 -> **End of support:** 2020-01-15</small>
 
+If you wish to pin your environment to a specific release, you can do so with a `required_providers` statement in your Terraform manifest.
+
+```hcl
+required_providers {
+    newrelic = "~> 1.19.0"
```
sanderblue, line 103: We could also include an example of pinning the version with the provider block. Most of the use cases we’ve seen have the version in the provider block itself, especially with multiple instances of the provider.

zlesnr: Sounds reasonable to me.  Do you have an example?  I was just reading here: https://www.terraform.io/docs/configuration/providers.html#provider-versions

ctrombley: `required_providers` is in the `provider` block already.  I think maybe we're just not used to seeing it since it's new syntax?  Pinning looks different in [1.12](https://www.terraform.io/docs/configuration/providers.html#provider-versions) than it did in [1.11](https://www.terraform.io/docs/configuration-0-11/providers.html#provider-versions).

zlesnr: Maybe I just link to upstream docs for "more info" sort of thing?

PR #714, opened 2020-06-23
docs: fix broken v1 links

ctrombley:
This PR introduces a separate list of resources and data sources for v1 so we don't generate unnecessary links for it.


ctrombley:
Closing since we put versioned docs in place along with the repository transfer.

PR #723, opened 2020-06-24
docs(provider): add environment variables and schema attribute table

sanderblue:
This PR adds a table displaying the provider's environment variables, how they map to their associated schema attributes, and descriptions.

### DIFF
--- a/website/docs/guides/provider_configuration.markdown
+++ b/website/docs/guides/provider_configuration.markdown
@@ -41,15 +41,19 @@ provider "newrelic" {}
 
 #### Environment variables reference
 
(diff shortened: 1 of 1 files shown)


website/docs/guides/provider_configuration.markdown
```diff
@@ -83,7 +87,9 @@ provider "newrelic" {
 
 -> <small>**Note:** The provider supports ***one*** region per instance of the provider.</small>
 
-
+[account ID]: https://docs.newrelic.com/docs/accounts/install-new-relic/account-setup/account-id
```
ctrombley, line 90: Maybe make mention that some resource like nrql conditions can override the provider-level account ID?

PR #724, opened 2020-06-24
feat(alerts): new alerts_location_failure_condition resource

zlesnr:


zlesnr:
Will require https://github.com/newrelic/newrelic-client-go/pull/366

newrelic/provider.go
```diff
@@ -127,24 +127,25 @@ func Provider() terraform.ResourceProvider {
...
+			"newrelic_alert_policy":                      resourceNewRelicAlertPolicy(),
+			"newrelic_alert_policy_channel":              resourceNewRelicAlertPolicyChannel(),
+			"newrelic_application_settings":              resourceNewRelicApplicationSettings(),
+			"newrelic_dashboard":                         resourceNewRelicDashboard(),
+			"newrelic_entity_tags":                       resourceNewRelicEntityTags(),
+			"newrelic_events_to_metrics_rule":            resourceNewRelicEventsToMetricsRule(),
+			"newrelic_infra_alert_condition":             resourceNewRelicInfraAlertCondition(),
+			"newrelic_insights_event":                    resourceNewRelicInsightsEvent(),
+			"newrelic_nrql_alert_condition":              resourceNewRelicNrqlAlertCondition(),
+			"newrelic_alerts_location_failure_condition": resourceNewRelicAlertsLocationFailureCondition(),
```
jthurman42, line 141: This is a Synthetics specific feature, should probably have that in the resource name?

`newrelic_alerts_synthetics_multilocation_condition` ?

newrelic/resource_newrelic_alerts_location_failure_condition.go
```diff
@@ -0,0 +1,208 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Set whether to enable the alert condition. Defaults to true.",
+			},
+			"entities": {
+				Type:        schema.TypeList,
+				Required:    true,
+				MinItems:    1,
+				Description: "The GUIDs of the Synthetics monitors to alert on. Defaults to true.",
```
jthurman42, line 48: Defaults to True?

newrelic/resource_newrelic_alerts_location_failure_condition.go
```diff
@@ -0,0 +1,208 @@
...
+				Elem:        &schema.Schema{Type: schema.TypeString},
+			},
+			"terms": {
```
jthurman42, line 51: I thought we were moving away from `terms` in favor of having explicit `critical` and `warning` config blocks.

newrelic/provider.go
zlesnr: I'm good with that.  I'll rename.

newrelic/resource_newrelic_alerts_location_failure_condition.go
zlesnr: The tests will have me fix all this soon enough.

newrelic/resource_newrelic_alerts_location_failure_condition.go
zlesnr: Good point.  If there are only "warning" and "critical" terms, I think it makes sense to do what you're suggesting.

newrelic/provider.go
```diff
@@ -127,24 +127,25 @@ func Provider() terraform.ResourceProvider {
...
+			"newrelic_alert_policy":                              resourceNewRelicAlertPolicy(),
+			"newrelic_alert_policy_channel":                      resourceNewRelicAlertPolicyChannel(),
+			"newrelic_application_settings":                      resourceNewRelicApplicationSettings(),
+			"newrelic_dashboard":                                 resourceNewRelicDashboard(),
+			"newrelic_entity_tags":                               resourceNewRelicEntityTags(),
+			"newrelic_events_to_metrics_rule":                    resourceNewRelicEventsToMetricsRule(),
+			"newrelic_infra_alert_condition":                     resourceNewRelicInfraAlertCondition(),
+			"newrelic_insights_event":                            resourceNewRelicInsightsEvent(),
+			"newrelic_nrql_alert_condition":                      resourceNewRelicNrqlAlertCondition(),
+			"newrelic_alerts_synthetics_multilocation_condition": resourceNewRelicAlertsSyntheticsMultiLocationCondition(),
```
ctrombley, line 141: Can I suggest `newrelic_synthetics_multilocation_alert_condition` for better consistency with the other synthetics condition resource?  I know this is a different naming convention than the other projects we own, but if we are going to make the switch in the TF provider as well I would want to do it across the board rather than having multiple conventions.

zlesnr: Sure, I can rename, no worries.

zlesnr:
I'll rename the resource and also include some docs, and then we're getting pretty close.

newrelic/provider.go
zlesnr: Renamed as suggested.

Issue #625, opened 2020-06-01
Remove use of REST API for Insight Custom Events queries


ctrombley:
### Objective
There are Insights queries that occur during the tests for Insights Custom Events resource and data source that are still making use of the REST API.  These can be migrated over to Nerdgraph, eliminating the need for the `insights_query_key` config.

### Acceptance Criteria
- Custom Events resource and data source tests no longer make use of REST APIs 
- `insights_query_key` is no longer required

jthurman42:
The Query key, if configured, creates the go-insights query client [here](https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/newrelic/config.go#L130) though it is only used in a [test](https://github.com/terraform-providers/terraform-provider-newrelic/blob/master/newrelic/resource_newrelic_insights_event_test.go#L70)

Personal API keys can be used instead to query insights via NerdGraph and can either be exposed as a data type, or just used directly in the test .

PR #725, opened 2020-06-25
refactor(insights_event): migrate insights querying to nerdgraph nrql query (#625)

sanderblue:
Resolves: #625 

**Note:** This PR removes the `insights_query_key` attribute from the provider configuration schema.


newrelic/resource_newrelic_insights_event_test.go
```diff
@@ -67,18 +68,32 @@ func testAccCheckNewRelicInsightsEventExists(n string, nrqls []string) resource.
...
+				errs = append(errs, fmt.Sprintf("[Error] Insights event not found (likely due to async issue) - query: %v", nrql))
 			}
 		}
 
+		if len(errs) > 0 {
+			fmt.Printf("test case TestAccNewRelicInsightsEvent_Basic failed (likely due to asynchronous)")
+
+			return fmt.Errorf("%v", strings.Join(errs, "\n"))
+		}
+
```
jthurman42, line 96: We should remove this test since we do not expose the query method, there is no reason to test for it.  Testing that an event was inserted in an async pipeline is going to be problematic, and not add any additional information about the workings of the provider than what the status code return on the create provides.

PR #741, opened 2020-06-30
fix(nrql_alert_condition): use better term operator

zlesnr:
Without this change, the term operator incorrectly uses the word "equal" when
the API requires the use of "equals".  Here we make the necessary code, test,
and docs updates to ensure users are able to make use of the "equal" term
operator on their static conditions.

### DIFF
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -32,9 +32,12 @@ func termSchema() *schema.Resource {
 			"operator": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				Default:      "equal",
+				Default:      "equals",
 				Description:  "One of (above, below, equal). Defaults to 'equal'.",
-				ValidateFunc: validation.StringInSlice([]string{"above", "below", "equal"}, false),
+				ValidateFunc: validation.StringInSlice([]string{"above", "below", "equals"}, true),
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return strings.EqualFold(old, new)
+				},
 			},
 			"threshold": {
 				Type:         schema.TypeFloat,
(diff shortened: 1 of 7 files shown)


zlesnr:
Will require https://github.com/newrelic/newrelic-client-go/pull/369

newrelic/resource_newrelic_nrql_alert_condition.go
```diff
@@ -32,9 +32,9 @@ func termSchema() *schema.Resource {
 			"operator": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				Default:      "equal",
+				Default:      "equals",
 				Description:  "One of (above, below, equal). Defaults to 'equal'.",
-				ValidateFunc: validation.StringInSlice([]string{"above", "below", "equal"}, false),
+				ValidateFunc: validation.StringInSlice([]string{"above", "below", "equals"}, false),
```
sanderblue, line 37: Thoughts on ignoring casing sensitivity here? I think we've been doing that in most places to make things easier on the user.

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -95,7 +95,7 @@ NRQL alert conditions support up to two terms. At least one `term` must have `pr
 The `term` block the following arguments:
 
 - `duration` - (Required) In minutes, must be in the range of `1` to `120`, inclusive.
-- `operator` - (Optional) `above`, `below`, or `equal`. Defaults to `equal`. Note that when using a `type` of `outlier`, the only valid option here is `above`.
+- `operator` - (Optional) `above`, `below`, or `equals`. Defaults to `equals`. Note that when using a `type` of `outlier`, the only valid option here is `above`.
```
sanderblue, line 98: Could mention the case insensitivity here if we go that route.

newrelic/resource_newrelic_nrql_alert_condition.go
zlesnr: Good suggestion.  It looks like we already `ToUpper` in the expand, so including a `DiffSuppressFunc` here seems to do the trick.  I'll add that and push it up.

PR #769, opened 2020-07-10
feat: add a newrelic_account data source

### LINKED ISSUE #732: New Relic Account Data source
### Feature Description
Now that the 2.x provider supports multiple accounts for certain resources, there should be an easy way to retrieve an `account_id`.  A simple wrapper to fetch an account by name would be handy

### Describe Alternatives
Manually configure account_ids on resources

### Additional context

Something like:

```
data "newrelic_account" "staging" {
  name = "My Cool Staging Account Name"
}
```

| Data | Description
| ---- | ----- |
| id | Account ID |
| name | Account Name |

#### NerdGraph Query

We can either wrap this quickly in the provider itself, or add back to `newrelic-client-go`:

```
{
  actor {
    accounts(scope: IN_REGION) {
      id
      name
    }
  }
}
```

ctrombley:
Resolves #732.


newrelic/data_source_newrelic_account.go
```diff
@@ -0,0 +1,111 @@
...
+func dataSourceNewRelicAccount() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNewRelicAccountRead,
+		Schema: map[string]*schema.Schema{
+			"scope": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      string(accounts.RegionScopeTypes.IN_REGION),
+				Description:  `The scope of the account in New Relic.  Valid values are "global" and "in_region".  Defaults to "in_region".`,
+				ValidateFunc: validation.StringInSlice([]string{"global", "in_region"}, true),
```
sanderblue, line 22: If we ignore casing here, does the API have any issues with different casing?

newrelic/data_source_newrelic_account.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+	}
+}
+
+func dataSourceNewRelicAccountRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ProviderConfig).NewClient
+
+	log.Printf("[INFO] Reading New Relic accounts")
+
+	scope := accounts.RegionScope(d.Get("scope").(string))
```
sanderblue, line 43: Just a followup regarding the casing questions above. Do we need to handle casing here?

Issue #770, opened 2020-07-13
Condition_scope will be included all the time you run terraform plan


MadhuraDhanajaya:
### Terraform Version
Terraform version = 0.12.24
Provider version = 1.16.0

### Affected Resource(s)
**newrelic_alert_condition**

### Terraform Configuration Files
```
resource "newrelic_alert_condition" "alert_conditions" {
  name     = condition_name
  type     =  each.value["metric_type"]
  entities = [local.entity_ids_by_service_names[each.value["service"]]]
  metric   = each.value["metric"]

  condition_scope = "instance"

  gc_metric             = each.value["gc_metric"]
  policy_id             =  policy_id 
  each.value["metric_type"])]
  violation_close_timer = each.value["violation_close_timer"]

  term {
    operator          = above
    threshold         = 95
    priority            = critical
    time_function  = all
    duration          = 5 
  }

```

### Actual Behaviour
  ``` 
module.alert_policies.newrelic_alert_condition.alert_conditions["application_name_heap_memory_usage"] will be updated in-place
  ~ resource "newrelic_alert_condition" "alert_conditions" {
      + condition_scope       = "instance"
        enabled               = true
        entities              = [
           122222,
        ]
        id                    = "*****"
        metric                = "heap_memory_usage"
        name                  = "abc_heap_memory_usage"
        policy_id             = ****
        type                  = "apm_jvm_metric"
        violation_close_timer = 24

        term {
            duration      = 5
            operator      = "above"
            priority      = "critical"
            threshold     = 95
            time_function = "all"
        }
    }

Plan: 0 to add, 478 to change, 0 to destroy.
```
### Expected Behavior
On running terraform plan or terraform apply only newly added resource or change in any resource has to be modified but here the **condition scope** which is set to **instance**  will get updated all the time we run `plan` and `apply` which has no changes in it. 

Result should be  **Plan: 0 to add, 0 to modify, 0 to destroy.**  

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform plan`
2. `terraform apply`

MadhuraDhanajaya:
Any updates on this?

sanderblue:
Hi @MadhuraDhanajaya, I've tried to reproduce this but so far I haven't been able to yet. Can you share an example config without variables or loops so we have a more specific scenario to debug? If we could scope to a single alert condition it might help with debugging. 

[...]

PR #774, opened 2020-07-16
fix(alert_condition): [v1.x] avoid condition_scope drift when the API doesn't return that field

sanderblue:
Resolves: #770


sanderblue:
Closing as this was an issue reported in v2.x

Issue #386, opened 2020-02-07
Upgrade to 1.13.0 breaks newrelic_alert_condition resource where condition_scope was not specified


pickgr:
### Terraform Version
Terraform v0.11.11
+ provider.newrelic v1.13.0

### Affected Resource(s)
newrelic_alert_condition

### Terraform Configuration Files
I think this can be reproduced with the example at
https://www.terraform.io/docs/providers/newrelic/r/alert_condition.html

### Expected Behavior
Before version 1.13.0, https://www.terraform.io/docs/providers/newrelic/r/alert_condition.html was optional.  When upgrading the provider, it must be added to the code.  However, after running apply, followed by a subsequent plan, the resource still shows drift for condition_scope

### Actual Behavior
There should be no drift after adding condition_scope to the code and updating the provider.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform plan` with 1.12.2 verify no drift
2. `terraform plan` with 1.13.0 - error regarding conditional_scope being mandatory
3.  specify conditional_scope in the resource
4. `terraform apply`
5. `terraform plan` observe that the resource still shows drift

Reverting to 1.12.2 is the only work-around I could find.

### References
Most likely related to these changes

https://github.com/terraform-providers/terraform-provider-newrelic/compare/v1.12.2...v1.13.0#diff-59adf773b6a5d1a8d428c72e2986db6aR121

zlesnr:
Hi @pickgr, I'm unable to reproduce the issue here using the manifest in the `alert_condition` documentation you linked.  Too, I'm not able to create the resource at all using either version of the provider unless the `condition_scope` is set.  I get a 422 from the server.

```
Error: 422 response returned: `condition_scope` must be `application` or `instance`.

  on test.tf line 11, in resource "newrelic_alert_condition" "foo":
  11: resource "newrelic_alert_condition" "foo" {
```
I can confirm though, that creating the resource on an older version results in an empty plan on both the old and newer versions.

When you revert to the old version, do you also keep the `condition_scope` set in the manifest?  Any more detail on your manifest would be helpful for me to try and reproduce this.

pickgr:
@zlesnr 

Hopefully this will help

Original code:

```
resource "newrelic_alert_condition" "livedoc_generation_response_time" {
  entities = [
    data.newrelic_key_transaction.livedoc_generation_prod_westus.id,
[...]

PR #775, opened 2020-07-17
fix(nrql_alert_condition): handle zero-value for threshold

sanderblue:
Depends on [newrelic-client-go #386](https://github.com/newrelic/newrelic-client-go/pull/386)


ctrombley:
Duplicate of #769

PR #782, opened 2020-07-21
chore: deprecate synthetics labels

ctrombley:


jthurman42:
@ctrombley should we disable the Synthetics labels tests here as well, since they no longer function on our test account?

jthurman42:
Nevermind, it's something else that's failing...

ctrombley:
> @ctrombley should we disable the Synthetics labels tests here as well, since they no longer function on our test account?

Those got deleted as part of this PR.

PR #800, opened 2020-07-28
fix(newrelic_alert_condition): allow instance scope for JVM app metrics

### LINKED ISSUE #795: Unable to create jvm metric alert for condition scope application
### Terraform Version
Terraform v0.12.29
+ provider.newrelic v2.3.0

### Affected Resource(s)
newrelic_alert_condition

### Terraform Configuration Files
```hcl
resource "newrelic_alert_condition" "heap_memory_high" {
  policy_id = newrelic_alert_policy.critical_alerts_policy.id
  name = "Heap Memory usage (High)"
  type = "apm_jvm_metric"
  entities = [data.newrelic_entity.css_useast.application_id]
  metric = "heap_memory_usage"
  condition_scope = "application"
// violation_close_timer = 24
  term {
    duration = 5
    operator = "above"
    priority = "critical"
    threshold = 90
    time_function = "all"
  }
}
```

### Debug Output
Not relevant

### Panic Output
No panic

### Expected Behavior
An alert should have been created

### Actual Behavior
An error is returned. Without violation_close_timer it says:
Error: 422 response returned: `violation_close_timer` (in hours) must be present and one of: 1, 2, 4, 8, 12, 24.

When I add this, I get 
Error: violation_close_timer only supported when condition_scope = 'instance'

That is actually not true when using the NewRelic webui. It allows me to set it when scope is application.
### Steps to Reproduce

1. `terraform apply`

### Important Factoids
Seems to be an issue with apm_jvm_metric alerts, using apm_app_metric as type works fine. 

### References
Couldn't find any.

ctrombley:
Resolves #795.

Scoping this validation to `apm_app_metric` condition type only so it doesn't affect users of `jvm_app_metric`. 

This also uncovered a state flattening bug for conditions of type `apm_jvm_metric` that could cause drift.  The state flattening required some additional logic since the underlying API fails to return the `condition_scope` field when that condition type is provisioned.

https://docs.newrelic.com/docs/alerts/rest-api-alerts/new-relic-alerts-rest-api/alerts-conditions-api-field-names#violation_close_timer

### DIFF
--- a/newrelic/structures_newrelic_alert_condition.go
+++ b/newrelic/structures_newrelic_alert_condition.go
@@ -22,12 +22,11 @@ func expandAlertCondition(d *schema.ResourceData) (*alerts.Condition, error) {
 	condition.Terms = expandAlertConditionTerms(d.Get("term").(*schema.Set).List())
 
 	if violationCloseTimer, ok := d.GetOk("violation_close_timer"); ok {
-		if condition.Type == "apm_app_metric" && condition.Scope == "instance" {
-			condition.ViolationCloseTimer = violationCloseTimer.(int)
-		} else {
+		if condition.Type == "apm_app_metric" && condition.Scope == "application" {
 			return nil, fmt.Errorf("violation_close_timer only supported for apm_app_metric when condition_scope = 'instance'")
 		}
 
+		condition.ViolationCloseTimer = violationCloseTimer.(int)
 	}
 
 	if attr, ok := d.GetOk("runbook_url"); ok {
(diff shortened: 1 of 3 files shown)


newrelic/structures_newrelic_alert_condition.go
```diff
@@ -22,10 +22,10 @@ func expandAlertCondition(d *schema.ResourceData) (*alerts.Condition, error) {
 	condition.Terms = expandAlertConditionTerms(d.Get("term").(*schema.Set).List())
 
 	if violationCloseTimer, ok := d.GetOk("violation_close_timer"); ok {
-		if condition.Scope == "instance" {
+		if condition.Type == "apm_app_metric" && condition.Scope == "instance" {
```
gjkamstra, line 25: I don't think this fixes my issue, when condition.Type == "apm_jvm_metric", the if at L25 is false, so we go to the else and we still get an error.
The error should not be raised when condition.Type == "apm_jvm_metric"

gjkamstra: If you change this the if statement to:
if condition.Type != "apm_app_metric" || (condition.Type == "apm_app_metric" && condition.Scope == "instance")

ctrombley: @gjkamstra You are 💯 correct, I have another commit queued up that looks pretty similar.  We had to provision some Java apps reporting to our test account to put integration tests in place, so this is taking a little longer than expected.  Thanks for your patience while we shave our yaks! 🐢

Issue #758, opened 2020-07-08
Data source for Synthetics monitor private locations


ctrombley:
Create a data source to expose the internal IDs of Synthetics private location monitor to be used with Synthetics alert conditions.

We should be able to model this resource after another simple data source such as [`newrelic_key_transaction`](https://github.com/newrelic/terraform-provider-newrelic/blob/master/newrelic/data_source_newrelic_key_transaction.go).

The New Relic Go client can be used to list the valid monitor locations via the [`Synthetics.GetMonitorLocations`](https://github.com/newrelic/newrelic-client-go/blob/master/pkg/synthetics/monitor_locations.go#L13) method.  This client method was previously implemented based on the underlying documentation for the API endpoint here: https://docs.newrelic.com/docs/apis/synthetics-rest-api/monitor-examples/manage-synthetics-monitors-rest-api#list-locations

The [Extending Terraform docs](https://www.terraform.io/docs/extend/writing-custom-providers.html) are a good place to dive deeper on resource and data source development in general.

ctrombley:
Resolved by #806

PR #806, opened 2020-07-29
feat(synthetics_monitor_location): add data source newrelic_synthetics_monitor_location

SowmyaJujjuri:
Resolves: #758

### DIFF
--- a/newrelic/data_source_newrelic_synthetics_monitor_location.go
+++ b/newrelic/data_source_newrelic_synthetics_monitor_location.go
@@ -12,31 +12,31 @@ func dataSourceNewRelicSyntheticsMonitorLocation() *schema.Resource {
 	return &schema.Resource{
 		Read: dataSourceNewRelicSyntheticsMonitorLocationRead,
 
-		Schema: map[string]*schema.Schema{ //required, optional, computed
+		Schema: map[string]*schema.Schema{
 			"label": {
 				Type:        schema.TypeString,
 				Required:    true,
-				Description: "The label of the synthetics monitor.",
+				Description: "The label of the Synthetics monitor location.",
 			},
 			"name": {
 				Type:        schema.TypeString,
 				Computed:    true,
-				Description: "The name of the synthetics monitor in New Relic.",
+				Description: "The name of the Synthetics monitor location.",
 			},
-			"highSecurityMode": {
+			"high_security_mode": {
 				Type:        schema.TypeBool,
 				Computed:    true,
-				Description: "",
+				Description: "The high security mode for the Synthetics monitor location",
 			},
 			"private": {
 				Type:        schema.TypeBool,
 				Computed:    true,
-				Description: "",
+				Description: "The private setting for the Synthetics monitor location",
 			},
 			"description": {
 				Type:        schema.TypeString,
 				Computed:    true,
-				Description: "",
+				Description: "The description mode for the Synthetics monitor location",
 			},
 		},
 	}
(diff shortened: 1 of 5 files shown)


newrelic/data_source_newrelic_synthetics_monitor_location.go
```diff
@@ -0,0 +1,77 @@
...
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The name of the Synthetics monitor location.",
+			},
+			"high_security_mode": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "The high security mode for the Synthetics monitor location",
```
ctrombley, line 29: Can we go with the following here:

"Represents if high security mode is enabled for the location.  A value of true means that high security mode is enabled, and a value of false means it is disabled."

newrelic/data_source_newrelic_synthetics_monitor_location.go
```diff
@@ -0,0 +1,77 @@
...
+			},
+			"private": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "The private setting for the Synthetics monitor location",
```
ctrombley, line 34: Can we go with the following here:

"Represents if this location is a private location. A value of true means that the location is private, and a value of false means it is public."

newrelic/data_source_newrelic_synthetics_monitor_location.go
```diff
@@ -0,0 +1,77 @@
...
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The description mode for the Synthetics monitor location",
```
ctrombley, line 39: Can we go with the following here:

"A description of the Synthetics monitor location."

newrelic/data_source_newrelic_synthetics_monitor_location.go
```diff
@@ -0,0 +1,77 @@
...
+			},
+		},
+	}
+}
+
+func dataSourceNewRelicSyntheticsMonitorLocationRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ProviderConfig).NewClient
+
+	log.Printf("[INFO] Reading Synthetics monitors location")
```
ctrombley, line 48: nit: Can we swap "monitors location" for "monitor locations"?

website/docs/d/synthetics_monitor_location.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+layout: "newrelic"
+page_title: "New Relic: newrelic_synthetics_monitor_location"
+sidebar_current: "docs-newrelic-datasource-synthetics-monitor-location"
+description: |-
+  Grabs a Synthetics monitor location by label.
+---
+
+# Data Source: newrelic\_synthetics\_monitor
+
+Use this data source to get information about a specific Synthetics monitor locations in New Relic that already exist. This can be used to set up a Synthetics alert condition.
```
ctrombley, line 11: nit: can we swap "locations" for "location" here, and remove "This can be used to set up a Synthetics alert condition"?

website/docs/d/synthetics_monitor_location.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+
+## Example Usage
+
+```hcl
+data "newrelic_synthetics_monitor" "bar" {
```
ctrombley, line 16: We'll want to replace the example here that is based on alert conditions with one that represents how this might be used in the wild, where the data source can be used to supply a `newrelic_synthetics_monitor` resource with a value for the `locations` attriubute.   Probably something like:

```
data "newrelic_synthetics_monitor_location" "bar" {
  label = "My private location"
}

resource "newrelic_synthetics_monitor" "foo" {
  name = "foo"
  type = "SIMPLE"
  frequency = 5
  status = "ENABLED"
  locations = [data.newrelic_synthetics_monitor_location.bar.name]

  uri                       = "https://example.com"               # Required for type "SIMPLE" and "BROWSER"
  validation_string         = "add example validation check here" # Optional for type "SIMPLE" and "BROWSER"
  verify_ssl                = true                                # Optional for type "SIMPLE" and "BROWSER"
}
```

Issue #763, opened 2020-07-09
TF plan re-creates slack notification channel even when there are no changes


hgibsonqb:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://discuss.newrelic.com/c/build-on-new-relic/developer-toolkit

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

```
Terraform v0.12.26
+ provider.newrelic v2.1.2
```

### Affected Resource(s)
Please list the resources as a list, for example:
- newrelic_alert_channel (slack)

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "newrelic_alert_channel" "apf_slack" {
  name = "apf_slack"
  type = "slack"

  config {
    url     = var.slack_webhook_url # These config values don't change
    channel = var.slack_channel
  }
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

Terraform plan output
```
 # module.apf_notification_channels.newrelic_alert_channel.apf_slack must be replaced
-/+ resource "newrelic_alert_channel" "apf_slack" {
      ~ id   = "3720481" -> (known after apply)
        name = "apf_slack"
        type = "slack"

      ~ config {
            channel = "appfactory-alerts"
          - headers = (sensitive value)
          - payload = (sensitive value)
          + url     = (sensitive value)
        }
    }

  # module.apf_stg_external_access_monitor.newrelic_alert_policy.apf_monitor_policy must be replaced
-/+ resource "newrelic_alert_policy" "apf_monitor_policy" {
      - account_id          = 2732602 -> null
      ~ channel_ids         = [
          - 3720481,
          - 3720018,
        ] -> (known after apply) # forces replacement
      ~ id                  = "951482" -> (known after apply)
        incident_preference = "PER_CONDITION"
        name                = "apf_monitor_policy-apf-stg"
    }
```

The channel id that it says will be generated/known after apply doesn't change.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
[...]

Issue #420, opened 2020-02-26
Eliminate drift in `newrelic_alert_channel` resources


ctrombley:
The alert channel API endpoint fails to return resource fields that it deems sensitive, which currently causes drift when using affected channel configurations.  In the past, we have recommended using the `lifecycle` block with an `ignore_changes` attribute as described in the [getting started guide](https://www.terraform.io/docs/providers/newrelic/guides/getting_started.html) to eliminate drift.

In situations where the underlying API fails to return a resource field that has been supplied in the config, a pattern that can be used to avoid drift is to simply not set the sensitive field when flattening the API response.  This pattern has been used to implement the `newrelic_synthetics_secure_credential` resource and it should be able to eliminate the drift in the `newrelic_alert_channel` resource as well.  We should experiment with this pattern and see if there is a way to eliminate the drift associated with sensitive data in a backward-compatible manner.

sanderblue:
Another thing that could help with detecting drift is making sure we call `READ` at the end of all modifications to synchronize state. We don't currently do that in the `newrelic_alert_channel` resource.

Reference: [Detecting Drift](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#update-state-after-modification)
> ... it is standard practice to call READ at the end of all modifications to synchronize immediately and avoid that diff.

We have several other resources that need to be updated to use this pattern as well.

davidji99:
It seems importing `newrelic_alert_channel` also doesn't set any of the `config` attributes. My state file for this resource looks like this:
```
"attributes": {
"config": [
{
"api_key": "",
"auth_password": "",
"auth_type": "",
"auth_username": "",
"base_url": "",
"channel": "",
"headers": {},
"headers_string": "",
"include_json_attachment": "",
"key": "",
"payload": {},
"payload_string": "",
"payload_type": "",
"recipients": "",
"region": "",
"route_key": "",
"service_key": "",
"tags": "",
"teams": "",
"url": "",
"user_id": ""
}
],
"configuration": null,
"id": "<SOME_ID>",
"name": "<SOME_NAME>,
"type": "<SOME_TYPE>"
},
```

On a subsequent `plan`, the diff shows a destroy and recreate:
```
  # newrelic_alert_channel.pagerduty-page-oncall must be replaced
-/+ resource "newrelic_alert_channel" "pagerduty" {
      ~ id   = "<SOME_ID>" -> (known after apply)
        name = "<SOME_NAME>"
[...]

PR #814, opened 2020-07-30
fix(alert_channel): avoid config drift

sanderblue:
Resolves: #763 
Resolves: #420

### DIFF
--- a/newrelic/resource_newrelic_alert_channel.go
+++ b/newrelic/resource_newrelic_alert_channel.go
@@ -260,7 +260,7 @@ func resourceNewRelicAlertChannelCreate(d *schema.ResourceData, meta interface{}
 
 	d.SetId(strconv.Itoa(channel.ID))
 
-	return nil
+	return resourceNewRelicAlertChannelRead(d, meta)
 }
 
 func resourceNewRelicAlertChannelRead(d *schema.ResourceData, meta interface{}) error {
--- a/newrelic/structures_newrelic_alert_channel.go
+++ b/newrelic/structures_newrelic_alert_channel.go
@@ -192,7 +192,6 @@ func flattenAlertChannelConfiguration(c *alerts.ChannelConfiguration, d *schema.
 
 	configResult := make(map[string]interface{})
 
-	configResult["auth_password"] = c.AuthPassword
 	configResult["auth_username"] = c.AuthUsername
 	configResult["base_url"] = c.BaseURL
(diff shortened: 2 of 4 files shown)


newrelic/structures_newrelic_alert_channel.go
```diff
@@ -203,12 +202,42 @@ func flattenAlertChannelConfiguration(c *alerts.ChannelConfiguration, d *schema.
 	configResult["recipients"] = c.Recipients
 	configResult["region"] = c.Region
 	configResult["route_key"] = c.RouteKey
-	configResult["service_key"] = c.ServiceKey
 	configResult["tags"] = c.Tags
 	configResult["teams"] = c.Teams
-	configResult["url"] = c.URL
 	configResult["user_id"] = c.UserID
 
+	if attr, ok := d.GetOk("config.0.api_key"); ok {
```
ctrombley, line 209: There are 12 sensitive schema values in the config.  Is this going to cover all cases?

sanderblue: I was hoping you'd chime in here. I don't know to be honest, but I also don't want to make too drastic of a change and cause cascading bugs somewhere else. I also [asked additional config examples](https://github.com/newrelic/terraform-provider-newrelic/issues/420#issuecomment-666807402) from some folks in hopes of getting more scenarios covered.

I'll continue to investigate

sanderblue: So far we're covered for fields the API considers sensitive, which are `api_key`, `auth_password`, `auth_type`, `route_key`, and `service_key` (I did a diff based on a full POST channel config compared to the GET response config I received). 

We still have some extreme outlier scenarios for `headers`/`headers_string` and `payload`/`payload_string`, but it seems to mostly involve detecting import state and conditionally handling that. I think we can address that in another PR if we deem this worthy.

Bottom line here is that this PR makes things a good bit better than they were regarding the config drift 😎

newrelic/resource_newrelic_alert_channel.go
```diff
@@ -260,7 +260,7 @@ func resourceNewRelicAlertChannelCreate(d *schema.ResourceData, meta interface{}
 
 	d.SetId(strconv.Itoa(channel.ID))
 
-	return nil
+	return resourceNewRelicAlertChannelRead(d, meta)
```
zlesnr, line 263: So much love for this.

Issue #821, opened 2020-07-31
Standardize `account_id` default behavior across all applicable resources (i.e. NerdGraph resources)


sanderblue:
### Objective
Now that we have a stable pattern of "inheriting" the `account_id` attribute value from the provider config block within resources and data sources, we should standardize how the provider handles this for all resources and data sources that leverage the NerdGraph API.

Essentially all `account_id` resource and data source schema attributes should look similar to this (más o menos)
```
"account_id": {
		Type:        schema.TypeInt,
		Optional:    true,
		Computed:    true,  <-- key ingredient
		Description: "The New Relic account ID to operate on.",
},
```

Use this [pull request](https://github.com/newrelic/terraform-provider-newrelic/pull/820) as a reference. 

Applicable resources for this adjustment:
- 

### Acceptance Criteria
- Ensure no configuration drift occurs in the resources that get updated.
- Tests pass 

**Note:** Acceptance tests are great, but for this scenario, it's a good idea to also try testing this out manually as if you were a real users - i.e. making a `main.tf` file and including the resources that you're testing out. You'll also want to make sure to unset your `NEW_RELIC_ACCOUNT_ID` environment variable before testing it out manually. You'll run `unset NEW_RELIC_ACCOUNT_ID; terraform apply` and then `unset NEW_RELIC_ACCOUNT_ID; terraform plan` where the plan step should result is 0 changes.

PR #824, opened 2020-08-03
chore(account_id): standardize "account_id" attribute behavior

zlesnr:
resolves: #821

### DIFF
--- a/newrelic/resource_newrelic_events_to_metrics_rule.go
+++ b/newrelic/resource_newrelic_events_to_metrics_rule.go
@@ -24,8 +24,8 @@ func resourceNewRelicEventsToMetricsRule() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"account_id": {
 				Type:        schema.TypeInt,
-				ForceNew:    true,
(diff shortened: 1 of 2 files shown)


zlesnr:
I tested this locally using the following examples from our docs.

```
provider "newrelic" {
  account_id = var.account_id
}

data "newrelic_entity" "app" {
  name = "Dummy App"
}

resource "newrelic_events_to_metrics_rule" "foo" {
  account_id = var.account_id
  name = "Example events to metrics rule"
  description = "Example description"
  nrql = "SELECT uniqueCount(account_id) AS 'Transaction.account_id' FROM Transaction FACET appName, name"
}

resource "newrelic_workload" "foo" {
    name = "Example workload"
    account_id = var.account_id

    entity_guids = [data.newrelic_entity.app.guid]

    entity_search_query {
        query = "name like 'Example application'"
    }

    scope_account_ids =  [var.account_id]
}
```

This combined with the `unset NEW_RELIC_ACCOUNT_ID` I believe exercises the code paths that we're aiming to standardize on in #821.

PR #827, opened 2020-08-03
fix(alerts): ensure threshold_occurrences case fold comparison

zlesnr:
Resolves: #829


zlesnr:
fixes #829

zlesnr:
The test failure here "Error: unable to expand NRQL condition term: one of `duration` or `threshold_duration` must be configured for block `term`, unable to expand NRQL condition term: one of `duration` or `threshold_duration` must be configured for block `term`" is confusing me.  It seems that in some cases, when calling `d.Get("term")` we end up with 4 terms, which should not be possible.  Looking at this yesterday, I wasn't able to come up with an answer.  When I came back to this PR this morning, the tests were all passing, but running them again continues to fail in the way that I was observing yesterday.

PR #828, opened 2020-08-04
fix(alerts): better state migration for violation_time_limit

zlesnr:


zlesnr:
I believe this will resolve a migration issue for any existing nrql alerts moving from 1.x to 2.x.  New resources get written with the new state, but old resources continue to use the old value.  I'll work to modify my state file and reproduce tomorrow.

zlesnr:
I don't believe we have an issue here.  Testing a migration manually and inspecting the `plan` as well as the state file makes me think this is operating as expected.

Issue #843, opened 2020-08-13
cannot create resource "newrelic_infra_alert_condition" of type "infra_host_not_reporting"


dbonfigli:
Starting from version 2.4.0 it is impossible to create a resource  ```newrelic_infra_alert_condition``` of type ```infra_host_not_reporting```.

### Terraform Version
From 2.4.0 up.

### Affected Resource(s)
```newrelic_infra_alert_condition``` of type ```newrelic_infra_alert_condition```.

### Debug Output
```
module.host_not_reporting_alert.newrelic_infra_alert_condition.host_not_reporting: Creating...

Error: value is not supported by condition type infra_host_not_reporting

  on ../../../modules/newrelic_infra_host_not_reporting_alert/main.tf line 9, in resource "newrelic_infra_alert_condition" "host_not_reporting":
   9: resource "newrelic_infra_alert_condition" "host_not_reporting" {
```

### Expected Behavior
I expected to create the resource.

### Actual Behavior
An error occours like reported in the debug output above.

### Steps to Reproduce
Use this:
```hcl
resource "newrelic_infra_alert_condition" "host_not_reporting" {
  policy_id = "xxx"
  name       = "Host Not Reporting"
  type       = "infra_host_not_reporting"
  critical {
    duration      = var.crit_duration
  }
}
```
and then `terraform apply`.

### Important Factoids
I think the commit 920ec678 created the problem.

zlesnr:
Thank you for raising the issue.  I'm not seeing `920ec678`, but perhaps you mean `e75713c7751b667aac798156ec9d9703920ec678` since this looks like what you are referencing.

dbonfigli:
> but perhaps you mean e75713c7751b667aac798156ec9d9703920ec678

yes, you are right, I meant that commit.

PR #844, opened 2020-08-13
fix: cannot create resource "newrelic_infra_alert_condition" of type …

dbonfigli:
This patch fix issue #843. 

When type of resource "newrelic_infra_alert_condition" is "infra_host_not_reporting" then critical.value cannot be defined in the resource definition because it does not make sense, if c.Critical.Value is 0 in the check at line 174 then this means that the value has not been defined in a terraform resource definition, so using >= is wrong and always cause the error ```Error: value is not supported by condition type infra_host_not_reporting``` when perfoming terraform apply.

Please note that it is not enough to apply this patch, you also need to merge the pull request #832, otherwise a nil pointer deference error occours.

### DIFF
--- a/newrelic/structures_newrelic_infra_alert_condition.go
+++ b/newrelic/structures_newrelic_infra_alert_condition.go
@@ -174,7 +174,7 @@ func validateAttributesForType(c *alerts.InfrastructureCondition) error {
 		if c.Critical.Function != "" {
 			return fmt.Errorf("time_function is not supported by condition type %s", c.Type)
 		}
-		if *c.Critical.Value != 0.0 {
+		if c.Critical.Value != nil && *c.Critical.Value != 0.0 {
 			return fmt.Errorf("value is not supported by condition type %s", c.Type)
 		}
 	}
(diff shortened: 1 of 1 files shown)


newrelic/structures_newrelic_infra_alert_condition.go
```diff
@@ -171,7 +171,7 @@ func validateAttributesForType(c *alerts.InfrastructureCondition) error {
 		if c.Critical.Function != "" {
 			return fmt.Errorf("time_function is not supported by condition type %s", c.Type)
 		}
-		if *c.Critical.Value >= 0.0 {
+		if *c.Critical.Value != 0.0 {
```
zlesnr, line 174: This seems like we should also check for nil pointer here as well.

dbonfigli: validateAttributesForType is only called in expandInfraAlertCondition where condition.Critical is filled by expandInfraAlertThreshold(d.Get("critical")) where d.Get is the terraform sdk function that given the schema.ResourceData returns always the default value if it does not exists in the configuration as explained in the doc:

```
// If the key does exist in the schema but doesn't exist in the configuration,
// then the default value for that type will be returned. For strings, this is
// "", for numbers it is 0, etc.
```

so, in expandInfraAlertThreshold, ```val, ok := rah["value"];``` ```ok``` will always be true and so alertInfraThreshold.Value would always be a valid pointer.

But yes, we could add a check for nil pointer there to be extra sure and future proof.

zlesnr: Ah I see what you mean.  Thank you for making the change though.  I just had nil pointer checks on the brain I suppose. 

The CLA looks good, thank you, and I'll ensure the tests run locally before merge.

PR #851, opened 2020-08-19
chore(contrib): include a contribution guide

### LINKED ISSUE #746: Add a meaningful CONTRIBUTING.md guide
### Feature Description

A `CONTRIBUTING.md` which not only includes "here's how to fork and submit a PR" (like many are resigned to in the OSS world), but one which helps people start making meaningful contributions who don't have the tribal knowledge that you do. (I've brought this up directly with @jthurman42 before, but wanted to create a specific issue for it.)

It could double as an onboarding guide for new hires at New Relic who join the team working on this project.

Here are some examples from a project which I believe does an excellent job:

* https://golangci-lint.run/contributing/architecture/
* https://golangci-lint.run/contributing/new-linters/

zlesnr:
Fixes #746

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,157 @@
+# Contributing
+
+At its core, this project is a Terraform provider, and so an understanding of
+what Terraform is and how it works would help anyone looking to contribute to
+this provider.
+
+For those new to Terraform, the following might be good starting places.
+
+-   [How Terraform Works][how_terraform_works]
+-   [Plugin Types - Providers][how_terraform_works]
+
+For those who are already familiar with Terraform, there are still plenty of
+good resources on the [Extending Terraform][extending_terraform] page that are
+worth looking at, but those might be a topic by topic basis, depending on the
+need.  Also worth reading are the [Best Practices][best_practices].
+
+## DTK Who?
+
+The Developer Toolkit Team is a small team dedicated to open source, and
+integrating New Relic APIs into the [Go Client][client_go], which we then
+leverage in other projects, including this one.  You can read more about the
+team [here][dtk].
+
+## Future facing APIs
+
+New Relic has several APIs, and it's worth understanding at a high level what
+this means for the provider.  Also note, that each of the APIs for a given
+product are the results of many teams, each owning their implementation,
+documentation, and feature set, etc.
+
+However, at a high level there is a concerted effort within New Relic to try
+and move away from the REST APIs, and towards a newer GraphQL based API that we
+call NerdGraph.  This migration is likely to take years.
+
+What is mostly clear at this point, is that each team will continue to own
+their migration, implementation, and feature set in terms of what they choose
+to put into the GraphQL API.
+
+### REST v2 APIs
+
+Most of the documentation for the REST APIs are available at the [API
+Explorer][api_explorer].  There you can see the call structures and responses,
+and what methods are available.
+
+### GraphQL API
+
+To play and experiment with GraphQL, you can use the [graphiql][graphiql]
+interface to perform queries and mutations.  Note that these are using the API,
+so it's possible to make changes on your account.  Caution is advised when
+performing mutations.
+
+New fields, and new endpoints are added regularly in GraphQL, and so the
+API surface area continues grow and scale.
+
+## API Clients
+
+There are a couple of different API clients that are used in this project, but
+the bulk of the work is being done in the [New Relic Go Client][client_go].
+
+There is also a Go Agent that is in use for testing, so that we have a mock
+application to use during integration tests, and there is also an Insights
+client that is likely on its way out of this project, so we won't go into
+detail here.
+
+## Development Process
+
+### main.tf
+
+When working to implement a feature or test a change, it's helpful to have an
+HCL file that is isolated from any real production/staging environment.  When
+using environment variables to [configure the provider][provider-config-guide],
+the following is all that is needed to get a working provider for most
+situations.
+
+First export the necessary environment variables.
+
+```bash
+# Add this to your .bash_profile or .bashrc
+export NEW_RELIC_API_KEY="<your New Relic Personal API key>"
+export NEW_RELIC_ADMIN_API_KEY="<your New Relic Admin API key>"
+export NEW_RELIC_REGION="US"
+```
+
+Then configure the provider.
+
+```hcl
+provider "newrelic" {}
+```
+
+Then you can begin to include a resource or data source that you want to test,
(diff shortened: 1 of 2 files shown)


LESSONS.md
```diff
@@ -0,0 +1,7 @@
+2020-08-18 08:49:56
+
+# Lessons Learned
+
+-   TypeSet vs TypeList
```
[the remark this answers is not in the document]
sanderblue: Honestly, I'm a little hesitant to do that mostly because I still feel like its a matter of having more in-depth knowledge of the underlying behavior. But I guess I wouldn't be opposed to saying something like "if the attribute can be a TypeList, then go with TypeList. We've found attributes that use TypeList to be bit less nuanced and easier to maintain and still accomplish the same goal as TypeSet"... please fact check me on the "still accomplish the same goal" part 😉

LESSONS.md
```diff
@@ -0,0 +1,7 @@
...
+-   Structure files vs single file per resource/data source
```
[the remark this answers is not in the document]
sanderblue: Correct on HashiCorp suggesting not to do it, but I haven't found any issues with the pattern. I think I might prefer having the structures files because if we moved those functions to the main resource files, some of those files would be impossible to navigate. I like seeing functionality side-by-side in my editor too and having the two separate files helps me there I think. All that being said, I am totally okay with whichever pattern we want to go with.

LESSONS.md
```diff
@@ -0,0 +1,7 @@
+2020-08-18 08:49:56
+
+# Lessons Learned
```
zlesnr, line 3: Do we have any other lessons we want to include here?

LESSONS.md
```diff
@@ -0,0 +1,7 @@
...
+
+-   TypeSet vs TypeList
+-   Structure files vs single file per resource/data source
+-   Single API per resource/datasource
```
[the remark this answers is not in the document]
sanderblue: I mean we kind of did. I think this can be read more like "Resources should have a 1:1 relationship with their associated part of the API". The example is any of the resources that use REST and/or NerdGraph are still a 1:1 relationship with their specific feature. The only resource that contradicts that is `alert_policy` I believe, which leveraged the policy REST endpoint and the policy_channel REST endpoint when it should only have been using the policy endpoint to adhere to the 1:1 relationship.

CONTRIBUTING.md
```diff
@@ -0,0 +1,144 @@
...
+checkout go.mod` in the provider to go back to a released version of client.
+You'll want to make sure the version number in the provider lines up with the
+client version you want to be using as well.
+
+### Commit messages
+
+To keep a style and make allow us to automate the generating of a change log,
+we require that that commit messages adhere to a standard.
+
+TL;DR here is that the commit message must match this regular expression.
```
sanderblue, line 120: Probably meant to remove "here is that" maybe?

CONTRIBUTING.md
```diff
@@ -0,0 +1,144 @@
...
+The simplest case is when the API calls you want to make are already in the [Go
+Client][client_go] and the only changes you need to make are in the provider
+code.  In that case, simply building the provider binary with the correct name
+and running the plan will get you there.
+
+```shell
+go build -o terraform-provider-newrelic && terraform init -upgrade && TF_LOG=INFO terraform plan
+```
+
+### Developing with the [Go Client][client_go]
```
sanderblue, line 94: 👍  nice addition

sanderblue: Thoughts on the heading title "Developing with our [Go Client][client_go]"

CONTRIBUTING.md
```diff
@@ -0,0 +1,144 @@
...
+
+## API Clients
+
+There are a couple of different API clients that are used in this project, but
+the bulk of the work is being done in the [New Relic Go Client][client_go].
+
+There is also a Go Agent that is in use for testing, so that we have a mock
+application to use during integration tests, and there is also an Insights
+client, that is likely on its way out of this project, so I won't go into
+detail here.
```
sanderblue, line 63:
```diff 
- I
+ we
```
😉

LESSONS.md
sanderblue: Oofda... yeah probably 😆 . I can't think of any other big ones at the moment, but we can be an effort to try and add things to this as we come across nuances and such.

LESSONS.md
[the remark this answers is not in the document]
ctrombley: From the Extending Terraform guide:

TypeSet
> TypeSet implements set behavior and is used to represent an unordered collection of items, meaning that their ordering specified does not need to be consistent, and the ordering itself has no impact on the behavior of the resource.

TypeList:
> Used to represent an ordered collection of items, where the order the items are presented can impact the behavior of the resource being modeled. An example of ordered items would be network routing rules, where rules are examined in the order they are given until a match is found. The items are all of the same type defined by the Elem property.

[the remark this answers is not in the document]
ctrombley: I don't think we should recommend one over the other - they both have their uses.

It probably is worth pointing out that the idiom for nesting schema structures is `TypeList` with `MaxItems: 1` as discussed [here](https://github.com/hashicorp/terraform-plugin-sdk/issues/155#issuecomment-489699737).

CONTRIBUTING.md
```diff
@@ -0,0 +1,144 @@
...
+
+For those who are already familiar with Terraform, there are still plenty of
+good resources on the [Extending Terraform][extending_terraform] page that are
+worth looking at, but those might be a topic by topic basis, depending on the
+need.
+
+## DTK Who?
+
+The Developer Toolkit Team is a small team dedicated to open source, and
+integrating New RElic APIs into the [Go Client][client_go], which we then
```
ctrombley, line 20: nit: s/New RElic/New Relic/

CONTRIBUTING.md
```diff
@@ -0,0 +1,144 @@
...
+### main.tf
+
+When working to implement a feature or test a change, it's helpful to have an
+HCL file that is isolated from any real projection/staging environment.  When
+using environment variables to [configure the provider][provider-config-guide],
+the following is all that is needed to get a working provider for most
+situations.
+
+```hcl
+provider "newrelic" {}
```
ctrombley, line 76: Maybe list the env vars here that are required at a minimum as well?

LESSONS.md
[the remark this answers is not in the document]
ctrombley: All said I am a fan of co-locating code in the same files.  This pattern propagated from the Extending Terraform guide at some point early on.

PR #859, opened 2020-08-31
docs: update process running example

nathanweddle:
1. Updated process running example to have a `where` clause. 
2. Removed (`) back ticks. These are only required if there is a space in the attribute.
3. Stated that `process_where` is required for a process running condition.


sanderblue (approved):
Awesome, big thanks for this!

PR #862, opened 2020-09-03
fix(nrql_alert_condition): add missing zeros to violation_time_limit_seconds to the new:old map

sanderblue:


newrelic/resource_newrelic_nrql_alert_condition_test.go
```diff
@@ -476,7 +476,7 @@ resource "newrelic_nrql_alert_condition" "foo" {
 		time_function = "any"
 	}
 
-	violation_time_limit_seconds = 3600
+	violation_time_limit_seconds = 86400
```
zlesnr, line 479: Since this is the deprecated config here, do we still want to change this?

zlesnr (approved):

newrelic/resource_newrelic_nrql_alert_condition_test.go
sanderblue: No, not necessarily, but it does test the bug scenario.

PR #865, opened 2020-09-04
Added support for expiration (loss of signal) and extrapolation (gap filling) settings

### LINKED ISSUE #810: Expose access to some underlying GraphQL data related to alerts
Hello again!

### Context

We're creating a large set of alert conditions, including ones which help us ensure that critical processes are running on our EC2 instances. We want to know when a process dies on a running instance, but avoid alerts when the whole instance starts shutting down.

We're using the [`newrelic_nrql_alert_condition`](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/nrql_alert_condition) resource, Terraform v0.12.29, and New Relic provider v2.3.0. For cross-reference, we have an open ticket with [New Relic Support](https://support.newrelic.com/tickets/410665/). The support agent directed us to [NerdGraph API: Loss of signal and gap filling](https://docs.newrelic.com/docs/nerdgraph-api-loss-signal-gap-filling).

In my experience, applying the following configuration to an existing alert condition resolves the issue of alerts going off when an EC2 instance terminates:

```plain
"openViolationOnExpiration": false  ← Do NOT open a new violation when the signal is lost
"closeViolationsOnExpiration": true ← Close lost signal violations when they expire
"expirationDuration": 1200          ← Amount of time, in seconds, before considering the signal as lost.
```

Here is a sample alert condition:

```hcl
resource "newrelic_nrql_alert_condition" "cond_process_ecs_init" {
  name        = "TEST Name"
  enabled     = true

  # Bind this condition to the following alert policy
  policy_id = newrelic_alert_policy.ops.id

  # Type of metric and where to read it from
  type           = "static"
  value_function = "single_value"

  # Query
  nrql {
    query       = "FROM ProcessSample SELECT uniqueCount(entityAndPid) WHERE commandName = 'amazon-ecs-init' AND contained = 'false' AND label.BusinessService = '${local.tag_business_service}' AND ec2State = 'running' FACET ec2InstanceId LIMIT MAX"
    since_value = "15" # Evaluation offset
  }

  # CRIT
  critical {
    operator              = "below"
    threshold             = 1
    threshold_occurrences = "all"
    threshold_duration    = 900
  }

  # Minutes until it auto-closes
  violation_time_limit = "ONE_HOUR"
}
```

These settings can be applied to _existing_ alert conditions with the following request to the NerdGraph GraphQL API:

```graphql
mutation {
  alertsNrqlConditionStaticUpdate(
    accountId: XXXXXXX
    id: XXXXXXXX
    condition: {
      expiration: {
        closeViolationsOnExpiration: true
        expirationDuration: 1200
        openViolationOnExpiration: false
      }
    }
  ) {
    id
  }
}
```

### Feature Description

Expose the following attributes of the GraphQL mutation (above) via Terraform as read/write: `openViolationOnExpiration`, `closeViolationsOnExpiration`, `expirationDuration`. This would allow us to keep everything directly in Terraform without having to make additional ad-hoc mutation calls to the GraphQL API.

### LINKED ISSUE #835: Please support new alert type
### Feature Description
https://newrelic.slack.com/archives/CP9Q30JDC/p1597079579145000

Apologies for those that aren't able to see this. It's not a released feature yet.

ahornerr:
TODO: Update newrelic-client-go after https://github.com/newrelic/newrelic-client-go/pull/447 is merged

Closes #810
Closes #835

### DIFF
--- a/newrelic/structures_newrelic_nrql_alert_condition.go
+++ b/newrelic/structures_newrelic_nrql_alert_condition.go
@@ -41,6 +41,20 @@ var (
 		alerts.NrqlConditionViolationTimeLimits.TwelveHours:     43200,
 		alerts.NrqlConditionViolationTimeLimits.TwentyFourHours: 86400,
 	}
+
+	// old:new
+	fillOptionMap = map[string]alerts.AlertsFillOption{
+		"none":       alerts.AlertsFillOptionTypes.NONE,
+		"last_value": alerts.AlertsFillOptionTypes.LAST_VALUE,
+		"static":     alerts.AlertsFillOptionTypes.STATIC,
+	}
+
+	// new:old
+	fillOptionMapNewOld = map[alerts.AlertsFillOption]string{
+		alerts.AlertsFillOptionTypes.NONE:       "none",
+		alerts.AlertsFillOptionTypes.LAST_VALUE: "last_value",
+		alerts.AlertsFillOptionTypes.STATIC:     "static",
+	}
 )
 
 // NerdGraph
@@ -125,6 +139,14 @@ func expandNrqlAlertConditionInput(d *schema.ResourceData) (*alerts.NrqlConditio
 	}
 
 	input.Terms = terms
+
+	if input.Expiration, err = expandExpiration(d); err != nil {
+		return nil, err
+	}
+
+	if input.Signal, err = expandSignal(d); err != nil {
+		return nil, err
+	}
 
 	return &input, nil
 }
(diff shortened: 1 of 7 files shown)


ctrombley:
Closes #835

newrelic/structures_newrelic_nrql_alert_condition.go
```diff
@@ -304,6 +326,41 @@ func expandNrqlTerms(d *schema.ResourceData, conditionType string) ([]alerts.Nrq
...
+	}
+
+	return &expiration, nil
+}
+
+// NerdGraph
+func expandSignal(d *schema.ResourceData) (*alerts.AlertsNrqlConditionSignal, error) {
+	var signal alerts.AlertsNrqlConditionSignal
+
+	mappedFillOption := fillOptionMap[d.Get("fill_option").(string)]
```
sanderblue, line 349: We will likely need some case transformation in the expand and flatten functions for the new enum types. Right now if a user tries to set `fill_option` to `Last_Value` or `last_valuE`, etc, this will probably break. For UX purposes, we try to keep things case-insensitive to provide slightly easier UX for users, but that means a little extra work for us in the code to avoid state drift and errors 😎 

There are a few ways to accomplish this. One pattern is to always store the value as lowercase in the state. Take a look at [`threshold_occurrences`](https://github.com/newrelic/terraform-provider-newrelic/blob/523bbac5cfb78ac1ba5a6c7899726f5efa84e5b2/newrelic/resource_newrelic_nrql_alert_condition.go#L58-L66) and how it's handled in the expand and flatten functions. This won't necessarily be the exact same pattern for `fill_option`, but it will be probably be similar. And if you have other ideas on how to accomplish this, we're all about feedback and ideas! 🙂

sanderblue (changes requested):
One comment regarding case sensitivity for attribute values.

zlesnr:
This looks good to me.  Thanks for putting this up.

newrelic/structures_newrelic_nrql_alert_condition.go
ahornerr: Good catch. I will ensure these are lowercased in the flatten/expand functions.

ctrombley (approved):
🌮 :shipit: 

Thanks @ahornerr !

sanderblue (approved):
:shipit:

PR #870, opened 2020-09-09
Fixed an issue with empty extrapolation settings causing API errors

ahornerr:
When excluding `fill_option` an API error would be returned. This PR fixes that issue.

```
Error: Argument "condition" has invalid value $condition.
In field "signal": Expected type "AlertsNrqlConditionSignalInput", found {fillOption: "", fillValue: null}.
In field "fillOption": Expected type "AlertsFillOption", found ""., null
```

### DIFF
--- a/newrelic/provider_test.go
+++ b/newrelic/provider_test.go
@@ -90,6 +90,7 @@ func testAccPreCheck(t *testing.T) {
 
 func testAccCreateApplication(t *testing.T) {
 	app, err := newrelic.NewApplication(
+		newrelic.ConfigFromEnvironment(),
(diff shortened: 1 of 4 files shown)


newrelic/provider_test.go
```diff
@@ -90,6 +90,7 @@ func testAccPreCheck(t *testing.T) {
 
 func testAccCreateApplication(t *testing.T) {
 	app, err := newrelic.NewApplication(
+		newrelic.ConfigFromEnvironment(),
```
ahornerr, line 93: This allows setting the `NEW_RELIC_HOST` env to staging for testing against our staging environment

Issue #849, opened 2020-08-18
Can't create newrelic_entity for a NR One dashboard due to unsupported VIZ domain


schmickie:
I have defined a dashboard using newrelic_dashboard. It has already been applied and exists in NR One. I then defined newrelic_entity and newrelic_entity_tags resources for this dashboard. I have the “name” for the newrelic_entity set to the title for the dashboard, since dashboards don’t have a name attribute. I set the domain to APM and the type to DASHBOARD, but the terraform provider returns the following error (where XXXX is my dashboard’s title).

Error: the name ‘XXXX’ does not match any New Relic One entity for the given search parameters

I wasn’t sure if the domain was supposed to be APM, so I tried every other possible option for that attribute, and they all give me the same error. It must be possible to set up a newrelic_entity (and associated tags) for a dashboard because DASHBOARD is one of the options for type on newrelic_entity. 

I used the graphql interface to look up the dashboard, and it looks like the domain is supposed to be VIZ, but the provider doesn't support that option.

Error: expected domain to be one of [APM BROWSER INFRA MOBILE SYNTH], got VIZ

### Terraform Version
```
$ terraform -v
2020/08/18 14:35:54 [INFO] Terraform version: 0.12.18
2020/08/18 14:35:54 [INFO] Go runtime version: go1.12.13
2020/08/18 14:35:54 [INFO] CLI args: []string{"C:\\Users\\lestes\\Dev\\tools\\terraform.exe", "-v"}
2020/08/18 14:35:54 [DEBUG] Attempting to open CLI config file: C:\Users\lestes\AppData\Roaming\terraform.rc
2020/08/18 14:35:54 [DEBUG] File doesn't exist, but doesn't need to. Ignoring.
2020/08/18 14:35:54 [INFO] CLI command args: []string{"version", "-v"}
Terraform v0.12.18
2020/08/18 14:35:54 [DEBUG] checking for provider in "."
2020/08/18 14:35:54 [DEBUG] checking for provider in "C:\\Users\\lestes\\Dev\\tools"
2020/08/18 14:35:54 [DEBUG] checking for provider in ".terraform\\plugins\\windows_amd64"
2020/08/18 14:35:54 [DEBUG] found provider "terraform-provider-newrelic_v2.5.0.exe"
[...]

PR #895, opened 2020-09-22
fix(entity): add VIZ domain, update docs to mention that all domains …

schmickie:
…are searched if none is specified, fix typo WORRKLOAD in docs


schmickie:
Resolves: #849

sanderblue (approved):
Awesome! Thanks for taking the time to work through this! Very much appreciated! 🎉

PR #908, opened 2020-09-29
docs: update supported Go information and test config

zlesnr:
fixes: #474

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -14,6 +14,7 @@ jobs:
         go-version:
            - 1.13.x
            - 1.14.x
+           - 1.15.x
         platform:
           - ubuntu-latest
           - macos-latest
--- a/README.md
+++ b/README.md
@@ -69,6 +69,16 @@ _Note:_ Acceptance tests _create real resources_, and often cost money to run. T
 $ make testacc
 ```
 
+#### Go Version Support
+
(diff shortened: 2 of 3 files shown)


.github/workflows/compile.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     strategy:
       matrix:
         go-version:
-           - 1.13.x
            - 1.14.x
+           - 1.15.x
```
sanderblue, line 16: Wondering if we should also keep `1.13.x` around just so we're informed if we introduce a breaking change for an older version of Go. Thoughts?

zlesnr: If we did and the build failed, would that prevent us from merging the PR?

sanderblue: Yeah, but likely won't run into that problem. And when we do, we can make a decision on whether or not to remove that version of the build matrix, or update the code in question, or create a separate parallel job that runs on Go 1.13 that's allowed to fail. 

Also totally fine keeping things as they are in this PR. I'm more so thinking about making sure we stay informed when/if we introduce something messes with things.

zlesnr: Okay, I've added 1.13 back in.  You make a good point.

sanderblue (approved):
🐑 it

sanderblue (approved):
🐑 it

PR #919, opened 2020-09-30
chore: replace some v2 GetPolicy() with nerdgrpah QueryPolicy()

zlesnr:
Without this change, in large environments it appears that we place too much
load on the backing services, which can cause sporadic behavior.  With recent
API key changes, we are now in a position to straddle the line between REST and
Nerdgraph for a single resource, and so here we use Nerdgraph to replace the
GetPolicy calls, which perform a list to detect if the policy in question exists
before throwing away the response.


sanderblue (approved):
🦙 says  🐑 it

PR #953, opened 2020-10-10
Add instructions for New Relic One users to get an api key

davetron5000:
Because there is no Web UI for New Relic One users to get access to their personal API key, they need to execute a query to do so.  The official New Relic docs are in flux and don't show the correct instructions.


sanderblue:
Hi @davetron5000, this is excellent. Really appreciate the contribution. It looks like the only blocker on this is related to commit message formatting (we use [convention commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/#summary) format). Can we get your commit message amended to read something similar to the following?

```
docs: add instructions for New Relic One users to get an api key
```

😎

davetron5000:
OK, done, and rebased off of the main branch.  Let me know if you need anything else!

sanderblue (approved):
Awesome! Big thanks!

Issue #923, opened 2020-10-01
2.8.0 is getting 500 errors from API


alx13:
API calls to PUT dashboard leading to 500 Internal Server Error

### Terraform Version
`v0.12.26`

### Affected Resource(s)
`newrelic_dashboard` at least

alx13:
And now I'm getting different error:
```
2020-10-01T12:24:33.928+0200 [DEBUG] plugin.terraform-provider-newrelic_v2.8.0: time=newrelic_dashboard.dashboard: Refreshing state... [id=1431069]
"2020-10-01T12:24:33+02:00" level=debug msg="performing request" method=POST url="https://api.newrelic.com/graphql"
2020-10-01T12:24:33.928+0200 [DEBUG] plugin.terraform-provider-newrelic_v2.8.0: time="2020-10-01T12:24:33+02:00" level=debug msg="performing request" method=POST url="https://api.newrelic.com/graphql"
2020-10-01T12:24:33.928+0200 [DEBUG] plugin.terraform-provider-newrelic_v2.8.0: 2020/10/01 12:24:33 [INFO] Reading New Relic entities
2020-10-01T12:24:33.928+0200 [DEBUG] plugin.terraform-provider-newrelic_v2.8.0: time="2020-10-01T12:24:33+02:00" level=debug msg="performing request" method=POST url="https://api.newrelic.com/graphql"
2020-10-01T12:24:33.931+0200 [DEBUG] plugin.terraform-provider-newrelic_v2.8.0: 2020/10/01 12:24:33 [INFO] Reading New Relic dashboard 1431069
2020-10-01T12:24:33.931+0200 [DEBUG] plugin.terraform-provider-newrelic_v2.8.0: time="2020-10-01T12:24:33+02:00" level=debug msg="performing request" method=GET url="https://api.newrelic.com/v2/dashboards/1431069.json"

Warning: `newrelic_alert_condition` has been deprecated.  Use `newrelic_nrql_alert_condition` instead.

  on alerts.tf line 11, in resource "newrelic_alert_condition" "apdex_score":
  11: resource "newrelic_alert_condition" "apdex_score" {

(and 3 more similar warnings elsewhere)


Error: rpc error: code = Unavailable desc = transport is closing



Error: rpc error: code = Unavailable desc = transport is closing



Error: rpc error: code = Unavailable desc = transport is closing



Error: rpc error: code = Unavailable desc = transport is closing


[...]

PR #962, opened 2020-10-15
fix(dashboard): use state migration to fix 500 error when upgrading from v2.7.5 to v2.8+

sanderblue:
Resolves: #923 

Steps to manually test locally:

1. Using v2.7.5 of the provider, configure and provision a dashboard with multiple widgets with at least one widget being something more complex like displaying the apdex is an app with a `metric_line_chart` visualization.
    ```hcl
    provider "newrelic" {
      version = "2.7.5"
    }

    resource newrelic_dashboard" "foo" {
      // ... configuration w/ widgets 'n such (no cross-account widgets)
    }
    ```

2. Install the latest local version of the provider (the code in this PR) by commenting out the `version` part of the provider configuration.

    ```hcl
    provider "newrelic" {
      # version = "2.7.5" 
    }

    // ...
    ```
    - Run ` rm -rf .terraform && rm terraform-provider-newrelic && go build && tf init` to reinitialize and install the provider using the local binary. **IMPORTANT: DO NOT DELETE THE STATE FILE!** 😉
    - Run `terraform apply` and things should run without any errors and no changes or diff should show up.

### DIFF
--- a/newrelic/resource_newrelic_dashboard.go
+++ b/newrelic/resource_newrelic_dashboard.go
@@ -153,6 +153,104 @@ func resourceNewRelicDashboard() *schema.Resource {
 				},
 			},
 			"widget": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				MaxItems:    300,
+				Description: "A nested block that describes a visualization. Up to 300 widget blocks are allowed in a dashboard definition.",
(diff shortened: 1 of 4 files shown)


ctrombley (approved):
🌮

Issue #955, opened 2020-10-12
newrelic_synthetics_monitor - 401 Unauthorized


qbabooba:
### Terraform Version
```
Terraform v0.13.4
```
### Affected Resource(s)
```
newrelic_synthetics_monitor
```

### Terraform Configuration
```hcl
provider registry.terraform.io/newrelic/newrelic v2.7.5
```

### Actual Behavior
newrelic_synthetics_monitor is failing with "401 response returned: HTTP 401 Unauthorized" all other resources in the configuration are working fine. Suspected that it needs ADMIN api key to work but new provider only allows for user API keys

Error: 401 response returned: HTTP 401 Unauthorized

### Expected Behavior
Provider should allow to publish configuration changes and new objects to newrelic_synthetics_monitor

### Steps to Reproduce

Create a new resource for Synthetic monitor, sample

```
provider "newrelic" {
  api_key = "NRAK-XXXXXXX"
  account_id = "xxxx"
  region = "US"
}
```

```
resource "newrelic_synthetics_monitor" "login" {
  name = "login page check"
  type = "SIMPLE"
  frequency = 5
  status = "ENABLED"
  locations = ["AWS_AP_SOUTHEAST_2"]

  uri                       = "https://somedomain.com"
  validation_string         = "sometext"
  verify_ssl                = true
}
```

apply configuration and observe 401 error

harrisj:
I have figured this out by looking at the source, but you can fix this by adding an `admin_api_key` to the Provider configuration with the Admin API key (should start with NRAA). Then it should work. The documentation needs to be fixed, because the official documentation does not mention the Admin API key at all as an option.

ctrombley:
Hi @harrisj, I think you are looking at a more recent version of the docs than the version of the provider you are pinned to.  You can change the version of the docs to match the version you're using with a dropdown at the top of the page.

Please update the terraform provider to v2.10.x or later to allow personal API keys to properly authenticate against the synthetics API.

qbabooba:
Hey @ctrombley  @harrisj 
Confirmed, working with 2.10+ and the documentation we were looking at was 2.10+. What confused me was the upgrade process to 2.x that stated that admin keys are no longer supported since 2.X. 
I suggest that documentation to be updated and state from 2.10.+ admin keys are no longer supported

harrisj:
Ah, you are indeed correct. I apologize for my mistake.

PR #973, opened 2020-10-19
chore(docs): update migration guide regarding when admin API key was removed

sanderblue:
Resolves: #955


jthurman42 (approved):
Thank you!

:shipit:

PR #975, opened 2020-10-20
Fix docs for alert_policy_channel

blckct:
The docs refer to data objects instead of resources, a small oversight but may be confusing for some users.


zlesnr:
Thank you for the PR.  Would you mind formatting your commit as mentioned in the contributing guide?

https://github.com/newrelic/terraform-provider-newrelic/blob/master/CONTRIBUTING.md#commit-messages

`docs(alert_policy_channel): update example reference` would be sufficient.

blckct:
@zlesnr Sorry, I was sure I did fix it properly on second try. Now it looks okay.

zlesnr (approved):
Great, thanks for the contribution.

PR #1015, opened 2020-11-07
Swap deprecation of violation_time_limit fields for NRQL alert condition

jaymcgrath:
Depends on https://github.com/newrelic/newrelic-client-go/pull/550
With the addition of `violation_time_limit_seconds` to NerdGraph's NRQL Alert Condition models, we now have a first class `seconds` field where users can choose an arbitrary value between 300s (5 minutes) and 2592000s (30 days). 
The old enum field `violation_time_limit` is restricted to a much smaller selection of values and is now deprecated in NerdGraph. 

Previously, a field `violation_time_limit_seconds` existed that accepted the same limited range of values as `violation_time_limit`, expressed as seconds. 

This PR deprecates the `violation_time_limit` field in favor of `violation_time_limit_seconds` to  align the NR terraform provider's representation of violationTTL with NerdGraph. 

Note to reviewer:

The fields `violation_time_limit` and `violation_time_limit_seconds` are distinct in NerdGraph, but behind the scenes, they set the same field for an alert condition. 

With `violation_time_limit` being an enum, and `violation_time_limit_seconds` is an arbitrary integer, values outside of the enum are represented as "NON_MATCHABLE_LIMIT_VALUE"

Since it always gets a value, even when configuring `violation_time_limit_seconds`,  I've marked `violation_time_limit` as `computed`, and set both fields to optional, and conflicting with one another.  Is this correct?

### DIFF
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -241,13 +241,12 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 				ConflictsWith: []string{"open_violation_on_group_overlap"},
 			},
 			"violation_time_limit_seconds": {
-				Deprecated:    "use `violation_time_limit` attribute instead",
 				Type:          schema.TypeInt,
 				Optional:      true,
-				Description:   "Sets a time limit, in seconds, that will automatically force-close a long-lasting violation after the time limit you select. Possible values are 3600, 7200, 14400, 28800, 43200, and 86400.",
+				Description:   "Sets a time limit, in seconds, that will automatically force-close a long-lasting violation after the time limit you select.  Must be in the range of 300 to 2592000 (inclusive)",
 				ConflictsWith: []string{"violation_time_limit"},
 				AtLeastOneOf:  []string{"violation_time_limit_seconds", "violation_time_limit"},
-				ValidateFunc:  validation.IntInSlice([]int{3600, 7200, 14400, 28800, 43200, 86400}),
+				ValidateFunc:  validation.IntBetween(300, 2592000),
 			},
 			// Exists in NerdGraph, but with different values. Conversion
 			// between new:old and old:new is handled via maps in structures file.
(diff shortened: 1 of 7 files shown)


ctrombley (dismissed):

jaymcgrath:
@ctrombley - Hi - After manual testing, I realized that I need to make a few more changes to before this is ready for merge.

1. add violation_time_limit_seconds field [to the newrelic client](https://github.com/newrelic/newrelic-client-go/blob/ae77dcc172d9bf59281a76fe92af550d89671e3d/pkg/alerts/nrql_conditions.go#L185)
2. remove [the logic that's translating the seconds value back to an enum representation](https://github.com/newrelic/terraform-provider-newrelic/blob/master/newrelic/structures_newrelic_nrql_alert_condition.go#L124) and just pass the integer value through
3. prepare post explaining why we're swapping the deprecated markers for these fields (adding fine grained control over violation TTL and to match nerdGraph's deprecation of the field)

jaymcgrath:
I believe this is ready for review, but please don't merge it until we can get a post together explaining why we did the deprecation swap.  Thanks! 😎 🙏

newrelic/resource_newrelic_nrql_alert_condition.go
```diff
@@ -274,7 +273,9 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 			},
 			"violation_time_limit": {
 				Type:          schema.TypeString,
+				Deprecated:    "use `violation_time_limit_seconds` attribute instead",
 				Optional:      true,
+				Computed:      true,
```
jaymcgrath, line 278: I added a computed to this field to prevent drift when using `violation_time_limit_seconds`.

jaymcgrath:
@ctrombley: this is ready for a review, but blocked by the broken CI in the go client repo, as https://github.com/newrelic/newrelic-client-go/pull/550 is a prereq for this one.

jaymcgrath:
> @ctrombley: this is ready for a review, but blocked by the broken CI in the go client repo, as [newrelic/newrelic-client-go#550](https://github.com/newrelic/newrelic-client-go/pull/550) is a prereq for this one.
> 
> Thanks!

Given the change management components, can we please wait til after the break (after 11/30) to review/merge this and the client PR  (https://github.com/newrelic/newrelic-client-go/pull/550)? 

Thanks

ahornerr (approved):

ctrombley (dismissed):

ctrombley (approved):

PR #1020, opened 2020-11-11
docs: include note about API key access

zlesnr:


website/docs/guides/migration_guide_v2.html.markdown
```diff
@@ -20,6 +20,14 @@ As of version 2.10.2, New Relic [**Admin API keys**](https://docs.newrelic.com/d
 
 -> <small>**Please note the following formatting for the provider's API key.** <br>Your **Personal API Key** has a prefix of `NRAK-` </small>
 
+**IMPORTANT** Please make sure that the Personal API Key has access to the
+account and resources you expect to be manipulating.  This can be confirmed on
```
ctrombley, line 24: Not all users have the "Users and Roles" UI.  I think this might cause confusion for those users.

ctrombley (changes requested):

website/docs/guides/migration_guide_v2.html.markdown
zlesnr: Okay, I'll drop that part.

ctrombley (approved):

PR #1025, opened 2020-11-12
fix(nrql_alert_condition): reverse attribute detection for migration

zlesnr:
Without this change, the deprecated attributes are preferred when they are found
within the state.  Here we reverse the order that the attributes are read to
determine if the particular resource has been migrated to the new API values.
This should ensure a clean migration from 1.x to 2.x in the case of an outlier
condition.


sanderblue (dismissed):
Nice 🐑

sanderblue (approved):

PR #1034, opened 2020-11-18
Chore: Fixing checks

jpvajda:
based on this article:

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#adding-a-system-path

added $GITHUB_PATH to fix failing checks

also made some fixes to the contributor file.


.github/workflows/changelog.yml
```diff
@@ -9,15 +9,6 @@ jobs:
   createPullRequest:
     runs-on: ubuntu-latest
     steps:
-      - name: Install Go
-        uses: actions/setup-go@v2
```
zlesnr, line 13: I think we still need the `setup-go` action here.

jpvajda: @zlesnr got it, so this entire block stays 

```- name: Install Go
      uses: actions/setup-go@v2
      with:
        go-version: 1.14.x
```

zlesnr: I think so yes.

zlesnr (approved):

Issue #1017, opened 2020-11-10
Provide v2 migration guidance for the `newrelic_nrql_alert_condition` resource


ctrombley:
The breaking changes introduced to this resource with v2 were not addressed in the v2 migration guide.  Let's add some information to the resource docs that show what changes are necessary to move from v1 to v2 of the provider for existing users of the NRQL condition resource.

PR #1037, opened 2020-11-19
docs(nrql_alert_condition): include notes about upgrading from 1.x

zlesnr:
fixes: #1017


ctrombley (approved):
Looks good, I might suggest a bullet list of change summaries before the examples (e.g. "`threshold` has been replaced with `threshold_duration`").   Or not, it's up to you.  I think this solves the issue as-is.

PR #1051, opened 2020-12-03
chore(deps): allow dependabot to bump versions for github actions

jlosito:
This allows dependabot to check the versions of any github action which this project uses and submit PRs with version bumps weekly.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-actions-up-to-date-with-dependabot


ctrombley:
Hi @jlosito!  Thank you for the contribution but we are currently using [renovate](https://github.com/marketplace/renovate) to stay up to date on our dependencies.  I think this will be redundant.

jlosito:
It doesn't look like renovate handles github actions, just the go dependencies.

zlesnr:
Just seeing this now.  I believe this is already happening with renovate.  See https://github.com/newrelic/terraform-provider-newrelic/commit/769c86984be375284509c7ce69c4e391fc9cb79c#diff-bdcc6a2a85f645f62724fe8dafbf0581cb0c1d65f6a76cb2985a9172e31a473c for an example.

Is there an action that needs to be updated here that isn't?

jlosito:
Indeed it does look like renovate is handling it.

zlesnr:
Okay no sweat.  Thanks for raising the issue anyway.

PR #1059, opened 2020-12-14
chore: added filter example

jpvajda:
added filter example to entity markdown


jpvajda:
closing and resubmitting through the standard git workflow as i need to correct the commit message.

PR #1086, opened 2021-01-06
Implement GraphQL Dashboard API

jthurman42:
*Note:* Dashboard API via NerdGraph is in closed beta, this PR will not work for everyone at this time.

Refactor `newrelic_dashboard` to use the new GraphQL API, enabling all new dashboard features (pages/multi-query/etc)

### DIFF
--- a/newrelic/provider.go
+++ b/newrelic/provider.go
@@ -144,6 +144,7 @@ func Provider() terraform.ResourceProvider {
 			"newrelic_infra_alert_condition":                    resourceNewRelicInfraAlertCondition(),
 			"newrelic_insights_event":                           resourceNewRelicInsightsEvent(),
 			"newrelic_nrql_alert_condition":                     resourceNewRelicNrqlAlertCondition(),
+			"newrelic_one_dashboard":                            resourceNewRelicOneDashboard(),
 			"newrelic_plugins_alert_condition":                  resourceNewRelicPluginsAlertCondition(),
 			"newrelic_synthetics_alert_condition":               resourceNewRelicSyntheticsAlertCondition(),
 			"newrelic_synthetics_monitor":                       resourceNewRelicSyntheticsMonitor(),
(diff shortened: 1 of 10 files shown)


newrelic/provider.go
```diff
@@ -144,6 +144,7 @@ func Provider() terraform.ResourceProvider {
 			"newrelic_infra_alert_condition":                    resourceNewRelicInfraAlertCondition(),
 			"newrelic_insights_event":                           resourceNewRelicInsightsEvent(),
 			"newrelic_nrql_alert_condition":                     resourceNewRelicNrqlAlertCondition(),
+			"newrelic_one_dashboard":                            resourceNewRelicOneDashboard(),
```
sanderblue, line 147: Did we inquire marketing and/or other sales folks on this name?  


Disclaimer: I'm on board with this resource name ✔️

newrelic/resource_newrelic_one_dashboard.go
```diff
@@ -0,0 +1,380 @@
...
+			"description": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The dashboard's description.",
+			},
+			"permissions": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "public_read_only",
+				ValidateFunc: validation.StringInSlice([]string{"private", "public_read_only", "public_read_write"}, false),
```
sanderblue, line 51: Not a blocker, but we've usually tried to make things case-insensitive for constants like these (per JT suggestions 😉 ), but I also see the merit with maintaining a slightly stricter API.

newrelic/provider.go
jthurman42: This is the name that was landed at after talking with some marketing folks.

newrelic/resource_newrelic_one_dashboard.go
jthurman42: They must be UPPERCASE for the API, and are upcase/tolower in the expand/flatten funcs, so updating the validator to not care is fine with me.

sanderblue (approved):
🐑 it

PR #1154, opened 2021-01-28
fix (newrelic_alert_muting_rule): condition tag validation

jaymcgrath:
fixes [issue #1153](https://github.com/newrelic/terraform-provider-newrelic/issues/1153)

muting rule condition tags have a special format that's parsed out before matching. They need to be in the format of `tag.some_value`.

The StringInSlice validation was treating tags like a normal attribute name, so any correctly configured tag was failing, and only the incorrectly formatted condition of `tag` would pass.

More details and repro steps [in the issue](https://github.com/newrelic/terraform-provider-newrelic/issues/1153)

### DIFF
--- a/newrelic/resource_newrelic_alert_muting_rule.go
+++ b/newrelic/resource_newrelic_alert_muting_rule.go
@@ -1,12 +1,31 @@
 package newrelic
 
 import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"log"
+	"regexp"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
-	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/newrelic/newrelic-client-go/pkg/errors"
 )
+
+func validateMutingRuleConditionAttribute(val interface{}, key string) (warns []string, errs []error) {
+	valueString := val.(string)
+	attemptedTagRegex := regexp.MustCompile(`^tag`)
+	correctTagRegex := regexp.MustCompile(`^tag\..+$`)
+
+	// tag.SomeValue attempted but does not match allowed format
+	if attemptedTagRegex.Match([]byte(valueString)) {
+		if !correctTagRegex.Match([]byte(valueString)) {
+			errs = append(errs, fmt.Errorf("%#v of %#v must be in the format tag.tag_name", key, valueString))
+			return
+		}
+		return
+	}
+	v := validation.StringInSlice([]string{"accountId", "conditionId", "policyId", "policyName", "conditionName", "conditionType", "conditionRunbookUrl", "product", "targetId", "targetName", "nrqlEventType", "tag", "nrqlQuery"}, false)
+	return v(valueString, key)
+}
 
 func resourceNewRelicAlertMutingRule() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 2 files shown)


jaymcgrath:
I can't seem to get the commit regex to pass linting. I've had persistent problems with this. After skimming https://www.conventionalcommits.org/en/v1.0.0/, and testing via regex101 (https://regex101.com/r/GIQUQF/1), I can't understand why `fix(newrelic_alert_muting_rule): condition tag validation ` isn't passing.

sanderblue:
Try `fix(alert_muting_rule): condition tag validation` for the commit message (altho what you wrote works, we usually dont prefix with `newrelic`).

For the lint error, this is the error I'm seeing:
```
										ValidateFunc:func(val interface{}, key string) (warns []string, errs []error) {
newrelic/resource_newrelic_alert_muting_rule.go:56:20: `if` block ends with a `return` statement, so drop this `else` and outdent its block (golint)
												} else {
												       ^
make: *** [golangci] Error 1
build/lint.mk:55: recipe for target 'golangci' failed
```

newrelic/resource_newrelic_alert_muting_rule.go
```diff
@@ -41,7 +43,23 @@ func resourceNewRelicAlertMutingRule() *schema.Resource {
...
+											attemptedTagRegex := regexp.MustCompile(`^tag`)
+											correctTagRegex := regexp.MustCompile(`^tag\..+$`)
+
+											// tag.SomeValue attempted but does not match allowed format
+											if attemptedTagRegex.Match([]byte(valueString)){
+												if !correctTagRegex.Match([]byte(valueString)){
+													errs = append(errs, fmt.Errorf("%#v of %#v must be in the format tag.tag_name", key, valueString))
+													return
+												} else {
+													return
```
sanderblue, line 57: I think the `else` condition/return can be removed so it continues to fallback to the general validation after.

newrelic/resource_newrelic_alert_muting_rule.go
```diff
@@ -41,7 +43,23 @@ func resourceNewRelicAlertMutingRule() *schema.Resource {
...
+												}
+											}
+											v := validation.StringInSlice([]string{"accountId", "conditionId", "policyId", "policyName", "conditionName", "conditionType", "conditionRunbookUrl", "product", "targetId", "targetName", "nrqlEventType", "tag", "nrqlQuery"}, false)
```
zlesnr, line 60: Should this just return `validation.StringInSlice`?

newrelic/resource_newrelic_alert_muting_rule.go
jaymcgrath: I believe it will fail because `tag.whatEver` won't be part of the slice of strings it's validating against, so I think we need to return here.

I left "tag" in the list for completeness' sake in the error message, but given this check and short circuit, it shouldn't ever let a plain "tag" through (without the `.tagName` suffix)

newrelic/resource_newrelic_alert_muting_rule.go
jaymcgrath: My understanding is that `validation.StringInSlice` returns a `schema.SchemaValidateFunc` which we need to call to perform the validation.

newrelic/resource_newrelic_alert_muting_rule.go
jaymcgrath: 😅 I just needed to outdent the return statement. To reiterate, we still want this return, since we've detected and validated a `tag.Whatever` and want to skip checking against the slice of fixed values.

newrelic/resource_newrelic_alert_muting_rule.go
sanderblue: Yeh, I see what you're saying. Returning that wouldn't return the expected type for the function, hence we return the result of calling the function.

zlesnr: Yes, you're right, my misunderstanding.

zlesnr (commented):
This looks good to me.  We probably want a test that confirms this behavior.

jaymcgrath:
I just added a couple of unit tests after breaking the validation out into its own function. Re-requesting review! 😃

sanderblue (approved):
Solid! Thanks!

PR #1167, opened 2021-02-01
feat(alert_muting_rule): add schedule support

### LINKED ISSUE #1156: How to create a muting rule with schedule?
### Objective
Documentation is incomplete.
The fields for schedule is not written anywhere.
What's the field, what's available value, and example.
https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/alert_muting_rule

### Acceptance Criteria
Example of how to create a muting rule with a schedule.

What I want a muting rule with this schedule:
- Start: Today
- End: Never
- Duration: 0 AM - 7 AM
- Occurrence: Everyday

jaymcgrath:
### Goals :soccer:
Add support for configuration of `newrelic_alert_muting_rule schedule` fields with Terraform. 
Closes #1156

### Dependencies
Requires https://github.com/newrelic/newrelic-client-go/pull/652. 
I've bumped the client version here to .57.2, so this should also close out #1162 

### Implementation Details :construction:
Add new MutingRuleScheduleSchema struct & field validation:
    * Add validateNaiveDateTime struct for testing conversion from go's time.time to an "offsetless" NaiveDateTime in NerdGraph

Add expand and flatten funcs for the schedule. Please note the following idiosyncracies which were accounted for in the funcs:
* Nerdgraph only accepts a NaiveDateTime but returns a full 8601 dateTime with an offset. To prevent drift, this offset must be removed before any of these dateTime fields are stored in state

* It's possible to configure a WEEKLY repeat with an empty list of `weekly_repeat_days`. This will initiate a once-weekly repeat on the day of the week for which the rule is first scheduled.
* To accomplish this, we need to serialize an empty slice instead of `nil` when weekly_repeat_days len is 0 but repeat is set to WEEKLY.

Update website docs to reflect new schedule fields.

### Testing Details :mag:
Add integration tests for creating, updating, and importing muting rules with schedule. These tests cover adding and removing different schedule components as well as adding and removing the entire schedule field.
Add unit tests for flatten and expand funcs.

### DIFF
--- a/newrelic/resource_newrelic_alert_muting_rule.go
+++ b/newrelic/resource_newrelic_alert_muting_rule.go
@@ -3,11 +3,12 @@ package newrelic
 import (
 	"fmt"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
-	"log"
 	"regexp"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/newrelic/newrelic-client-go/pkg/errors"
+	"log"
+	"time"
 )
 
 func validateMutingRuleConditionAttribute(val interface{}, key string) (warns []string, errs []error) {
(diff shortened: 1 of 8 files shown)


zlesnr:
This looks good to me.  I've not had a chance to pull it down and test.

versions.tf
```diff
@@ -0,0 +1,18 @@
+terraform {
```
sanderblue, line 1: Looks like this file snuck in. I'll make a note to update our `.gitignore` file.

website/docs/r/alert_muting_rule.html.markdown
```diff
@@ -45,6 +45,7 @@ The following arguments are supported:
   * `enabled` - (Required) Whether the MutingRule is enabled.
   * `name` - The name of the MutingRule.
   * `description` - The description of the MutingRule.
+  * `schedule` - (Optional) Specify a schedule for enabling the MutingRule. See [Schedule](#schedule) below for details
```
sanderblue, line 48: Since this feature seems to be gaining interesting, can we add an example showing the use of `schedule` in a section titled "Additional Example" at the bottom of this doc file? Here's an [example from NRQL alert conditions](https://github.com/newrelic/terraform-provider-newrelic/blob/master/website/docs/r/nrql_alert_condition.html.markdown#additional-examples).

sanderblue (changes requested):

newrelic/resource_newrelic_alert_muting_rule.go
```diff
@@ -26,6 +27,68 @@ func validateMutingRuleConditionAttribute(val interface{}, key string) (warns []
...
+}
+
+func scheduleSchema() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"end_repeat": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "The datetime stamp when the MutingRule schedule should stop repeating.",
+				ConflictsWith: []string{"schedule.repeat_count"},
```
sanderblue, line 48: I think for this to work as client side validation, I think this needs to use the index `0` - e.g. `[]string{"schedule.repeat_count"}`. Also I'm not sure if we need to do that because it appears the API returns a helpful error message.

e.g. 

```
newrelic_alert_muting_rule.foo: Creating...

Error: A downstream error occurred., [{"extensions":{"code":"BAD_USER_INPUT","validationErrors":[{"name":"schedule","reason":"repeat_count 1 provided in addition to end_repeat 2022-06-11T12:00:00 Please remove end_repeat or repeat_count"}]},"message":"Validation Error"}]

  on main.tf line 34, in resource "newrelic_alert_muting_rule" "foo":
  34: resource "newrelic_alert_muting_rule" "foo" {
```

jaymcgrath: I added a 0 to both and they tested fine! Nice catch! Thanks for calling this out, adding this gotcha to our onboarding doc

sanderblue (approved):
Awesome! Thanks for the contribution! 🎉

PR #1188, opened 2021-02-09
[one_dashboard]: Inherit nrql_query account_id from parent dashboard

### LINKED ISSUE #1174: BUG: Unable to create NR1 dashboard via Terraform with default account_id
### Summary
With the recent addition of `newrelic_one_dashboard`, I copy pasted the usage example from the [provider docs](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/one_dashboard). Since it's stated that `account_id` is optional, I removed those 2 lines. As per the docs, New Relic should default to the associated account from the API key used. Instead the following error is returned:

>The argument "account_id" is required, but no definition was found.

Please note, this is different from the error whereby an `account_id` isn't supplied to the resource ("The argument "account_id" is required, but was not set.").

### Attempts
I've cycled through my: User, Admin and REST keys, without success. I don't get why this resource *needs* the `account_id` to be set explicitly when others don't. This seems erroneous and conflicts with documentation.

What's more, in order to fetch me `account_id`, [I need my `account_id`](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/data-sources/account) (or the account name; but who uses names as unique identifiers?). Not sure if I'm missing something obvious here.

Also raised on the [forum](https://discuss.newrelic.com/t/137999). Thanks for your time.

### Code
```tf
terraform {
  required_version = "~> 0.14.0"
  required_providers {
    newrelic = {
      source = "newrelic/newrelic"
      version = "~> 2.17.0"
    }
  }
}

provider "newrelic" {
  account_id = 123456
  api_key = "NRAK-abcdefg"
}

resource "newrelic_one_dashboard" "exampledash" {
  name = "New Relic Terraform Example"
  page {
    name = "New Relic Terraform Example"
    widget_billboard {
      title = "Requests per minute"
      row = 1
      column = 1
      nrql_query {
        query = "FROM Transaction SELECT rate(count(*), 1 minute)"
      }
    }
  }
}
```

### AC

The expected behavior would be for this to inherit from the Dashboard's account_id which is inherited from the provider block.

jthurman42:
Currently you must specify the `account_id` for every `nrql_query`.  This change follows the rest of the provider in allowing that value to be inherited from the `account_id` of the dashboard (which is inherited from the API key used).

Resolves #1174

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -409,8 +409,6 @@ github.com/newrelic/go-agent/v3 v3.10.0 h1:qHcDwtC1DjOTI0r0nCtf8jT2LfEoNQ6gSo2tc
 github.com/newrelic/go-agent/v3 v3.10.0/go.mod h1:1A1dssWBwzB7UemzRU6ZVaGDsI+cEn5/bNxI0wiYlIc=
 github.com/newrelic/go-insights v1.0.3 h1:zSNp1CEZnXktzSIEsbHJk8v6ZihdPFP2WsO/fzau3OQ=
 github.com/newrelic/go-insights v1.0.3/go.mod h1:A20BoT8TNkqPGX2nS/Z2fYmKl3Cqa3iKZd4whzedCY4=
-github.com/newrelic/newrelic-client-go v0.57.1 h1:NnMETNhk7XY2cPE/6dK9u3IZHiFCqXI06DIxUoDrvLA=
-github.com/newrelic/newrelic-client-go v0.57.1/go.mod h1:k6e8L+H4I1n7oimYAwAe1b07wyutjNTkFpw0yGIEbug=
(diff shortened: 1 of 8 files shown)


jpvajda:
😄

zlesnr (approved):

rdhar:
This is brilliant, thank you! How soon after merging does it make its way to release?

jpvajda:
@rdhar we usually cut at least 1 release each week, so this will go out in the next release of our provider.  Right now the team is engage is some high priority work, but as soon as we cut another release this will get in. You can follow the release tags here:

https://github.com/newrelic/terraform-provider-newrelic/releases/tag/v2.18.0

Issue #1181, opened 2021-02-08
newrelic_one_dashboard: Heatmap are not being created/exported


Nordes:
### Feature Description
I didn't know if I should have marked this as a bug or simply a missing feature.

As I have dashboard with **heatmap**, they are impossible to be created with the provider. I tough that by exporting and importing it would at least tell me I am destroying something (which it did). The structure should be pretty similar to the other type of dashboard widget. So, I don't know why they are simply not available. 

### Describe Alternatives
No alternative

### Additional context
None

jthurman42:
Looks like the Heatmap configuration is returned in the `rawConfiguration` part of the API (vs the `configuration` section, which has specific implemented widgets).  Example:

```
              {
                "configuration": null,
                "id": "<REDACTED>",
                "layout": {
                  "column": 5,
                  "height": 3,
                  "row": 1,
                  "width": 4
                },
                "linkedEntities": null,
                "rawConfiguration": {
                  "nrqlQueries": [
                    {
                      "accountId": <REDACTED>,
                      "query": "FROM Transaction SELECT histogram(duration, buckets: 100, width: 0.1) FACET appName "
                    }
                  ]
                }
```

Adding another widget type to the provider that configures via the `rawConfiguration` would enable this.

@jpvajda I suggest we add a `widget_heatmap` to vet out the strategy, but overall need to audit any other missing visualization types.

Issue #1172, opened 2021-02-04
Import of newrelic_one_dashboard fails with tables + Filter Current dashboard flag


Nordes:
### Terraform Version
Terraform v0.14.2
+ provider registry.terraform.io/newrelic/newrelic v2.17.0

### Affected Resource(s)
Please list the resources as a list, for example:
- `newrelic_one_dashboard`

### Terraform Configuration
No configuration (Import issue)

![image](https://user-images.githubusercontent.com/446572/106959169-5184f780-6708-11eb-932a-f9994817e2d4.png)

### Actual Behavior
What actually happened? **Error** 

```sh
newrelic/playground $ rm -f terraform.tfstate &&  terraform import newrelic_one_dashboard.this MjkzMzM2NXxWSVp8REFTSEJPQVJEfGRhOjEwNzA5OA
newrelic_one_dashboard.this: Importing from ID "MjkzMzM2NXxWSVp8REFTSEJPQVJEfGRhOjEwNzA5OA"...
newrelic_one_dashboard.this: Import prepared!
  Prepared newrelic_one_dashboard for import
newrelic_one_dashboard.this: Refreshing state... [id=MjkzMzM2NXxWSVp8REFTSEJPQVJEfGRhOjEwNzA5OA]

Error: Invalid address to set: []string{"page", "0", "widget_table", "2", "linked_entity_guids"}
```

### Expected Behavior
I would expect the import to work like when I uncheck the option in the UX. (one.newrelic.com)

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create a dashboard with a NRQL request such as: `SELECT count(*), uniques(appName)  FROM TransactionError FACET transactionName SINCE 24 hours ago UNTIL now` displayed as table.
2. Tick the box `filter the current dashboard` (green check)
3. Go back to your console after noting the dashboard id
4.  `terraform init`
5. `terraform import newrelic_one_dashboard.this <some guid here>`

Basically, it's an empty terraform project with only the newrelic provider + the resource named as mentioned in the import.

workaround to make it work: Simply untick the option and all goes ok, except that it's not really what I expected to do.

### Debug Output
https://gist.github.com/Nordes/82bb8570aaa6407fd61898c56ddf574c

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Important Factoids
N/A

### References
N/A

jthurman42:
[...]

PR #1208, opened 2021-02-17
Improvements to newrelic_one_dashboard

jthurman42:
This may address the following issues, or if it gets too big, a few of them in the following order:
 

- [X] Tables with filters do not work #1172 
- [x] Heatmaps do not work #1181 

Added support for the additional 4 widget types (bullet, funnel, heatmap, histogram)


ctrombley (approved):
🚢

Issue #1078, opened 2021-01-05
Add nrqlDropRules to Terraform Provider


sanderblue:
### Objective
Add `nrqlDropRules` capabilities via new resource. 

There has been a request to add functionality for managing NRQL drop rules via Terraform. An example NerdGraph query can be seen [here](https://api.newrelic.com/graphiql?#query=mutation%20%7B%0A%20%20nrqlDropRulesCreate%0A%7D%0A)

jthurman42:
Changes needed for this feature were added in newrelic/newrelic-client-go#664 by @paul91, and are included in the https://github.com/newrelic/newrelic-client-go/releases/tag/v0.58.0 release.

PR #1241, opened 2021-02-23
feat: Add NRQL Drop Rule support

paul91:
Adds New Relic Terraform Provider support for managing New Relic NRQL Drop Rules (https://docs.newrelic.com/docs/accounts/accounts/data-management/drop-data-using-nerdgraph)

Feel free to suggest any nits/style changes you recommend to make it more consistent with other resources.

Should solve https://github.com/newrelic/terraform-provider-newrelic/issues/1078


newrelic/resource_newrelic_nrql_drop_rule.go
```diff
@@ -0,0 +1,196 @@
...
+			NRQL:        d.Get("nrql").(string),
+		},
+	}
+
+	created, err := client.Nrqldroprules.NRQLDropRulesCreate(accountID, createInput)
+	if err != nil {
+		return err
+	}
+
+	rule := created.Successes[0]
```
jthurman42, line 84: Should check for `nil` here as `created` is a pointer.

newrelic/resource_newrelic_nrql_drop_rule.go
```diff
@@ -0,0 +1,196 @@
...
+
+	id := fmt.Sprintf("%d:%s", created.Successes[0].AccountID, rule.ID)
```
jthurman42, line 86: Since you already extract `created.Successes[0]` to `rule`, `created.Successes[0].AccountID` could be `rule.AccountID` which would read easier.

newrelic/resource_newrelic_nrql_drop_rule.go
```diff
@@ -0,0 +1,196 @@
...
+func resourceNewRelicNRQLDropRuleRead(d *schema.ResourceData, meta interface{}) error {
+	providerConfig := meta.(*ProviderConfig)
+
+	if !providerConfig.hasNerdGraphCredentials() {
+		return errors.New("err: NerdGraph support not present, but required for Read")
+	}
+
+	client := providerConfig.NewClient
+
+	log.Printf("[INFO] Reading New Relic entity tags for entity guid %s", d.Id())
```
jthurman42, line 102: Log message for tags?

newrelic/resource_newrelic_nrql_drop_rule.go
```diff
@@ -0,0 +1,196 @@
...
+
+	_, err = client.Nrqldroprules.NRQLDropRulesDelete(accountID, deleteInput)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getNRQLDropRuleIDs(id string) (int, string, error) {
```
jthurman42, line 169: `parse` is probably a better name here, I was confused above when looking at the code why there were two `get` calls for the read.

newrelic/resource_newrelic_nrql_drop_rule_test.go
```diff
@@ -0,0 +1,100 @@
...
+		_, err = client.Nrqldroprules.GetList(accountID)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
+}
+
+func testAccNewRelicNRQLDropRuleConfig(description string) string {
```
jthurman42, line 90: It would be good to have a test for various actions, and for `account_id` inheritance.

newrelic/resource_newrelic_nrql_drop_rule.go
```diff
@@ -0,0 +1,196 @@
...
+func resourceNewRelicNRQLDropRuleCreate(d *schema.ResourceData, meta interface{}) error {
+	providerConfig := meta.(*ProviderConfig)
+
+	if !providerConfig.hasNerdGraphCredentials() {
+		return errors.New("err: NerdGraph support not present, but required for Create")
+	}
+
+	client := providerConfig.NewClient
+
+	accountID := d.Get("account_id").(int)
```
jthurman42, line 69: `account_id` should be optional in Terraform, but it is required by the API. We get around this by trying to inherit the `account_id` from the Provider configuration.  [Here](https://github.com/newrelic/terraform-provider-newrelic/blob/master/newrelic/resource_newrelic_one_dashboard.go#L360) is a recent example.

website/docs/r/nrql_drop_rule.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+  nrql = "SELECT * FROM Log WHERE container_name = 'noise'"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+  * `account_id` - (Required) Account where the drop rule will be put.
+  * `description` - (Required) The description of the drop rule.
```
jthurman42, line 33: description is optional.  Should mention that changing it forces a new rule to be created.

website/docs/r/nrql_drop_rule.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+# Resource: newrelic\_nrql\_drop\_rule
+
+Use this resource to create, and delete New Relic NRQL Drop Rules.
+
+-> **IMPORTANT!** Version 2.0.0 of the New Relic Terraform Provider introduces some [additional requirements](/docs/providers/newrelic/index.html) for configuring the provider.
+<br><br>
+Before upgrading to version 2.0.0 or later, it is recommended to upgrade to the most recent 1.x version of the provider and ensure that your environment successfully runs `terraform plan` without unexpected changes.
+
+## Example Usage
```
jthurman42, line 17: A `drop_attributes` example would be great as well, and perhaps a quick sentence about what the example does.

jthurman42 (changes requested):
@paul91 This is fantastic, thank you so much for the contribution!  I've made a few comments on areas where we know folks will ask / open follow-up tickets so we can get ahead of those.  Please let me know if you have any questions.

paul91:
@jthurman42 thanks for the thoughtful review! I believe I have addressed your feedback. Happy to make any other additional iterations as needed.

paul91:
@jthurman42 let me know if there is anything else I can do to help move this forward. Thanks!

jthurman42:
@paul91 Sorry I missed your updates... This looks great, please re-base / push and I'll get it merged in.

jthurman42 (approved):

:shipit:

Issue #1214, opened 2021-02-18
newrelic_entity is case sensitive


aelmekeev:
### Terraform Version
```
Terraform v0.14.5
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.io/newrelic/newrelic v2.18.0
```
### Affected Resource(s)
- `newrelic_entity`

### Terraform Configuration
```hcl
provider "newrelic" {
  account_id = <redacted>
}

data "newrelic_entity" "app_ca" {
  name   = "service (ca)"
  domain = "APM"
  type   = "APPLICATION"
}

data "newrelic_entity" "app_uk" {
  name   = "service (uk)"
  domain = "APM"
  type   = "APPLICATION"
}
```

```hcl
terraform {
  required_providers {
    newrelic = {
      source  = "newrelic/newrelic"
      version = "~> 2.18.0"
    }
    template = {
      source  = "hashicorp/template"
      version = "~> 2.1"
    }
  }
  required_version = ">= 0.13"
}
```

### Actual Behavior
There is a number of apps that were created quite long time ago that have names like `Service (us)` vs "modern" naming like `service (ca)`.
There is no issue in reporting the data from services that have this naming issue, i.e. service with all lower case `app_name` reports to `Service (environment)`. But there is an issue when we try to find entity with `newrelic_entity` using custom module for the app that has lowercase `name`.

```
Error: the name 'service (uk)' does not match any New Relic One entity for the given search parameters

  on main.tf line 11, in data "newrelic_entity" "app_uk":
  11: data "newrelic_entity" "app_uk" {
```

### Expected Behaviour
I'm not 100% sure if this should be considered a bug in terraform provider but:
1. alias is not helping
2. since app config already has a proper name it looks like we could potentially face an issue with renaming even through some temporary name (even we would not care about historical data) unless we remove the app completely

That's why I've been asked to raise this as an issue here by NewRelic support.

### Steps to Reproduce

1. `terraform plan`

### Debug Output
https://gist.github.com/aelmekeev/110dc5db82ab754ebadfbcaeeceefe3d

### Panic Output
N/A

### Important Factoids
N/A

### References
N/A

jthurman42:
[...]

PR #1250, opened 2021-03-01
feat(newrelic_entity): Add ignore_case to name search for entity

jthurman42:
The API does not enforce case, though the provider comparison does.  Keep the previous behavior (case-sensitive) and add a new schema attribute `ignore_case`.

Resolves: #1214


sanderblue (approved):
LGTM. I guess we kind of need to move forward with this to provide backwards compatibility.

PR #1259, opened 2021-03-03
Alert Policy / Channel fixes

### LINKED ISSUE #1130: newrelic_alert_policy_channel resource recreated every time
IT seems like when the channel_ids list is ordered in the API response some way and in the HCL config another way, the resource is re-created again and again.
The provider should use set instead of list to avoid this ordering issue.

### Terraform Version
Terraform v0.14.2
+ provider registry.terraform.io/newrelic/newrelic v2.15.0

### Affected Resource(s)
Please list the resources as a list, for example:
- `newrelic_alert_policy_channel `

### Terraform Configuration

```hcl
resource "newrelic_alert_policy_channel" "path-prod-hr-api" {
  policy_id = newrelic_alert_policy.path-prod-hr-api.id
  channel_ids = [
    newrelic_alert_channel.ops.id,
    data.newrelic_alert_channel.slack.id,
    newrelic_alert_channel.insight.id
  ]
}
```

### Actual Behavior
```
  # newrelic_alert_policy_channel.path-synthetics must be replaced
-/+ resource "newrelic_alert_policy_channel" "path-synthetics" {
      ~ channel_ids = [ # forces replacement
            2602225,
          - 3654790,
            3703482,
          + 3654790,
        ]
      ~ id          = "947883:2602225:3654790:3703482" -> (known after apply)
        # (1 unchanged attribute hidden)
    }
```

### Expected Behavior
The resource should not be changed

### Steps to Reproduce
1. Create the resource
2. change the order of the channel IDs (channel resources) in the array
3. terraform plan/apply

### LINKED ISSUE #1163: Bug: 'include_json_attachment' cannot be enabled
### Summary

The `newrelic_alert_channel` resource's `include_json_attachment` config cannot be enabled by either `1` or `"1"`; it's just stuck switched off. [Documentation link](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/alert_channel).

### Configuration

```tf
terraform {
  required_version = "~> 0.14.0"
  required_providers {
    newrelic = {
      source = "newrelic/newrelic"
      version = "~> 2.16.0"
    }
  }
}

resource "newrelic_alert_channel" "email_me" {
  name = "Email : username@gmail.com"
  type = "email"
  config {
    recipients = "username@gmail.com"
    include_json_attachment = "1"
  }
}
```

Also raised on [forum](https://discuss.newrelic.com/t/137235). Thanks for your time.

jthurman42:
It appears that the API now requires a string of `true` or `false`, this may of changed as many of the examples had `1` as the config option.

### newrelic_alert_policy

* Ensure that no matter how you say "true" or "false", it is parsed and sent to the API as one of those values.
* Add validation function (allow "0" and "1" for backwards compatibility)
* Add unit tests for both "true" and "1" of this value
* Update docs to say "true" and "false"

Resolves #1163 

### newrelic_alert_policy_channel

* Change schema from List to Set
* Add test with what should be unordered channel IDs

Resolves #1130


newrelic/resource_newrelic_alert_channel.go
```diff
@@ -152,10 +152,26 @@ func resourceNewRelicAlertChannel() *schema.Resource {
...
-							Type:        schema.TypeString,
-							Optional:    true,
-							ForceNew:    true,
-							Description: "0 or 1. Flag for whether or not to attach a JSON document containing information about the associated alert to the email that is sent to recipients.",
+							Type:         schema.TypeString,
+							Optional:     true,
+							ForceNew:     true,
+							Description:  "true or false. Flag for whether or not to attach a JSON document containing information about the associated alert to the email that is sent to recipients.",
+							ValidateFunc: validation.StringInSlice([]string{"true", "false", "1", "0"}, true),
+							DiffSuppressFunc: func(k, old, current string, d *schema.ResourceData) bool {
```
jthurman42, line 160: Ignore current `1`s  or `0`s

newrelic/structures_newrelic_alert_channel_test.go
```diff
@@ -1,4 +1,4 @@
-// +build integration
+// +build unit
```
jthurman42, line 1: not sure why this file was integration, it makes no API calls.

newrelic/resource_newrelic_alert_policy_channel.go
```diff
@@ -26,7 +26,7 @@ func resourceNewRelicAlertPolicyChannel() *schema.Resource {
 				Description: "The ID of the policy.",
 			},
 			"channel_ids": {
-				Type:        schema.TypeList,
+				Type:        schema.TypeSet,
```
jthurman42, line 29: Thought this might have to be a migration, but appears from testing that the implementation of Set vs List are compatible in the state file.

jpvajda (commented):
I have very little to add in the form of a proper review, I'm wondering if someone on the DTK team would have a moment to do a proper review.

newrelic/resource_newrelic_alert_policy_channel.go
sanderblue: Compatible in the state file yes, but the behavior is different. Any chance we need a migrate state function? - e.g. basically would be the same as `migrateStateNewRelicDashboardV0toV1` that we already have in Dashboards?

jthurman42: I wouldn't say zero chance, though testing didn't break anything.  Probably best to be safe and implement it that way (though `migrateStateNewRelicDashboardV0toV1` is a no-op, so not sure the difference besides explicitly calling out that the schema changed, which is valid

newrelic/resource_newrelic_alert_channel.go
sanderblue: Not a blocker, but just putting the thought out there... wondering if theres any difference between using `DiffSuppressFunc` vs `StateFunc` where `StateFunc` converts the boolean value to `0` or `1` to store it in the state. It probably doesn't make a difference in this case 😎

sanderblue (approved):
🐑 it

zlesnr (approved):

PR #1270, opened 2021-04-06
fix: multilocation alert violation time

max-brewer:
Updated the multilocation alert violation time limit validation to match web UI
Previously it was using a set list rather than the 5 minute to 30 day time limit that the web UI allows.

### DIFF
--- a/newrelic/resource_newrelic_synthetics_multilocation_alert_condition.go
+++ b/newrelic/resource_newrelic_synthetics_multilocation_alert_condition.go
@@ -80,8 +80,8 @@ func resourceNewRelicSyntheticsMultiLocationAlertCondition() *schema.Resource {
 			"violation_time_limit_seconds": {
 				Type:         schema.TypeInt,
 				Required:     true,
-				ValidateFunc: validation.IntInSlice([]int{0, 3600, 7200, 14400, 28800, 43200, 86400}),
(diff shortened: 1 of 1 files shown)


newrelic/resource_newrelic_synthetics_multilocation_alert_condition.go
```diff
@@ -80,8 +80,8 @@ func resourceNewRelicSyntheticsMultiLocationAlertCondition() *schema.Resource {
 			"violation_time_limit_seconds": {
 				Type:         schema.TypeInt,
 				Required:     true,
-				ValidateFunc: validation.IntInSlice([]int{0, 3600, 7200, 14400, 28800, 43200, 86400}),
-				Description:  "The maximum number of seconds a violation can remain open before being closed by the system.  Must be one of: 0, 3600, 7200, 14400, 28800, 43200, 86400",
+				ValidateFunc: validation.IntBetween(300, 2592000),
```
jthurman42, line 83: The previous validation allowed 0 (zero), so this is a breaking change for folks that may of configured that value.

xmariopereira: @jthurman42  just noticed that this somehow connected to my issue: #1519. :)  maybe the solution for this passes by to make violation_time_limit_seconds optional like in the other parts of NewRelic Alert conditions

xmariopereira: since the documentation is not up to date, I have created the #1520

Issue #1280, opened 2021-04-23
Documentation needs updating for newrelic_infra_alert_condition to support 'description' field


atawfik-elsevier:
Hi there,

v2.4.0 newrelic_infra_alert_condition resource has description added in https://github.com/newrelic/terraform-provider-newrelic/issues/779 But documentation doesn't mention it: https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/infra_alert_condition

kidk:
Thanks for reporting this @atawfik-elsevier A fix has been merged and the docs should be updated soon.

PR #1293, opened 2021-05-06
fix(infra_alert_condition): Added missing documentation

kidk:
Fix for #1280 - Documentation needs updating for newrelic_infra_alert_condition to support 'description' field


ctrombley (approved):
:shipit:

Issue #606, opened 2020-05-22
Better Infrastructure validation for `select` attribute


sanderblue:
Related : #589 

### Objective
It looks like we might need some client-side validation for the `select` attribute. 

More specifically, using a combination of `type = "infra_metric"`  with `select = "memoryUsedPercent"` does not work. 

Instead, it looks using `memoryFreeBytes/memoryTotalBytes*100` does work.

e.g.
```hcl
# ...
type       = "infra_metric"
select     = "memoryFreeBytes/memoryTotalBytes*100"
# ...
```

### Acceptance Criteria
- Some level of client-side validation is set up. 
    > <small>(or implement some other mechanism to help the user out, although I didn't seem to get errors back from the API for this scenario)</small>
- Where applicable, update the Infra Alert Condition documentation to hopefully provide better explanations/examples of the allowed combinations of values for the different types of metrics.

Benno007:
Sorry to bring up an old issue, but this would be great. I just spent quite a while trying to setup an alert for `provider.swapUsageBytes.Average`, which is the value in the dropdown in the infrastructure menu. I was just about to drop using newrelic provider altogether. Just using `provider.swapUsageBytes.Average` was enough to completely crash Terraform (0.14.11, newrelic provider 0.22.1)

I was assuming the available values would match what is in the dropdown? Is this not the case?

<img width="999" alt="Screen Shot 2021-05-17 at 8 56 44 am" src="https://user-images.githubusercontent.com/313195/118415462-d01d0800-b6ed-11eb-81f7-d47649d2b99d.png">



```
!!!!!!!!!!!!!!!!!!!!!!!!!!! TERRAFORM CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!
Terraform crashed! This is always indicative of a bug within Terraform.
A crash log has been placed at "crash.log" relative to your current
working directory. It would be immensely helpful if you could please
report the crash with Terraform[1] so that we can fix this.
```

```
2021-05-17T08:35:35.320+1000 [WARN]  plugin.stdio: received EOF, stopping recv loop: err="rpc error: code = Unavailable desc = transport is closing"
2021-05-17T08:35:35.322+1000 [DEBUG] plugin: plugin process exited: path=.terraform/providers/registry.terraform.io/newrelic/newrelic/2.22.1/darwin_amd64/terraform-provider-newrelic_v2.22.1 pid=2613
[...]

PR #1309, opened 2021-05-17
fix(newrelic_infra_alert_condition): Prevent index out of range

jthurman42:
`expandInfraAlertThreshold` assumes we have data, so validate the length of data passed is greater than zero, otherwise return `nil`

Reported in #606


sanderblue (approved):

sanderblue:
Tests failing :/

jthurman42:
Test failures were unrelated to the change, and passed before merge.

Issue #1327, opened 2021-06-02
New Relic Plugin EOL and impact


mbazhlekova:
On **Wednesday, June 16th, 2021**, New Relic will no longer support or maintain [Plugins](https://docs.newrelic.com/docs/plugins/). Any plugin that is currently in use should be replaced by the [New Relic Infrastructure agent](https://docs.newrelic.com/docs/infrastructure/infrastructure-monitoring/get-started/get-started-infrastructure-monitoring/). Along with support and maintenance, access to the current plugin pages on the New Relic One platform will be removed.

The *"newrelic_plugins_alert_condition"* as well as *"newrelic_plugin"* and *"newrelic_plugin_component"* have been marked as deprecated and will not be supported as of **June 16, 2021**. Instead you should use *"newrelic_nrql_alert_condition."*

For more information, check out [https://discuss.newrelic.com/t/new-relic-plugin-eol-wednesday-june-16th-2021/127267](https://discuss.newrelic.com/t/new-relic-plugin-eol-wednesday-june-16th-2021/127267)

As always, we’re here to help. If you have any questions, doubts or remarks don’t hesitate to respond to this ticket or reach out to your New Relic representative.

PR #1332, opened 2021-06-08
WIP Remove deprecated newrelic_plugins_alert_condition resource

jaymcgrath:
As explained in #1327, New Relic Plugins are reaching EOL June 16th, 2021. 

This seeks to replace the already deprecated newrelic_plugins_alert_condition resource with an error message, as suggested in Step 4 of the [Hashicorp best practices guide for removing resources](https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-data-source-or-resource-removal)

I've also removed references to this resource from the docs index and replaced the resource's docs page with a message about the removal.

I'm marking this as a WIP PR as I'm having some issues getting my environment back up to speed locally, and hoping that, given how it's all code/test removal and docs updates,  it just builds in the CI 🤞


jthurman42:
This is an interesting case as TF wise we should merge this into a 3.x release, but the API is going to stop working regardless...  @kidk do we have a plan for `newrelic_dashboards` as well here? Good to lump them together.

ctrombley:
We have a deprecation of the plugins resource coming through in the next couple days.  Closer to the actual EOL we'll move forward with a more definitive change like this one.  Thanks @jaymcgrath!

cc  @mbazhlekova @kidk

website/docs/d/plugin.html.markdown
```diff
@@ -21,27 +21,6 @@ data "newrelic_plugin" "foo" {
   guid = "com.example.my-plugin"
 }
```
ctrombley, line 23: Add an EOL message to this data source similar to the one we added for the resource?

ctrombley:
Closing in favor of #1350.

PR #1375, opened 2021-06-25
feat: 0 violation TTL for Infra Conditions returns warning

nathanweddle:
1. Updated `resource_newrelic_infra_alert_condition` to return a warning if a `0` is supplied. `0` is no longer valid.
2. If a `0` is applied, we will replace the value with the default, which is 24.
3. Updated `violation_close_timer` text in the docs.  A TTL will be required moving forward.
4. Added valid values and unit of measure (hours)
5. Added default value

### DIFF
--- a/newrelic/resource_newrelic_infra_alert_condition.go
+++ b/newrelic/resource_newrelic_infra_alert_condition.go
@@ -164,8 +164,8 @@ func resourceNewRelicInfraAlertCondition() *schema.Resource {
 				Type:         schema.TypeInt,
 				Optional:     true,
 				Default:      24,
-				ValidateFunc: intInSlice([]int{0, 1, 2, 4, 8, 12, 24, 48, 72}),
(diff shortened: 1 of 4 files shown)


nathanweddle:
I had a lot of trouble getting my code to commit was getting this when I tried to commit. My commit message was valid but I kept seeing this
```
git commit -m "feat(resource_newrelic_infra_alert_condition): 0 violation TTL for Infra Conditions returns warning"

Checking commit message format...

=== terraform-provider-newrelic === [ tools            ]: Installing tools required by the project...
=== terraform-provider-newrelic === [ deps             ]: Installing package dependencies required by the project...
=== terraform-provider-newrelic === [ lint-commit      ]: Checking that commit messages are properly formatted (go-gitlint)...
dbbe108: subject does not match regex [(chore|docs|feat|fix|refactor|tests?)(\([^\)]+\))?: .*]
make: *** [lint-commit] Error 1
```

PR #1425, opened 2021-08-27
fix(one_dashboard): Make billboard thresholds optional.

a-feld:
Currently, billboard widgets will default to having a warning and critical threshold of 0. This means that by default billboards with any value above 0 will display with a red "critical error" background.

**There is currently no way to specify "no critical / warning threshold" for billboards**

This PR changes the structure of billboards so that the thresholds are defined in a map. Example:

```terraform
resource "newrelic_one_dashboard" "exampledash" {
  name = "New Relic Terraform Example"
  page {
    name = "New Relic Terraform Example"
    widget_billboard {
      row    = 1
      column = 1
      title  = "Example"
      nrql_query { query = "FROM NrUsage SELECT count(*)" }
      threshold = {
        warning  = 1
        critical = 2
      }
    }
  }
}
```

image

### DIFF
--- a/newrelic/resource_newrelic_one_dashboard.go
+++ b/newrelic/resource_newrelic_one_dashboard.go
@@ -3,6 +3,7 @@ package newrelic
 import (
 	"context"
 	"log"
+	"regexp"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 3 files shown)


kidk:
Thanks @a-feld We need to find a way to implement this without changing the structure of the resource. 

This PR would mean all customers currently using this resource would need to migrate and change it. We try to avoid this as much as possible. 

I wonder if the threshold warning and error can be set to a nil value.

I'll have some more time in October to look at this.

kidk:
This has been fixed in v2.34.1

PR #1435, opened 2021-09-15
feat(alerts): update docs with streaming methods

akane0915:
[JIRA ticket](https://newrelic.atlassian.net/browse/AINTER-7817)

This PR will update the documentation for NRQL Alert conditions to describe the new Streaming Methods fields.  Note that there was no prior documentation for the `signal` field (streaming methods fields are within the signal field), so I added this as well.

### DIFF
--- a/website/docs/r/nrql_alert_condition.html.markdown
+++ b/website/docs/r/nrql_alert_condition.html.markdown
@@ -59,6 +59,12 @@ resource "newrelic_nrql_alert_condition" "foo" {
     threshold_duration    = 600
     threshold_occurrences = "ALL"
   }
+
+  signal {
+    aggregation_window = 60
+    aggregation_method = "event_flow"
+    aggregation_delay = 120
+  } 
 }
 ```
 See additional [examples](#additional-examples).
@@ -95,7 +101,7 @@ The following arguments are supported:
 - `expiration_duration` - (Optional) The amount of time (in seconds) to wait before considering the signal expired.
 - `open_violation_on_expiration` - (Optional) Whether to create a new violation to capture that the signal expired.
 - `close_violations_on_expiration` - (Optional) Whether to close all open violations when the signal expires.
-
+- `signal` - (Optional) Configuration that defines the signal that the NRQL condition will use to evaluate. See [SIGNAL](#signal) below for details.
 ## NRQL
 
 The `nrql` block supports the following arguments:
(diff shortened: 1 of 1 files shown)


website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -127,6 +133,21 @@ The `term` block the following arguments:
...
+The stream of telemetry data that's watched and alerted on. [Configuration of the signal](https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/alert-conditions/create-nrql-alert-conditions/#advanced-signal) defines what the NRQL condition will use to evaluate.
+
+For more information, see [Streaming Alerts](https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/advanced-alerts/understand-technical-concepts/streaming-alerts-key-terms-concepts/).
+
+The `signal` block supports the following arguments:
+
+- `aggregation_window` - controls the duration of the time window used to evaluate the NRQL query, in seconds. The value must be at least 30 seconds, and no more than 15 minutes (900 seconds). Default is 60 seconds.
+- `fill_option` - determines the type of value that should be used to fill gaps (empty windows). Defaults to `STATIC`. Outlier NRQL conditions may only use `NONE`.
+- `fillValue` - used for filling if using the `STATIC` fill option. Defaults to 0.
+- `aggregation_method` - determines when we consider an aggregation window to be complete so that we can evaluate the signal for violations. Default is `CADENCE`.
```
founddrama, line 147: @akane0915 IIRC `CADENCE` is the default "for now" with `EVENT_FLOW` to become the default shortly after GA. Do we have follow-up work logged for that?

founddrama: Should also enumerate the options here like we do with `fill_option`

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -127,6 +133,21 @@ The `term` block the following arguments:
...
+## Signal
+
+The stream of telemetry data that's watched and alerted on. [Configuration of the signal](https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/alert-conditions/create-nrql-alert-conditions/#advanced-signal) defines what the NRQL condition will use to evaluate.
+
+For more information, see [Streaming Alerts](https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/advanced-alerts/understand-technical-concepts/streaming-alerts-key-terms-concepts/).
+
+The `signal` block supports the following arguments:
+
+- `aggregation_window` - controls the duration of the time window used to evaluate the NRQL query, in seconds. The value must be at least 30 seconds, and no more than 15 minutes (900 seconds). Default is 60 seconds.
+- `fill_option` - determines the type of value that should be used to fill gaps (empty windows). Defaults to `STATIC`. Outlier NRQL conditions may only use `NONE`.
```
founddrama, line 145: @akane0915 if we scroll up earlier in the docs, `fill_option` has some description text already -- we should probably use the same text as that

And on that note: we should make sure to enumerate the options: `static` + `none` + `last_value` (**WHICH!** it looks like Terraform wants these values to be lowercase vs. GraphQL which has them in upper?)

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -127,6 +133,21 @@ The `term` block the following arguments:
...
+- `fillValue` - used for filling if using the `STATIC` fill option. Defaults to 0.
+- `aggregation_method` - determines when we consider an aggregation window to be complete so that we can evaluate the signal for violations. Default is `CADENCE`.
+- `aggregation_delay` - how long we wait for data that belongs in each aggregation window. Depending on your data, a longer delay may increase accuracy but delay notifications. Use `aggregationDelay` with the `EVENT_FLOW` and `CADENCE` methods. The maximum delay is 1200 seconds (20 minutes) when using `EVENT_FLOW` and 3600 seconds (60 minutes) when using `CADENCE`. In both cases, the minimum delay is 0 seconds and the default is 120 seconds.
```
founddrama, line 148: @akane0915 there's an `aggregationDelay` in the middle of the description that should be `aggregation_delay`

also: can we confirm whether the methods are ALL_CAPS or lower_case for Terraform? If they're ALL_CAPS then we don't need to change them… otherwise we're looking at updating to `cadence` and `event_flow`

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -127,6 +133,21 @@ The `term` block the following arguments:
...
+- `aggregation_timer` - how long we wait after each data point arrives to make sure we've processed the whole batch. Use `aggregationTimer` with the `EVENT_TIMER` method. The timer value can range from 0 seconds to 1200 seconds (20 minutes); the default is 60 seconds.
```
founddrama, line 149: @akane0915 there's an `aggregationTimer` in the middle of the description that should be `aggregation_timer`

also: can we confirm whether the methods are ALL_CAPS or lower_case for Terraform? If they're ALL_CAPS then we don't need to change them… otherwise we're looking at updating to `event_timer`

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -59,6 +59,12 @@ resource "newrelic_nrql_alert_condition" "foo" {
     threshold_duration    = 600
     threshold_occurrences = "ALL"
   }
+
+  signal {
+    aggregation_window = 60
+    aggregation_method = "event_flow"
+    aggregation_delay = 120
+  }
```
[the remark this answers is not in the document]
founddrama: And depending on where we land re: nixing `evaluation_offset`, we should update the baseline + outlier condition examples as well. (Open question about the "1.x to 2.x" upgrade section and its examples…)

founddrama (changes requested):
@akane0915 Left a bunch on in-line comments around some items that we should address. There are some questions I have about whether some arguments are being _moved_ to the `signal` block, and/or if we're supporting them in "both" places. There are some other questions around making sure that we're using the right upper/lowercase namings for things as well as snake_case vs. camelCase.

PR #1439, opened 2021-09-21
Update newrelic-client-go v0.63.0

jthurman42:
This version of the client has some breaking changes, so manually updating.


sanderblue (approved):
🐑 it

PR #1448, opened 2021-09-27
feat(servicelevel): Add service level resource

asanuy:
- Bump newrelic-client-go version
- Add new resource to manage service levels

### DIFF
--- a/newrelic/resource_newrelic_service_level.go
+++ b/newrelic/resource_newrelic_service_level.go
@@ -1,0 +1,302 @@
+package newrelic
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/newrelic/newrelic-client-go/pkg/common"
+	"github.com/newrelic/newrelic-client-go/pkg/errors"
+)
+
+func resourceNewRelicServiceLevel() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceNewRelicServiceLevelCreate,
+		ReadContext:   resourceNewRelicServiceLevelRead,
+		UpdateContext: resourceNewRelicServiceLevelUpdate,
+		DeleteContext: resourceNewRelicServiceLevelDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
+			"guid": {
+				Type:         schema.TypeString,
+				Required:     true,
+				Description:  "",
+				ValidateFunc: validation.StringIsNotWhiteSpace,
+			},
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				Description:  "",
+				ValidateFunc: validation.StringIsNotWhiteSpace,
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "",
+			},
+			"events": {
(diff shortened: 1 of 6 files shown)


jthurman42:
@asanuy Thanks for the contribution.  A few items to resolve:

* Lint check is [failing](https://github.com/newrelic/terraform-provider-newrelic/pull/1448/checks?check_run_id=3746640197)
* Need to add this resource to the documentation (see [workloads](https://github.com/newrelic/terraform-provider-newrelic/blob/main/website/docs/r/workload.html.markdown) for an example)

newrelic/resource_newrelic_service_level.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	identifier := serviceLevelIdentifier{
+		AccountID: createInput.Events.AccountID,
+		ID:        created.ID,
+		GUID:      guid,
+	}
+
+	d.SetId(identifier.String())
+	d.Set("sli_id", created.ID)
```
jthurman42, line 213: This fails linting, and while I dislike the practice, historically we change this to read `_ = d.Set(....)`

newrelic/resource_newrelic_service_level.go
```diff
@@ -0,0 +1,302 @@
...
+			"events": {
+				Type:        schema.TypeList,
+				Required:    true,
+				Description: "",
+				MinItems:    1,
+				MaxItems:    1,
+				Elem:        eventsSchema(),
+			},
+			"objective": {
+				Type:        schema.TypeSet,
```
sanderblue, line 53: Not a blocker by any means, but out of curiosity, could this also be `schema.TypeList`? I lack the domain knowledge of the `serviceLevel` feature, and the main reason I ask is because `TypeSet` has given us some "interesting" behavior to deal with in some scenarios, where often times a `TypeList` works just fine and is a little easier to work with when evaluating state changes.

sanderblue: Also the field in the GraphQL mutation is plural `objectives`. We've typically tried to make sure Terraform attributes match the API fields to continuity across products 🙂

asanuy: Hello @sanderblue, thanks for the feedback!

Regarding the `TypeSet`, ideally I'd prefer to model it as a set because it wouldn't make much sense to have two objectives that are exactly the same. For instance, you would want one objective that is "realistic" with a target of 95% and another one that is "optimistic" with a target of 99%.

Regarding the naming of the `objectives`, as far as I could see in the documentation, we usually use a singular name for blocks, for instance, with [Entity tags](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/entity_tags#tag) or [Workload search queries](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/workload#entity_search_query). In fact, the naming in singular sounds more natural to me considering that you can define multiple blocks, eg:

```
resource "newrelic_service_level" "foo" {
   ...
   objective { ... }
   objective { ... }
}
```

Nevertheless, I am happy to change it if you think that pluralizing the name is the recommended approach 🙂

sanderblue: Fair enough. Thanks for the thought process behind the decision. Looks good to me.

sanderblue (approved):
Thanks for the contribution! Looks good 😎

rdhar:
Where can I find documentation to learn more about "service level" resource?

asanuy:
Hello @rdhar, thanks for your interest.

Regarding your question about the documentation, unfortunately it is not public available yet. Service levels product hasn't been released to general public.

Out of curiosity, would you be particularly interested on it? 🙂

rdhar:
Thanks for getting back to me, @asanuy, much appreciated.

Certainly, more than happy to learn more about the product and what it provides. At my organisation, I'm responsible to driving automation and Terraform has been a huge part of that push. As a result, always keen to find out what else can be accomplished at scale using Terraform.

PR #1540, opened 2021-11-29
Add password for monitor script

mbazhlekova:
We've had some requests to support a password field for private monitor script locations and calculate the HMAC internally. Previously we were just accepting the already calculated HMAC and passing it through. This change removes the old field in favor of just using password. I didn't find a good way to keep both and have them work. Open to any suggestions.


newrelic/resource_newrelic_synthetics_monitor_script.go
```diff
@@ -42,10 +45,11 @@ func resourceNewRelicSyntheticsMonitorScript() *schema.Resource {
 							Required:    true,
 							Description: "The monitor script location name",
 						},
-						"hmac": {
+						"vse_password": {
```
sanderblue, line 48: Hmm, I could be missing some context here, but are we fully replacing the `hmac` attribute? If so, this will be a breaking change. 

To make this backwards compatible, we could add the `vse_password` attribute while still keeping `hmac`, and then we can add the `Deprecated` schema option on the `hmac` attribute to let folks know it's being removed/replaced.

mbazhlekova: @sanderblue Yeah this would be a breaking change. Will having the two values conflict? Only one should be set.

kidk: I would add both, make it clear in the description/docs, and do validation in the structure like here: https://github.com/newrelic/terraform-provider-newrelic/blob/main/newrelic/structures_newrelic_nrql_alert_condition.go#L94

PR #1566, opened 2021-12-27
Fix accepted values for violation_time_limit_seconds

mbazhlekova:


sanderblue (approved):

teru01:
@mbazhlekova 
Why was it reverted?

cacao12:
Hi @mbazhlekova, just wondering the same thing as @teru01 It seems the current limits are inconsistent with the UI. 

Also if you import an already created 'alert_condition', doing a `terraform plan` shows that these values are successfully imported into the state file but terraform wants to change them to a value within the range [here](https://github.com/newrelic/terraform-provider-newrelic/pull/1566/files#diff-7d9824cfcb2b7af0bdfe77851f06f867e360c047fe3d8dc73ffb0364705b8881L85) 

Example output from the plan
image

PR #1586, opened 2022-01-13
docs: add example of workload using tags

lucasmessinar:
Adding a workload example using tags in the dynamic query for more clarification in this topic.


kidk:
Thanks for taking the time to submit this @lucasmessinar 👍 

The team will review as soon as possible.

PR #1608, opened 2022-01-27
Add AWS Cloud Link Account resource

mbazhlekova:


website/docs/r/cloud_aws_link_account.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+  metric_collection_mode = "PUSH"
+  name = "account name"
+}
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `arn` - (Required) The Amazon Resource Name (ARN) of the IAM role.
+* `metric_collection_mode` - (Optional) How metrics will be collected. One of `PUSH` or `PULL`
```
kidk, line 31: Can we clarify the difference here? Is `PUSH` metricstream? and `PULL` the integrations? I would also make it clear that metricstream is the advised method. As the individual integrations will be deprecated in the future.

website/docs/r/cloud_aws_link_account.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The ID of the AWS linked account.
+
+## Import
+
+Linked AWS accounts can be imported using the `id`, e.g.
```
kidk, line 42: Where can the user find this id?

website/docs/r/cloud_aws_link_account.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+layout: "newrelic"
+page_title: "New Relic: newrelic_cloud_aws_link_account"
+sidebar_current: "docs-newrelic-resource-cloud-aws-link-account"
+description: |-
+    Link an AWS account to New Relic.
+---
+
+# Resource: newrelic\_cloud\_aws\_link\_account
+
+Use this resource to link an AWS account to New Relic.
```
kidk, line 11: Can we also add a link to https://docs.newrelic.com/docs/infrastructure/amazon-integrations/aws-integrations-list/aws-metric-stream/#set-up-metric-stream and https://docs.newrelic.com/docs/infrastructure/amazon-integrations/get-started/connect-aws-new-relic-infrastructure-monitoring? We should probably also clarity the difference between the two with doc link.

website/docs/r/cloud_aws_link_account.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+
+## Prerequisite
+
+Setup is required in AWS for this resource to work properly. Complete the [steps outlined here](https://docs.newrelic.com/docs/infrastructure/amazon-integrations/get-started/connect-aws-new-relic-infrastructure-monitoring#connect) before using this resource.
```
kidk, line 15: We need to also add this option: https://docs.newrelic.com/docs/infrastructure/amazon-integrations/aws-integrations-list/aws-metric-stream/#set-up-metric-stream here.

newrelic/resource_newrelic_cloud_aws_link_account_test.go
```diff
@@ -0,0 +1,119 @@
...
+		metric_collection_mode = "PUSH"
+		name = "%[1]s"
+	}
+	`, name, arn)
+}
+
+func testAccNewRelicAwsLinkAccountConfigUpdated(name string, arn string) string {
+	return fmt.Sprintf(`
+	resource "newrelic_cloud_aws_link_account" "foo" {
+		arn = "%[2]s"
```
kidk, line 114: Do we need a `PULL` test? Or would you expect it to behave the same?

mbazhlekova: It behaves the same from the API

mbazhlekova:
> Been testing this full end to end and I wonder if we need a retry mechanism. Because it takes some time for AWS to apply policies I usually get this message:
> 
> ```
> ╷
> │ Error: ERR_INVALID_DATA Validation failed: The ARN you entered does not permit the correct access to your AWS account - please check the ARN and try again.
> │
> │   with newrelic_cloud_aws_link_account.newrelic_cloud_integration,
> │   on cloud-integration-aws.tf line 77, in resource "newrelic_cloud_aws_link_account" "newrelic_cloud_integration":
> │   77: resource "newrelic_cloud_aws_link_account" "newrelic_cloud_integration" {
> │
> ╵
> ```
> 
> It works after a retry, so I'm wondering if we try for a couple of times before failing. As it always works on the second apply.

I added a retry around the create and it will Retry for 10 seconds on the ARN error. I'm not sure if this is the best solution because it just matches on the error message [https://github.com/newrelic/terraform-provider-newrelic/pull/1608/files#diff-e7f2180b00f9dc0400207dc507066a63d1b515cee7da2daf4acabd59f12d2ca5R69](https://github.com/newrelic/terraform-provider-newrelic/pull/1608/files#diff-e7f2180b00f9dc0400207dc507066a63d1b515cee7da2daf4acabd59f12d2ca5R69)

It resolved the error on create.

newrelic/resource_newrelic_cloud_aws_link_account.go
```diff
@@ -0,0 +1,215 @@
...
+					return resource.RetryableError(fmt.Errorf("%s : %s", err.Type, err.Message))
+				}
+				diags = append(diags, diag.Diagnostic{
+					Severity: diag.Error,
+					Summary:  err.Type + " " + err.Message,
+				})
+			}
+		}
+
+		d.SetId(strconv.Itoa(cloudLinkAccountPayload.LinkedAccounts[0].ID))
```
kidk, line 81: Should we do a quick length check here?

kidk (approved):

ChandraIragavarapu (approved):
looks good

Issue #1611, opened 2022-01-28
Invalid nrql in drop-rule causes plugin panic


kevinfrommelt-wf:
### Terraform Version
Terraform v0.15.1
on darwin_amd64
+ provider registry.terraform.io/newrelic/newrelic v2.34.1

### Affected Resource(s)
- `newrelic_nrql_drop_rule`

### Terraform Configuration
```hcl
provider "newrelic" {
  api_key       = "<key>"
  admin_api_key = "<key>"

  account_id = <account_id>
  region     = "US"
}

resource "newrelic_nrql_drop_rule" "invalid_nrql" {
  account_id  = <account_id>
  description = "Drop rule with invalid nrql"
  action      = "drop_attributes"
  nrql        = <<EOT
FROM ContainerSample
SELECT commandLine"
EOT
}
```

### Actual Behavior
When applying a drop-rule with an invalid query, the plugin panics.

### Expected Behavior
The plugin does not panic and returns/logs an error.

### Steps to Reproduce
1. `terraform apply`

### Debug Output
https://gist.github.com/kevinfrommelt-wf/35233d6723d666833ef1b41f0420ec0e

### Panic Output
https://gist.github.com/kevinfrommelt-wf/35233d6723d666833ef1b41f0420ec0e

### Important Factoids
None

### References
Seems to be failing here: https://github.com/newrelic/terraform-provider-newrelic/blob/v2.34.1/newrelic/resource_newrelic_nrql_drop_rule.go#L89

kidk:
This has been fixed in our latest release.

PR #1632, opened 2022-02-04
Fix for #1611 - Invalid nrql in drop-rule causes plugin panic

kidk:


kidk:
This is ready for review @mbazhlekova

mbazhlekova (approved):

mbazhlekova (approved):

PR #1656, opened 2022-02-17
feat(Alerts): deprecate nrql condition value_function, sum, single_value

akane0915:
# Description

https://newrelic.atlassian.net/browse/AINTER-8206

Corresponding Go Client ticket https://github.com/newrelic/newrelic-client-go/pull/850

**This PR requires a MINOR release and should ensure the changelog has an entry noting NRQL condition  deprecation of fields `value_function`, `sum`, and `single_value`.**

This PR should:
- Make nrql condition `value_function` optional for create and update
- Add deprecation warning messages for `sum`, `single_value`, and `value_function`
- Update documentation
- Ensure the changelog has an entry noting the deprecation
- Be released as a MINOR release

Deprecation warning example:
image

## Type of change

## Checklist:

## How to test this change?

I added integration tests.  Also, please manually test.  See below.

To manually test:
- Create a static condition without a value function (the condition-api will default to `single_value` when it receives no value function)
- Update that condition to add a `single_value` value function.  Note that there are no changes because `single_value` was already set in the condition api.  There should be no errors.
- Update that condition to add a `sum` value function.  You should be able to successfully to this. 

Then try:
- Create a static condition with a `sum` value function
- Update that condition to remove the value function and add slide by:
image

Also please:
- Ensure you see the deprecation message for value function
- Confirm documentation reflects deprecation and optional value function

## Implementation notes:
- Left `omitEmpty` in Go Client update struct
- Did not add `computed` on Provider update struct
- Added a special line that in the absence of a value function, we send “single value” by default in the expand method on update.  This is necessary because we could not remove `omitEmpty` from the client update struct.
- Added a diff suppress function so that if there is no value_function configured and the API returns “single_value” we don’t generate a diff
- Removed warning that compelled a user to include a value_function when creating static conditions

### DIFF
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -244,12 +244,14 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 			// between new:old and old:new is handled via maps in structures file.
 			// Conflicts with `baseline_direction` when using NerdGraph.
 			"value_function": {
+				Deprecated:   "'value_function' is deprecated.  Remove this field and condition will evaluate as 'single_value' by default.  To replicate 'sum' behavior, use 'slide_by'.",
 				Type:         schema.TypeString,
 				Optional:     true,
-				Description:  "Valid values are: 'single_value' or 'sum'",
+				Description:  "Values are: 'single_value' (deprecated) or 'sum' (deprecated)",
 				ValidateFunc: validation.StringInSlice([]string{"single_value", "sum"}, true),
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					return strings.EqualFold(old, new) // Case fold this attribute when diffing
+					// If a value is not provided and the condition uses the default value (single_value), don't show a diff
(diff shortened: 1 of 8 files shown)


jaymcgrath (approved):
This PR looks great! You've remembered the stuff I usually forgot (docs and examples) and provided excellent reasoning and tests for the changes you're introducing. 

We're going to do some manual testing today to validate that making this field optional doesn't have any unintended effects, but this looks good to me. :shipit:

akane0915:
@jaymcgrath I updated this deprecation message.  Thoughts? https://github.com/newrelic/terraform-provider-newrelic/pull/1656/files#diff-39c4f39fbb1990471e6b8f4fdf080932cff74489028d3f898e862e12d4f0086bR247

jaymcgrath:
> @jaymcgrath I updated this deprecation message. Thoughts? https://github.com/newrelic/terraform-provider-newrelic/pull/1656/files#diff-39c4f39fbb1990471e6b8f4fdf080932cff74489028d3f898e862e12d4f0086bR247

It's great! I think it's very direct and offers a clear solution to what to do for either value of the deprecated field.

jaymcgrath (approved):
Re-approving after deprecation message update

kidk (approved):

Issue #1709, opened 2022-03-17
Creating newrelic_cloud_azure_link_account Fails during Apply


ironf1st:
Hi there,

Thank you for opening an issue. In order to better assist you with your issue, we kindly ask to follow the template format and instructions. Please note that we try to keep the Terraform issue tracker reserved for **bug reports** and **feature requests** only. General usage questions submitted as issues will be closed and redirected to New Relic's Explorers Hub https://discuss.newrelic.com/c/build-on-new-relic/developer-toolkit.

## Please include the following with your bug report

> :warning: **Important:** Failure to include the following, such as omitting the Terraform configuration in question, may delay resolving the issue.

- [x] Your New Relic `provider` [configuration](#terraform-configuration-files) (sensitive details redacted)
- [x] A list of [affected resources](#affected-resources) and/or data sources
- [x] The [configuration](#terraform-configuration-files) of the resources and/or data sources related to the bug report (i.e. from the list mentioned above)
- [x] Description of the [current behavior](#actual-behavior) (the bug)
- [x] Description of the [expected behavior](#expected-behavior)
- [x] Any related [log output](#debug-output)

### Terraform Version

Terraform v1.1.5
+ provider registry.terraform.io/hashicorp/azurerm v2.98.0
+ provider registry.terraform.io/newrelic/newrelic v2.40.0

### Affected Resource(s)
newrelic_cloud_azure_link_account

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration
> Please include your `provider` configuration (sensitive details redacted) as well as the configuration of the resources and/or data sources related to the bug report.
```hcl
resource "newrelic_cloud_azure_link_account" "self" {
  for_each = module.sub_info.name_map # This module output gives a map of maps with the subscription name as the key for the second map and the values are the output of the subscriptions block from the azurerm_subscriptions data source.

  account_id       = var.account_id
  application_id   = var.app_id
  client_secret_id = var.secret_id
  subscription_id  = each.value.subscription_id
  tenant_id        = var.tenant_id
  name             = each.key
}
```

### Actual Behavior
Terraform crashed on apply when creating 22 `newrelic_cloud_azure_link_account` resources. However each time some resources would apply successfully. After 2 applies that deployed 18 of the 22 resources successfully, every apply fails when creating the first two resources.

### Expected Behavior
All 22 resources should have been created.

### Steps to Reproduce

[...]

PR #1710, opened 2022-03-18
fix(newrelic_cloud): only set Id after error check

kidk:
# Description

Possible fix for #1709

We should only set the ID if we haven't received an error back from the API. My only concern here is that the API could return an error, but still create the resource, leaving Terraform in a bad state. 

## Type of change

## Checklist:


kidk:
Closing this for now. Need to do some more research.

Issue #1556, opened 2021-12-10
Error: expected entity tag team to have been updated but was not found


thedebugger:
## Please include the following with your bug report

> :warning: **Important:** Failure to include the following, such as omitting the Terraform configuration in question, may delay resolving the issue.

- [x ] Your New Relic `provider` [configuration](#terraform-configuration-files) (sensitive details redacted)
- [ x] A list of [affected resources](#affected-resources) and/or data sources
- [x ] The [configuration](#terraform-configuration-files) of the resources and/or data sources related to the bug report (i.e. from the list mentioned above)
- [x ] Description of the [current behavior](#actual-behavior) (the bug)
- [ x] Description of the [expected behavior](#expected-behavior)
- [ x] Any related [log output](#debug-output)

### Terraform Version
```
Terraform v1.0.11
on linux_amd64
+ provider registry.terraform.io/newrelic/newrelic v2.34.0
```

### Affected Resource(s)
- `newrelic_entity_tags.nr_dashboard_tags`

issue seen  v2.30.0,  v2.30.2 and  v2.34.0

### Terraform Configuration
https://gist.github.com/thedebugger/2e26fbb8ca484d464a0701f98f8a0d27

### Actual Behavior
terraform errored with " expected entity tag team to have been updated but was not found"

### Expected Behavior
terraform should update the nr_dashboard_tags accordingly

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create nr_one_dashboard and nr_dashboard_tags
2. Delete the dashboard resource manually from state file and from NR (or manually change the nr_dashboard_tags to something non-existent)
3. run terraform apply

### Debug Output
Can provide debug output if required since we have hundred of resources. But here are the info logs https://gist.github.com/thedebugger/2e26fbb8ca484d464a0701f98f8a0d27

### Important Factoids
no

### References
* Most likely it is failing here https://github.com/newrelic/terraform-provider-newrelic/blob/aa6ceb1e2ef89f0c369f3c4f059646615afb2faa/newrelic/resource_newrelic_entity_tags.go#L185

bkalcho:
I am experiencing the same error in newrelic provider version `2.46.2`

gelato:
Still not working at 2.49.1 - more to it, now I can't use my module because each apply introduces same changes because of tainted tag resources.

kidk:
Thanks all. We will take a look.

matthew-mbo:
[...]

PR #1756, opened 2022-04-14
fix: Tags resource implicit dependency

asanuy:
## Description
Very often tags resource is used referencing the dependency (i.e. entity) implicitly. If that entity is re-created, the tags resource doesn't refresh and continues referencing to the old entity GUID.

I believe that the fix should also address the following issue: https://github.com/newrelic/terraform-provider-newrelic/issues/1556

## Type of change

## Checklist:


mbazhlekova (approved):
🚀

PR #1769, opened 2022-04-21
fix(one_dashboard): Make limit for bullet widget required

mbazhlekova:
# Description

There is a mismatch between Terraform's schema and the API. The Dashboards API requires limit while it's optional in Terraform leading to confusion. 

Fixes # (issue)

## Type of change

## Checklist:

## How to test this change?

- Create a `one_dashboard` resource with a `widget_bullet` type but without a `limit` value.
- Terraform apply

We should see an error that this value is required.


kidk (approved):

kidk:
:shipit:

PR #1800, opened 2022-05-11
docs(provider): add info regarding setting/upgrading the provider version

sanderblue:


website/docs/guides/provider_configuration.html.markdown
```diff
@@ -10,12 +10,50 @@ description: |-
 
 The New Relic Terraform Provider supports two methods of configuring the provider.
 
-1. [Using environment variables](#configuration-via-environment-variables) (recommended)
 1. [Using the `provider` block](#configuration-via-the-provider-block)
+2. [Using environment variables](#configuration-via-environment-variables) (recommended)
```
kidk, line 14: You mention we recommend to use environment variables below. So it would move to to `1.` here.

website/docs/guides/provider_configuration.html.markdown
```diff
@@ -10,12 +10,50 @@ description: |-
...
+
+
+```hcl
+# Configure terraform
+terraform {
+  required_version = "~> 1.0"
+  required_providers {
+    newrelic = {
+      source  = "newrelic/newrelic"
+      version = "~> 2.44"
```
kidk, line 30: Same comment as above on version

website/docs/guides/provider_configuration.html.markdown
sanderblue: I honestly think the environment variables part created more toil for customers because it involves configuring the provider in multiple documents so I also removed "recommended". I think having the configuration docs start with everything in one file makes things easier to digest, especially for a new customer.

website/docs/guides/getting_started.html.markdown
```diff
@@ -40,6 +40,7 @@ terraform {
   required_providers {
     newrelic = {
       source  = "newrelic/newrelic"
+      version = " ~> 2.44"
```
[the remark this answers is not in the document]
sanderblue: Fair point

kidk (approved):

PR #1817, opened 2022-05-24
feat(synthetics): migrate `newrelic_sythetics_monitor_location` data source from legacy REST API to GraphQL API

sanderblue:
The legacy Synthetics REST API has been fully deprecated and will reach EOL soon. This PR migrates the monitor location data source to the new GraphQL endpoints.

Questions:
- Can we add a `tags` attribute to help narrow search results since multiple private locations can have the same name? Since this is additive, it's not considered a breaking change.
- Does this migration to the new API constitute a breaking change and major version bump given the response data no longer contains data for the computed attributes in the data source?

Checklist:
- [ ] Add integration test case to test the negative scenarios (e.g. entity not found and error scenarios)
- [ ] Deprecate legacy computed attributes


sanderblue:
Closing in favor of #1827

PR #1620, opened 2022-02-02
#1599

Fallenstedt:
Work in progress

See https://github.com/newrelic/terraform-provider-newrelic/issues/1599

PR #1902, opened 2022-07-27
fix(nrql_alert_condition): do not show diff for streaming methods when not provided and defaults are used

### LINKED ISSUE #1900: newrelic_nrql_alert_condition: aggregation_method and aggregation_delay show diff when not provided
### Terraform Version
```
Terraform v1.2.5
on darwin_amd64
+ provider registry.terraform.io/newrelic/newrelic v2.49.1
```

### Affected Resource(s)
- `newrelic_nrql_alert_condition`

### Terraform Configuration

```hcl
provider "newrelic" {
  account_id    = var.account_id
  admin_api_key = var.admin_api_key
  api_key       = var.api_key
  region        = var.region
}
```
Example:
```hcl
resource "newrelic_nrql_alert_condition" "max_critical" {
  policy_id                    = var.policy_id_critical
  type                         = "static"
  name                         = "Test alert"
  description                  = "Test alert description"
  runbook_url                  = "tbd"
  enabled                      = true

  nrql {
    query = local.running_node_count
  }

  critical {
    operator              = "equals"
    threshold             = var.max_node_count
    threshold_duration    = 300
    threshold_occurrences = "ALL"
  }
}
```

### Actual Behavior
Subsequent terraform plans will show diff for `aggregation_delay` and `aggregation_method`, if they are not provided.

### Expected Behavior
Should not show diff for `aggregation_delay` and `aggregation_method` if they are not provided. Instead, use their defaults.

### Steps to Reproduce
1. Create a resource as provided above, without deprecated `evaluation_offset`.
1. `terraform apply`
2. `terraform plan`
3.  Observe the following diff:
```hcl
Terraform will perform the following actions:

  # newrelic_nrql_alert_condition.max_critical will be updated in-place
  ~ resource "newrelic_nrql_alert_condition" "max_critical" {
      - aggregation_delay              = "120" -> null
      - aggregation_method             = "event_flow" -> null
        id                             = "1024960:26935418"
        name                           = "Test alert"
        # (16 unchanged attributes hidden)

        # (3 unchanged blocks hidden)
    }
```

### Workaround
Set these attributes with default values:
```hcl
aggregation_delay  = 120
```

### References
Same issue and fix attempt from the past:
- https://github.com/newrelic/terraform-provider-newrelic/issues/1544
- https://github.com/newrelic/terraform-provider-newrelic/pull/1574

antonblr:
# Description

Fixes #1900

When the `aggregation_method`, `aggregation_delay` and `aggregation_timer` fields are not provided in Terraform config it uses default values. This currently results in `terraform plan` diff, even though the values are not changing.

Changes:
* `sinceValueExists` currently evaluates as empty sting with `Get()` (and hence it is never `nil`) (added in [#1620](https://github.com/newrelic/terraform-provider-newrelic/pull/1620)). Switched to `GetOk()`. From [ResourceData.Get](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/schema#ResourceData.Get):
>Get returns the data for the given key, or nil if the key doesn't exist in the schema. If the key does exist in the schema but doesn't exist in the configuration, then the default value for that type will be returned. For strings, this is "", for numbers it is 0, etc. If you want to test if something is set at all in the configuration, use GetOk.

* `aggregation_method` value comparison is case insensitive now

## Type of change

## Checklist:

- [x] New and existing unit tests pass locally with my changes.
## How to test this change?

A basic test case would be (see https://github.com/newrelic/terraform-provider-newrelic/issues/1900 for example):

- terraform apply `newrelic_nrql_alert_condition` without any `aggregation_*`  and `evaluation_offset` properties specified
- `terrafrom plan` => should show no diffs


newrelic/resource_newrelic_nrql_alert_condition.go
```diff
@@ -324,14 +325,15 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
...
+					_, sinceValueExists := d.GetOk("nrql.0.since_value")
 					if sinceValueExists {
 						return false
 					}
 					// If a value is not provided and the condition uses the default value, don't show a diff
 					oldInt, _ := strconv.ParseInt(old, 0, 8)
 					newInt, _ := strconv.ParseInt(new, 0, 8)
-					return oldInt == 120 && (newInt == 0) && d.Get("aggregation_method") == "EVENT_TIMER"
+					aggregationMethod := strings.ToLower(d.Get("aggregation_method").(string))
+					return oldInt == 60 && newInt == 0 && aggregationMethod == "event_timer"
```
antonblr, line 336: changed `120` to the default `60` as per [nrql_alert_condition docs](https://github.com/newrelic/terraform-provider-newrelic/blob/main/website/docs/r/nrql_alert_condition.html.markdown#argument-reference)

kidk:
Hi @antonblr 

Just to let you know we're still looking at this internally. Apologies for the delay.

baNROne (approved):
This looks great. Thanks for your contribution @antonblr !

newrelic/resource_newrelic_nrql_alert_condition.go
baNROne: 👏

PR #1916, opened 2022-08-04
feat(workflows): add workflows resources

lzaga-newrelic:
# Description
Add workflows as a new resource for New Relic provider.

## Type of change
New feature - workflows

## Checklist:

## How to test this change?

- Create a notifications destination via NerdGraph.
- Create notifications channel via NerdGraph
- Add inside testing folder `newrelic.tf` file this code and update the channel id:
```
resource "newrelic_workflow" "foo" {
  name = "workflow-example-update"
  account_id = 10867072
  enrichments_enabled = false
  destinations_enabled = true
  workflow_enabled = true
  muting_rules_handling = "NOTIFY_ALL_ISSUES"

  enrichments {
    nrql {
      name = "Log"
      configurations {
       query = "SELECT count(*) FROM Log"
      }
    }

    nrql {
      name = "Metric"
      configurations {
        query = "SELECT * FROM Metric"
      }
    }
  }

  issues_filter {
    name = "Filter1"
    type = "FILTER"

    predicates {
      attribute = "source"
      operator = "EQUAL"
      values = [ "newrelic" ]
    }
  }

  destination_configurations {
    channel_id = "7510bd1b-3f66-43d1-9731-75ea0a05886c"
  }
}
```
- Hit `terraform init` and `terraform apply`

### DIFF
--- a/newrelic/resource_newrelic_entity_tags.go
+++ b/newrelic/resource_newrelic_entity_tags.go
@@ -65,7 +65,7 @@ func resourceNewRelicEntityTags() *schema.Resource {
 			},
 		},
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(10 * time.Second),
+			Create: schema.DefaultTimeout(20 * time.Second),
 		},
 	}
 }
@@ -314,7 +314,9 @@ func tagValuesExist(t *entities.TaggingTagInput, values []string) bool {
 
 func getTag(tags []*entities.TaggingTagInput, key string) *entities.TaggingTagInput {
 	for _, t := range tags {
+		log.Printf("[INFO] Checking tag %s compared to tag %s", t.Key, key)
 		if t.Key == key {
+			log.Printf("[INFO] All good! %s = %s", t.Key, key)
 			return t
 		}
 	}
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -293,14 +293,14 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				ValidateFunc: validation.StringInSlice([]string{"CADENCE", "EVENT_FLOW", "EVENT_TIMER"}, true),
-				Description:  "The method that determines when we consider an aggregation window to be complete so that we can evaluate the signal for violations. Default is CADENCE.",
+				Description:  "The method that determines when we consider an aggregation window to be complete so that we can evaluate the signal for violations. Default is EVENT_FLOW.",
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					sinceValueExists := d.Get("nrql.0.since_value") != nil
+					_, sinceValueExists := d.GetOk("nrql.0.since_value")
 					if sinceValueExists {
 						return false
 					}
 					// If a value is not provided and the condition uses the default value, don't show a diff
-					return (old == "event_flow" && new == "") || strings.EqualFold(old, new)
+					return (strings.EqualFold(old, "event_flow") && new == "") || strings.EqualFold(old, new)
 				},
 			},
 			"aggregation_delay": {
@@ -309,14 +309,15 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 				Description:  "How long we wait for data that belongs in each aggregation window. Depending on your data, a longer delay may increase accuracy but delay notifications. Use aggregationDelay with the EVENT_FLOW and CADENCE aggregation methods.",
 				RequiredWith: []string{"aggregation_method"},
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					sinceValueExists := d.Get("nrql.0.since_value") != nil
+					_, sinceValueExists := d.GetOk("nrql.0.since_value")
 					if sinceValueExists {
 						return false
 					}
 					// If a value is not provided and the condition uses the default value, don't show a diff
 					oldInt, _ := strconv.ParseInt(old, 0, 8)
 					newInt, _ := strconv.ParseInt(new, 0, 8)
-					return oldInt == 120 && (newInt == 0) && (d.Get("aggregation_method") == "EVENT_FLOW" || d.Get("aggregation_method") == "CADENCE")
+					aggregationMethod := strings.ToLower(d.Get("aggregation_method").(string))
+					return oldInt == 120 && newInt == 0 && (aggregationMethod == "event_flow" || aggregationMethod == "cadence")
 				},
 			},
 			"aggregation_timer": {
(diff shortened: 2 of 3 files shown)


newrelic/resource_newrelic_workflow_test.go
```diff
@@ -0,0 +1,191 @@
...
+			{
+				Config: testNewRelicNotificationChannelConfigByType(rName, "WEBHOOK", "IINT", destinationID, `{
+					key = "payload"
+					value = "{\n\t\"id\": \"test\"\n}"
+					label = "Payload Template"
+				}`),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNewRelicNotificationChannelExists(resourceName, channelId),
+				),
+			},
```
lzaga-newrelic, line 49: not sure if this will work here @sanderblue  - how can I add here a channel creation?

sanderblue: Sorry for the delay. If we want to create a notification channel to use in the test, we can add it to the test configurations in the test file by adding a notification resource and referencing it in the workflow config string. 

Something like this
```hcl
resource "newrelic_notification_destination" "foo" {
  name = "example-destination"
  type = "WEBHOOK"

  properties {
    // ..
  }

  auth = {
    // ..
  }
}

resource "newrelic_notification_channel" "foo" {
  destination_id = newrelic_notification_destination.foo.destination_id   # reference the destination 
  name = "webhook-example"
  type = "WEBHOOK"
  // .. etc

  properties {
    // ..
  }
}

resource "newrelic_workflow" "foo" {
  // ..

  enrichments {
    // ..
  }

  issues_filter {
    // ..
  }

  destination_configurations {
    channel_id = newrelic_notification_channel.foo.channel_id   # reference the channel 
  }
}
```

newrelic/resource_newrelic_workflow.go
```diff
@@ -0,0 +1,371 @@
...
+		},
+		Schema: map[string]*schema.Schema{
+			// Required
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "(Required) The name of the workflow.",
+			},
+			"destination_configurations": {
+				Type:        schema.TypeList,
```
tjstansell, line 35: the order of these shouldn't matter, should it?  Why is this not `schema.TypeSet`?

newrelic/resource_newrelic_workflow.go
```diff
@@ -0,0 +1,371 @@
...
+			},
+			"muting_rules_handling": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice(listValidMutingRulesTypes(), false),
+				Description:  fmt.Sprintf("The type of the muting rule handling. One of: (%s).", strings.Join(listValidMutingRulesTypes(), ", ")),
+			},
+
+			// Optional
+			"workflow_enabled": {
```
tjstansell, line 125: This seems odd to have a `workflow_enabled` flag on a workflow resource.  I guess you're trying to clarify this is separate than the other `*_enabled` flags, but it seems redundant to me.  Having an `enable` flag on the resource itself would be more logical to me.

newrelic/resource_newrelic_workflow.go
```diff
@@ -0,0 +1,371 @@
...
+			},
+			"destinations_enabled": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Indicates whether the destinations are enabled.",
+			},
+			"enrichments": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Description: "Workflow's destination configuration.",
```
tjstansell, line 143: this description needs to get updated.

newrelic/resource_newrelic_workflow.go
```diff
@@ -0,0 +1,371 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Indicates whether the enrichments are enabled.",
+			},
+			"destinations_enabled": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Indicates whether the destinations are enabled.",
+			},
+			"enrichments": {
```
tjstansell, line 140: this is a block definition for enrichments, right?  so shouldn't this be called `enrichment` because each block represents a single enrichment?

newrelic/resource_newrelic_workflow.go
```diff
@@ -0,0 +1,371 @@
...
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			// Required
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "(Required) The name of the workflow.",
+			},
+			"destination_configurations": {
```
tjstansell, line 34: This is another case of what looks like a block using a plural.  It seems this should simply be called `destination` since it's a block definition for a single destination.
```
  destination {
    channel_id = 12345
  }
  destination {
    channel_id = 23456
  }
```

newrelic/resource_newrelic_workflow.go
```diff
@@ -0,0 +1,371 @@
...
+						},
+						"type": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validation.StringInSlice(listValidWorkflowsFilterTypes(), false),
+							Description:  fmt.Sprintf("(Required) The type of the filter. One of: (%s).", strings.Join(listValidWorkflowsFilterTypes(), ", ")),
+						},
+
+						// Optional
+						"predicates": {
```
tjstansell, line 81: another case where this is a block and so should be using singular form `predicate` since it's a block that defines a single predicate configuration.

newrelic/resource_newrelic_workflow.go
```diff
@@ -0,0 +1,371 @@
...
+							Description: "(Required) Nrql type Enrichments.",
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									// Required
+									"name": {
+										Type:        schema.TypeString,
+										Required:    true,
+										Description: "(Required) Enrichment's name.",
+									},
+									"configurations": {
```
tjstansell, line 158: another plural block that should be singular `configuration`.

newrelic/resource_newrelic_workflow.go
sanderblue: Correct, we can make this a `TypeSet`

newrelic/resource_newrelic_workflow.go
lzaga-newrelic: should we also do it to notification channels and destinations `properties`?

newrelic/resource_newrelic_workflow.go
lzaga-newrelic: ops, sorry :)

sanderblue:
Merging this pull request and will finish the fine tuning in a new branch off `main`. Thanks for the great work @lzaga-newrelic 🎉

tjstansell:
So, you ack'd a bunch of my feedback ... but apparently didn't make any changes as a result.  even the description that was wrong wasn't updated.  I would suggest the maintainers of this project take a look at https://www.terraform.io/plugin/sdkv2/best-practices/naming as most of my comments were to address the points that did not follow that guideline.

tjstansell:
I'll also note that by merging this as-is and actually releasing it, now if you wanted to fix all of these attribute names you'll need a schema migration since folks could be using the existing names.  It would have been a lot simpler to address those concerns before merging.

kidk:
Thanks for your feedback and comments @tjstansell We have indeed missed some of your feedback and we will try to implement it in the following weeks. As mentioned this is now a bit harder so might take some time. I think because of the time pressure on this release the team wasn't able to process everything, but we will take our learnings from this and do better in the future.

newrelic/resource_newrelic_workflow.go
lzaga-newrelic: Fixed in this pr: https://github.com/newrelic/terraform-provider-newrelic/pull/1990

newrelic/resource_newrelic_workflow.go
lzaga-newrelic: fixed in this pr: https://github.com/newrelic/terraform-provider-newrelic/pull/1990

newrelic/resource_newrelic_workflow.go
lzaga-newrelic: fixed in this pr: https://github.com/newrelic/terraform-provider-newrelic/pull/1990

PR #1949, opened 2022-08-26
Alert channel multi account support

### LINKED ISSUE #1948: newrelic data resource ignores providers account id when listing alert channels.
```
terraform -v
Terraform v1.1.9
on linux_amd64
```

### Affected Resource(s)
**newrelic_alert_channel** data source. ```https://github.com/newrelic/terraform-provider-newrelic/blob/main/newrelic/data_source_newrelic_alert_channel.go```

```
func dataSourceNewRelicAlertChannelRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
	client := meta.(*ProviderConfig).NewClient

	log.Printf("[INFO] Reading New Relic Alert Channels")

	channels, err := client.Alerts.ListChannelsWithContext(ctx)
	if err != nil {
		return diag.FromErr(err)
	}
```

which uses Rest Endpoint.

```
// ListChannelsWithContext returns all alert channels for a given account.
func (a *Alerts) ListChannelsWithContext(ctx context.Context) ([]*Channel, error) {
	alertChannels := []*Channel{}
	nextURL := a.config.Region().RestURL("/alerts_channels.json")

	for nextURL != "" {
		response := alertChannelsResponse{}
		resp, err := a.client.GetWithContext(ctx, nextURL, nil, &response)

		if err != nil {
			return nil, err
		}

		alertChannels = append(alertChannels, response.Channels...)

		paging := a.pager.Parse(resp)
		nextURL = paging.Next
	}

	return alertChannels, nil
}
```

Please use **graphql** endpoint so that account wise listing is possible. 

**Policy data resource** already has a good working example. ```https://github.com/newrelic/terraform-provider-newrelic/blob/main/newrelic/data_source_newrelic_alert_channel.go```

```
	policies, err := client.Alerts.QueryPolicySearchWithContext(ctx, accountID, params)
	if err != nil {
		return diag.FromErr(err)
	}

	var policy *alerts.AlertsPolicy

	for _, c := range policies {
		if strings.EqualFold(c.Name, name) {
			policy = c

			break
		}
	}
```

Please use the same for alerts channels.This alert channel bug is a big blocker in our automation process as we cant create user key in all subaccounts due to a single data resource bug.

kidk:
# Description

Fixes #1948 

## Type of change

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,6 @@ website/vendor
 !command/test-fixtures/**/*.tfstate
 !command/test-fixtures/**/.terraform/
 
-# Ignore tf files in testing, except for newrelic.tf
+# Ignore testing directory except for newrelic.tf
 testing/*.tf
 !testing/newrelic.tf
--- a/newrelic/resource_newrelic_alert_policy.go
+++ b/newrelic/resource_newrelic_alert_policy.go
@@ -66,6 +66,7 @@ func resourceNewRelicAlertPolicyCreate(ctx context.Context, d *schema.ResourceDa
 
 	client := providerConfig.NewClient
 	accountID := selectAccountID(providerConfig, d)
+	updatedContext := updateContextWithAccountID(ctx, accountID)
 
 	policy := alerts.AlertsPolicyInput{}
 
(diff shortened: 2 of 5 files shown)


kidk:
Needs some more testing before we can merge it.

PR #2082, opened 2022-11-01
chore(newrelic_nrql_alert_condition): move read to create function to avoid root resource not found

kidk:
# Description

Change to read function to avoid setId("") happening. I think we need to verify the resource has been created by reading it back before setting the Id.

## Type of change

## Checklist:

### DIFF
--- a/newrelic/resource_newrelic_alert_policy_channel.go
+++ b/newrelic/resource_newrelic_alert_policy_channel.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
@@ -53,6 +54,9 @@ func resourceNewRelicAlertPolicyChannel() *schema.Resource {
 				Upgrade: migrateStateNewRelicAlertPolicyChannelV0toV1,
 				Version: 0,
 			},
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Second),
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


newrelic/resource_newrelic_nrql_alert_condition.go
```diff
@@ -412,9 +413,25 @@ func resourceNewRelicNrqlAlertConditionCreate(ctx context.Context, d *schema.Res
 		return diag.FromErr(err)
 	}
 
-	d.SetId(serializeIDs([]int{d.Get("policy_id").(int), conditionID}))
+	retryErr := resource.RetryContext(ctx, d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
```
mbazhlekova, line 416: Should we define a value for `schema.TimeoutCreate`? I wasn't sure if there was a default value.

newrelic/resource_newrelic_alert_policy_channel.go
```diff
@@ -116,9 +118,26 @@ func resourceNewRelicAlertPolicyChannelCreate(ctx context.Context, d *schema.Res
 		return diag.FromErr(err)
 	}
 
+	retryErr := resource.RetryContext(ctx, d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
```
mbazhlekova, line 121: Should we define a value for schema.TimeoutCreate? I wasn't sure if there was a default value.

newrelic/resource_newrelic_nrql_alert_condition.go
kidk: I think it has a default, but I added it to be sure. Thanks 👍

newrelic/resource_newrelic_alert_policy_channel.go
kidk: I think it has a default, but I added it to be sure. Thanks 👍

mbazhlekova (approved):

PR #2109, opened 2022-11-17
Fix url instead of domain, on the domain field for newrelic_synthetic…

spideyz0r:
The document recommends the use of a URL instead of a domain.

This PR aims to fix the docs, the user should place a domain, not a URL.

Using the New Relic UI, having a URL is not even allowed.

-This change is updating the documentation.

No tests are required, just updating the doc.


mbazhlekova:
Hi @spideyz0r thank you for your contribution! One thing, can you amend your commit message to follow conventional commits, like `docs(synthetics_cert_check_monitor): fix url instead of domain in domain field`?

spideyz0r:
@mbazhlekova sure, see if it looks better now, please

mbazhlekova (approved):

PR #2114, opened 2022-11-21
fix(notifications channel/destination update entity): Updated provider to work with new client version

amaor-newrelic:
# Description

Currently there's no way to update the boolean field "active" to be false, as the "omitempty" tag(?) parses false as empty. The change has been done in the client and it is a breaking change, so this PR was created to work with the new version

Fixes # (issue)

## Type of change

## Checklist:

- [ ] New and existing unit tests pass locally with my changes. Go [here](https://github.com/newrelic/terraform-provider-newrelic#testing) for instructions on running tests locally.

## How to test this change?

- replace client in go mod with version from this pr
- create a channel or destination
- update the entity's active field to "false"
- entity should update and no further changes will be pending when using terraform plan


newrelic/structures_newrelic_notifications_channel.go
```diff
@@ -18,9 +18,10 @@ func expandNotificationChannel(d *schema.ResourceData) notifications.AiNotificat
 }
 
 func expandNotificationChannelUpdate(d *schema.ResourceData) notifications.AiNotificationsChannelUpdate {
+	active := d.Get("active").(bool)
```
lzaga-newrelic, line 21: This will be nil if there is no active property, right?

newrelic/structures_newrelic_notifications_channel.go
```diff
@@ -18,9 +18,10 @@ func expandNotificationChannel(d *schema.ResourceData) notifications.AiNotificat
...
 	channel := notifications.AiNotificationsChannelUpdate{
 		Name:   d.Get("name").(string),
-		Active: d.Get("active").(bool),
+		Active: &active,
```
lzaga-newrelic, line 24: Do you need to upgrade the go-client version here?

lzaga-newrelic:
Also, looks like tests are failing :)

PR #2141, opened 2022-12-05
NR-61328: verbose error message nrql drop rules :bug:

NSSPKrishna:
# Description

- Added the diagnostics from api

## Type of change

## Checklist:

### DIFF
--- a/newrelic/resource_newrelic_nrql_drop_rule.go
+++ b/newrelic/resource_newrelic_nrql_drop_rule.go
@@ -82,7 +82,7 @@ func resourceNewRelicNRQLDropRuleCreate(ctx context.Context, d *schema.ResourceD
 	if len(created.Failures) != 0 || created == nil {
 		for _, resp := range created.Failures {
 			switch resp.Error.Reason {
-			case "GENERAL":
+			case nrqldroprules.NRQLDropRulesErrorReasonTypes.GENERAL:
 				return diag.Errorf("err: drop rule create result wasn't returned. Validate the action value or NRQL query.")
 			default:
 				diags = append(diags, diag.Diagnostic{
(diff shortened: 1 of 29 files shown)


newrelic/resource_newrelic_nrql_drop_rule.go
```diff
@@ -149,11 +162,28 @@ func resourceNewRelicNRQLDropRuleDelete(ctx context.Context, d *schema.ResourceD
...
+	deleted, err := client.Nrqldroprules.NRQLDropRulesDeleteWithContext(ctx, accountID, deleteInput)
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
+	var diags diag.Diagnostics
+	if len(deleted.Failures) != 0 {
+		for _, resp := range deleted.Failures {
+			switch resp.Error.Reason {
+			case "GENERAL":
```
sanderblue, line 174: Can we create a list of enums for these constants?

e.g. 

```go
type DropRuleErrorType string 

var DropRuleErrorTypes = struct {
	GENERAL DropRuleErrorType
	RULE_NOT_FOUND DropRuleErrorType
}{
	GENERAL: "GENERAL",
	RULE_NOT_FOUND: "RULE_NOT_FOUND",
}
```

sanderblue: Do you think `INVALID_QUERY` should also be a part of this list too?

sanderblue (changes requested):
Looks good, just one minor change request.

newrelic/resource_newrelic_nrql_drop_rule.go
NSSPKrishna: It has [these](https://github.com/newrelic/newrelic-client-go/blob/main/pkg/nrqldroprules/types.go#L28) types returned out of which GENERAL is the only error that doesn't have a proper description.

sanderblue: Oh nice, then let's use the types here instead of a raw string?

Something like this?
```go
switch resp.Error.Reason {
case nrqldroprules.NRQLDropRulesErrorReasonTypes.GENERAL:
	// ...
```

NSSPKrishna: ok

sanderblue (approved):

mbazhlekova (approved):

PR #2164, opened 2022-12-16
chore: update NewRelic client version to 2.8.0

IliaShurygin:
# Description

2.8.0 version of NewRelic go client introduced some backward-incompatible changes. 
This PR fixes the broken code without changing the behaviour.

## Checklist:

## How to test this change?

Existing tests should cover everything

### DIFF
--- a/newrelic/resource_newrelic_workflow.go
+++ b/newrelic/resource_newrelic_workflow.go
@@ -512,8 +512,7 @@ func resourceNewRelicWorkflowUpdate(ctx context.Context, d *schema.ResourceData,
 	accountID := selectAccountID(providerConfig, d)
 	updatedContext := updateContextWithAccountID(ctx, accountID)
 
-	// Setting to false to fix compilation errors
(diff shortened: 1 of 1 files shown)


newrelic/resource_newrelic_workflow.go
```diff
@@ -512,8 +512,7 @@ func resourceNewRelicWorkflowUpdate(ctx context.Context, d *schema.ResourceData,
 	accountID := selectAccountID(providerConfig, d)
 	updatedContext := updateContextWithAccountID(ctx, accountID)
 
-	// Setting to false to fix compilation errors
-	workflowResponse, err := client.Workflows.AiWorkflowsUpdateWorkflowWithContext(updatedContext, accountID, false, *updateInput)
+	workflowResponse, err := client.Workflows.AiWorkflowsUpdateWorkflowWithContext(updatedContext, accountID, true, *updateInput)
```
IliaShurygin, line 515: `true`  represents the current behaviour before the client update.

`false` is the desired value, but I will create a separate PR to change it because it would change the behaviour I would thus need to change the tests and add a couple of new ones. 

I do want to mix the build fix with business changes.

mbazhlekova (approved):

PR #2174, opened 2022-12-20
feat(provider): add option to configure user agent service name

sanderblue:


newrelic/config.go
```diff
@@ -40,9 +40,13 @@ type Config struct {
 }
 
 // Client returns a new client for accessing New Relic
-func (c *Config) Client() (*nr.NewRelic, error) {
+func (c *Config) Client(serviceName string) (*nr.NewRelic, error) {
 	options := []nr.ConfigOption{}
 
+	if serviceName == "" {
```
sanderblue, line 46: Note to self: This should use the env var.

mbazhlekova (approved):

kidk (approved):

Julien4218 (approved):

jkodroff (approved):
With the caveat that I don't have _deep_ expertise on this, LPGTM. ;)

As long as we're sure the service name flag correctly ends up in the HTTP request and NR can use it to get the metrics it needs, we're good.

PR #2181, opened 2022-12-27
NR-74274,NR-74632:  added data source provider and docs for test grok

ChandraIragavarapu:


newrelic/data_source_newrelic_test_grok_pattern.go
```diff
@@ -0,0 +1,131 @@
...
+
+func dataSourceNewRelicTestGrokRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	providerConfig := meta.(*ProviderConfig)
+	client := providerConfig.NewClient
+	accountID := selectAccountID(providerConfig, d)
+
+	log.Printf("[INFO] Test a Grok pattern against a list of log lines.")
+	lines := d.Get("log_lines").(*schema.Set).List()
+
+	if len(lines) == 0 {
```
kidk, line 86: I think you can do this in the resource definition.

newrelic/data_source_newrelic_test_grok_pattern.go
```diff
@@ -0,0 +1,131 @@
...
+		perms[i] = line.(string)
+	}
+
+	res, err := client.Logconfigurations.GetTestGrokWithContext(ctx, accountID, d.Get("grok").(string),
+		perms)
+
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(fmt.Sprintf("%d", rand.Int()))
```
kidk, line 101: If you just want to convert int to string this might be better: https://stackoverflow.com/questions/10105935/how-to-convert-an-int-value-to-string-in-go Both have same result, or close to, but SprintF is less efficient.

NSSPKrishna: Thanks for the review I'll do that.

newrelic/data_source_newrelic_test_grok_pattern.go
NSSPKrishna: ✅

PR #2290, opened 2023-03-13
refactor(entity): use GetEntitySearchByQuery instead of GetEntitySearch

sanderblue:
The PR replaces the use of `GetEntitySearchWithContext` with `GetEntitySearchByQueryWithContext`. This change provides us more precision with regard to querying for entity data. 

Under the hood, this means we're using `entitySearch.query` instead of `entitySearch.queryBuilder`. The `queryBuilder` was less precise due to loose comparisons in the query it constructs. 

👎  `queryBuilder` uses a fuzzy search with `LIKE` which is less precise. 🙁 
```
name LIKE 'app-name'
```
<br>

👍 `query` gives us more control because we can control the operators we use, such as `=` for better precision. 🙂 
```
name = 'app-name'
```

Note: This refactor should _not_ cause a breaking change.

### DIFF
--- a/newrelic/resource_newrelic_synthetics_secure_credential_test.go
+++ b/newrelic/resource_newrelic_synthetics_secure_credential_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	"github.com/newrelic/newrelic-client-go/v2/pkg/entities"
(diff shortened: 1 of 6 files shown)


newrelic/resource_newrelic_synthetics_secure_credential_test.go
```diff
@@ -10,17 +10,18 @@ import (
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	"github.com/newrelic/newrelic-client-go/v2/pkg/entities"
 )
 
 func TestAccNewRelicSyntheticsSecureCredential_Basic(t *testing.T) {
 	resourceName := "newrelic_synthetics_secure_credential.foo"
-	rName := generateNameForIntegrationTestResource()
+	rName := acctest.RandString(7)
```
NSSPKrishna, line 22: wouldn't this be excluded during cleanups?

sanderblue: Yes, good call. I'll update.

NSSPKrishna (approved):

PR #2298, opened 2023-03-22
feat(service_level): add data source service_level_alert_helper

NilVentosa:
# Description

Adding the data source `service_level_alert_helper` to help users calculate the necessary fields to set up alerts on Service Levels. This data source only does calculations and fills computed fields, it does not use the client. The goal was to make it easier for users to calculate the `threshold` field of an alert resource so they don't have to go to the documentation, find the correct formula and substitute all the variables, when wanting to set an alert on burning error budget of a Service Level.

## Type of change

## Checklist:

## How to test this change?

- Follow the steps in the provided documentation to use this new data source to help with the creation of an alert on a Service Level.


newrelic/data_source_newrelic_service_level_alert_helper.go
```diff
@@ -0,0 +1,156 @@
...
+	d.SetId(sliGUID + rnd)
+
+	var sloPeriod = d.Get(SloPeriodField).(int)
+	var sloTarget = d.Get(SloTargetField).(float64)
+	var alertType = d.Get(AlertTypeField).(string)
+
+	_, tOk := d.GetOk(CustomConsumptionField)
+	_, eOk := d.GetOk(CustomEvaluationPeriodField)
+
+	var err error
```
jdvr, line 97: nit: You don't need this err var as you are returning a on each err branch

newrelic/data_source_newrelic_service_level_alert_helper.go
```diff
@@ -0,0 +1,156 @@
...
+	CustomEvaluationPeriodField = "custom_evaluation_period"
+	ConsumptionField            = "tolerated_budget_consumption"
+	EvaluationPeriodField       = "evaluation_period"
+	ThresholdField              = "threshold"
+	NRQLField                   = "nrql"
+
+	Custom              = "custom"
+	FastBurn            = "fast_burn"
+	FastBurnPeriod      = 60
+	FastBurnConsumption = 2
```
jdvr, line 28: If I am not wrong, all this conts are file private, you can use lowercase

website/docs/d/service_level_alert_helper.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+```
+Then use the helper to obtain the necessary fields to set up an alert on that Service Level.
+
+```hcl
+
+data "newrelic_service_level_alert_helper" "foo_custom" {
+    alert_type = "custom"
+    sli_guid = "MXxBUE18QVBQTElDQVRJT058MQ"
```
carl0FF, line 57: `MXxBUE18QVBQTElDQVRJT058MQ` is the GUID of the service/application over which we want to track the SLI, as in (see above)

```
resource "newrelic_service_level" "foo" {
    guid = "MXxBUE18QVBQTElDQVRJT058MQ"
    name = "Latency"
    ...
```
and, if I'm understanding this right, the value of `sli_guid` should be another value.

So, in order to avoid inducing the reader to error,  I'd either use explicitly another GUID value , or even maybe use `newrelic_service_level.sli_guid` if it makes sense for the example purpose?

website/docs/d/service_level_alert_helper.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+  slide_by = 60
+}
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+  * `alert_type` - (Required) The type of alert we want to set. Valid values are `custom` and `fast_burn`.
```
carl0FF, line 95: Out of curiosity, why isn't `slow_burn` available?

carl0FF: Additionally, I'd mention somewhere what `fast_burn` means in terms of thresholds and so on.

NilVentosa: slow_burn is not yet available due to some limitations, it will be available in the future

carl0FF:
Hey @NilVentosa, I have a question (and please bear with me because I just know terraform superficially, and I’ve never implemented a provider):  why haven’t you guys attempted a similar approach as [synthetics](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/synthetics_alert_condition)? I mean something like this:
```terraform
resource "newrelic_service_level" "foo" {
    guid = "MXxBUE18QVBQTElDQVRJT058MQ"
    name = "Latency"
    description = "Proportion of requests that are served faster than a threshold."
    events {
        account_id = 12345678
        valid_events {
            from = "Transaction"
            where = "appName = 'Example application' AND (transactionType='Web')"
        }
        good_events {
            from = "Transaction"
            where = "appName = 'Example application' AND (transactionType= 'Web') AND duration < 0.1"
        }
    }
    objective {
        target = local.foo_target
        time_window {
            rolling {
                count = local.foo_period
                unit = "DAY"
            }
        }
    }
}

resource "newrelic_service_level_alert_condition" "foo" {
  policy_id = newrelic_alert_policy.foo.id

  name = "foo"
  sli_id = newrelic_service_level.foo.id
  type = "fast_burn"
  runbook_url = "https://www.example.com/"
}

resource "newrelic_service_level_alert_condition" "bar" {
  policy_id = newrelic_alert_policy.foo.id

  name = "bar"
  sli_id = newrelic_service_level.foo.id
  type = "custom"
  slo_target = <custom_value>
  slo_period = <custom_value>
  runbook_url = "https://www.example.com/"
}
```
so that `newrelic_service_level_alert_condition` internally calls or reuses [newrelic_nrql_alert_condition](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/alert_condition).

This would be a more compact definition, and would be consistent with other non-generic alert resources, such as those from [infra](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/infra_alert_condition) or [synthetics](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/synthetics_multilocation_alert_condition).

NilVentosa:
Hi @carl0FF bear in mind I am also not a terraform or terraform provider wizard. I think for a couple of reasons we don't want to do what synthetics and infra did. 
* In the first place it looks like they recommend not using their custom alert resources and using the generic [newrelic_nrql_alert_condition](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/nrql_alert_condition).
* Second, there are a lot more fields in the alert creation that we want to leave to our users.
The aim of this data source is not to create the alerts, but to help with a couple of fields that are not obvious, like the threshold or the nrql.
Hope this answers your question satisfactorily.

carl0FF:
> It looks like they recommend not using their custom alert resources and using the generic [newrelic_nrql_alert_condition](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/nrql_alert_condition).

Good catch, I overlooked that remark.

> There are a lot more fields in the alert creation that we want to leave to our users.

Yep, it's not explicit in my proposal, but the idea would be that `newrelic_service_level_alert_condition` would expose all of those that aren't coupled to the SLI alert per se, and simply forward them to `newrelic_nrql_alert_condition` (in a similar fashion as the decorator pattern).

newrelic/data_source_newrelic_service_level_alert_helper_test.go
```diff
@@ -0,0 +1,235 @@
...
+    custom_tolerated_budget_consumption = 34
+}
+`)
+}
+
+func testAccCheckNewRelicServiceLevelAlertHelper_FastBurn(n string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+
+		rs, ok := s.RootModule().Resources[n]
+		// fmt.Println(s.RootModule())
```
asanuy, line 146: I guess this println was used for debugging, we can safely remove it.

newrelic/data_source_newrelic_service_level_alert_helper_test.go
```diff
@@ -0,0 +1,235 @@
...
+    custom_evaluation_period = 120
+}
+`)
+}
+
+func testAccCheckNewRelicServiceLevelAlertHelper_Custom(n string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+
+		rs, ok := s.RootModule().Resources[n]
+		// fmt.Println(s.RootModule())
```
asanuy, line 196: I guess this println was used for debugging, we can safely remove it.

NilVentosa:
> > Seems to me that then we would have to modify our resource every time alerts adds/removes/modifies one of their fields.
> 
> True, but (1) it doesn't [seem to happen that often](https://github.com/newrelic/terraform-provider-newrelic/commits/main/newrelic/resource_newrelic_alert_condition.go), and (2) automated tests would help preventing such changes from going unnoticed.

We've consulted with Alerts and they prefer the data source approach over us creating the alert.
Thanks for the comment it was a good idea that didn't occur to us.

newrelic/data_source_newrelic_service_level_alert_helper.go
```diff
@@ -0,0 +1,154 @@
...
+	customEvaluationPeriodField = "custom_evaluation_period"
+	consumptionField            = "tolerated_budget_consumption"
+	evaluationPeriodField       = "evaluation_period"
+	thresholdField              = "threshold"
+	nrqlField                   = "nrql"
+
+	custom              = "custom"
+	fastBurn            = "fast_burn"
+	fastBurnPeriod      = 60
+	fastBurnConsumption = 2
```
sanderblue, line 28: Some comment regarding naming collisions. These seems like variables that could benefit from typing with an enum style variable.

```go
type serviceLevelAlertType string

var serviceLevelAlertTypes = struct {
	CUSTOM    serviceLevelAlertType
	FAST_BURN serviceLevelAlertType
}{
	CUSTOM:    "custom",
	FAST_BURN: "fast_burn",
}
```

```
// ...
"alert_type": {
	Required:     true,
	ValidateFunc: validation.StringInSlice([]string{
		string(serviceLevelAlertTypes.CUSTOM), 
		string(serviceLevelAlertTypes.FAST_BURN),
	}, true),
},

// ...
switch serviceLevelAlertType(alertType) {
case serviceLevelAlertTypes.CUSTOM:
  // ...
case serviceLevelAlertTypes.FAST_BURN:
  // ...
}
```

Thoughts on this idea?

newrelic/data_source_newrelic_service_level_alert_helper.go
```diff
@@ -0,0 +1,154 @@
...
+
+	err := d.Set(nrqlField, "FROM Metric SELECT 100 - clamp_max(sum(newrelic.sli.good) / sum(newrelic.sli.valid) * 100, 100) as 'SLO compliance'  WHERE sli.guid = '"+sliGUID+"'")
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func calculateThreshold(sloTarget float64, toleratedBudgetConsumption float64, sloPeriod int, evaluationPeriod int) float64 {
```
sanderblue, line 152: Would be great to have a unit test for this 🙂

newrelic/data_source_newrelic_service_level_alert_helper_test.go
```diff
@@ -0,0 +1,233 @@
...
+				Config: testAccNewRelicServiceLevelAlertHelperCustomConfig(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNewRelicServiceLevelAlertHelper_Custom(resourceName),
+				),
+			},
+		},
+	})
+}
+
+func TestAccNewRelicServiceLevelAlertHelper_FastBurnError(t *testing.T) {
```
sanderblue, line 47: This is great, really like the inclusion of negative tests like this. 👏 😎

newrelic/provider.go
```diff
@@ -132,6 +132,7 @@ func Provider() *schema.Provider {
 			"newrelic_synthetics_private_location":  dataSourceNewRelicSyntheticsPrivateLocation(),
 			"newrelic_synthetics_secure_credential": dataSourceNewRelicSyntheticsSecureCredential(),
 			"newrelic_test_grok_pattern":            dataSourceNewRelicTestGrokPattern(),
+			"newrelic_service_level_alert_helper":   dataSourceNewRelicServiceLevelAlertHelper(),
```
sanderblue, line 135: Thoughts on naming the data source `newrelic_service_level_alert_query`?

sanderblue (changes requested):
Nice work! Very clever usage of a Terraform data source! This will definitely help customers. 👏 

I left a couple minor comments regarding the `const` declarations based on the repositories patterns as well as some open comments/questions to get your thoughts. Thanks for an awesome contribution! 👍

sanderblue (approved):
Looks great! Thanks 👍

PR #2330, opened 2023-04-18
feat(synthetics): add device emulation options to `newrelic_synthetics_script_browser_monitor` resource

sanderblue:


newrelic/resource_newrelic_synthetics_script_monitor.go
```diff
@@ -73,6 +73,16 @@ func syntheticsScriptBrowserMonitorAdvancedOptionsSchema() map[string]*schema.Sc
 			Optional:    true,
 			Description: "Capture a screenshot during job execution.",
 		},
+		"device_orientation": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "The device orientation the user would like to represent. Valid values are LANDSCAPE, PORTRAIT, or NONE.",
```
NSSPKrishna, line 79: Any thoughts on having ValidateFunc here?

pranav-new-relic (approved):

newrelic/structures_newrelic_synthetics_script_api_monitor.go
```diff
@@ -22,15 +22,25 @@ func buildSyntheticsScriptAPIMonitorInput(d *schema.ResourceData) synthetics.Syn
...
-	if v, ok := d.GetOk("runtime_type_version"); ok {
-		input.Runtime.RuntimeTypeVersion = synthetics.SemVer(v.(string))
+
+	_, scriptLangOk := d.GetOk("script_language")
+	_, runtimeTypeOk := d.GetOk("runtime_type")
+	_, runtimeVersionOk := d.GetOk("runtime_type_version")
+
+	if scriptLangOk || runtimeTypeOk || runtimeVersionOk {
+		input.Runtime = &synthetics.SyntheticsRuntimeInput{}
+
```
NSSPKrishna, line 32: The API is expecting to send all the 3 values so may be changing to `and` would be better.

newrelic/structures_newrelic_synthetics_script_api_monitor.go
```diff
@@ -51,14 +61,24 @@ func buildSyntheticsScriptAPIMonitorUpdateInput(d *schema.ResourceData) syntheti
...
-		input.Runtime.RuntimeType = v.(string)
-	}
-	if v, ok := d.GetOk("runtime_type_version"); ok {
-		input.Runtime.RuntimeTypeVersion = synthetics.SemVer(v.(string))
+
+	_, scriptLangOk := d.GetOk("script_language")
+	_, runtimeTypeOk := d.GetOk("runtime_type")
+	_, runtimeVersionOk := d.GetOk("runtime_type_version")
+
+	if scriptLangOk || runtimeTypeOk || runtimeVersionOk {
```
NSSPKrishna, line 69: Any thoughts on changing to `and` as the API is expecting to have all the 3 values?

NSSPKrishna (changes requested):

newrelic/resource_newrelic_synthetics_script_monitor.go
sanderblue: I'd rather let the API handle the error scenario. I will add a test for that as well.

newrelic/structures_newrelic_synthetics_script_api_monitor.go
sanderblue: Due to `runtime` being optional at the top level with `runtime.scriptLanguage` being optional and `runtime.runtimeType` and `runtime.runtimeTypVersion` being required, it's a little tricky. First we check to make sure at least one of the values is set, and if it is set, then we can set the `runtime` object. I think this might fix that other weird synthetics issues we were running into a little while back. I'll keep working on this though and add appropriate tests.

newrelic/structures_newrelic_synthetics_script_api_monitor.go
sanderblue: Commented above :)

PR #2395, opened 2023-06-11
feat(data_source_newrelic_notifications_destination): Support name in destination data source

lzaga-newrelic:
# Description

Add support for creating destination data sources by name. Jira ticket link [here](https://issues.newrelic.com/browse/NR-88092).

## Type of change

## Checklist:

## How to test this change?

- Create a destination with a specific name via New Relic notifications destination UI - in the Alerts tab
- Add this code to your terraform file with the name of the destination:
```
data "newrelic_notification_destination" "foo" {
  name = <destination_name>
}
```

### DIFF
--- a/newrelic/data_source_newrelic_notifications_destination.go
+++ b/newrelic/data_source_newrelic_notifications_destination.go
@@ -20,19 +20,19 @@ func dataSourceNewRelicNotificationDestination() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"id": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				Description: "The ID of the destination.",
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The name of the destination.",
 			},
 			"account_id": {
 				Type:        schema.TypeInt,
 				Optional:    true,
 				Computed:    true,
 				Description: "The account ID under which to put the destination.",
-			},
-			"name": {
-				Type:        schema.TypeString,
-				Computed:    true,
-				Description: "The name of the destination.",
 			},
 			"type": {
 				Type:        schema.TypeString,
@@ -67,9 +67,22 @@ func dataSourceNewRelicNotificationDestinationRead(ctx context.Context, d *schem
 	providerConfig := meta.(*ProviderConfig)
 	accountID := selectAccountID(providerConfig, d)
 	updatedContext := updateContextWithAccountID(ctx, accountID)
-	id := d.Get("id").(string)
-	filters := ai.AiNotificationsDestinationFilter{ID: id}
+	var filters ai.AiNotificationsDestinationFilter
 	sorter := notifications.AiNotificationsDestinationSorter{}
+
+	nameValue, nameOk := d.Get("name").(string)
+	if nameOk {
+		filters = ai.AiNotificationsDestinationFilter{Name: nameValue}
+	}
+
+	idValue, idOk := d.Get("id").(string)
+	if idOk {
+		filters = ai.AiNotificationsDestinationFilter{ID: idValue} // will override name parameter
+	}
+
+	if !nameOk && !idOk {
+		return diag.FromErr(fmt.Errorf("missing 'id' or 'name' parameters"))
+	}
 
 	destinationResponse, err := client.Notifications.GetDestinationsWithContext(updatedContext, accountID, "", filters, sorter)
 	if err != nil {
@@ -83,7 +96,7 @@ func dataSourceNewRelicNotificationDestinationRead(ctx context.Context, d *schem
 
 	if len(destinationResponse.Entities) == 0 {
 		d.SetId("")
-		return diag.FromErr(fmt.Errorf("the id '%s' does not match any New Relic notification destination", id))
+		return diag.FromErr(fmt.Errorf("the id or name you provided does not match any New Relic notification destination"))
 	}
 
 	errors := buildAiNotificationsResponseErrors(destinationResponse.Errors)
--- a/newrelic/data_source_newrelic_notifications_destination.go
+++ b/newrelic/data_source_newrelic_notifications_destination.go
@@ -92,11 +92,11 @@ func dataSourceNewRelicNotificationDestinationRead(ctx context.Context, d *schem
 	}
 
 	if len(destinationResponse.Entities) == 0 {
-		if !idOk || (idValue != "" && nameValue == "") {
-			d.SetId("")
+		d.SetId("")
+		if idValue != "" && nameValue == "" {
 			return diag.FromErr(fmt.Errorf("the id provided does not match any New Relic notification destination"))
 		}
-		if !nameOk || (nameValue != "" && idValue == "") {
+		if nameValue != "" && idValue == "" {
 			return diag.FromErr(fmt.Errorf("the name provided does not match any New Relic notification destination"))
 		}
 	}
--- a/newrelic/data_source_newrelic_notifications_destination.go
+++ b/newrelic/data_source_newrelic_notifications_destination.go
@@ -92,11 +92,11 @@ func dataSourceNewRelicNotificationDestinationRead(ctx context.Context, d *schem
 	}
 
 	if len(destinationResponse.Entities) == 0 {
-		d.SetId("")
-		if !idOk || idValue == "" {
+		if !idOk || (idValue != "" && nameValue == "") {
+			d.SetId("")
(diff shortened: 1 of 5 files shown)


lzaga-newrelic:
Waiting for this PR: https://github.com/newrelic/newrelic-client-go/pull/1014

newrelic/data_source_newrelic_notifications_destination.go
```diff
@@ -67,10 +69,18 @@ func dataSourceNewRelicNotificationDestinationRead(ctx context.Context, d *schem
...
 
+	nameValue, nameOk := d.Get("name").(string)
+	if nameOk && nameValue != "" {
+		filters = ai.AiNotificationsDestinationFilter{Name: nameValue}
+	}
+	idValue, idOk := d.Get("id").(string)
+	if idOk && idValue != "" {
+		filters = ai.AiNotificationsDestinationFilter{ID: idValue}
+	}
+
```
pranav-new-relic, line 83: @lzaga-newrelic small question - the ID is added to the filter only if the name is not found and has not been added to the filter. Does this mean the name takes precedence over the ID?

pranav-new-relic: Never mind, sorry - I realised only one of **id** and **name** can be specified and upon testing this, found that the condition has been enforced and throws an error if both **id** and **name** are specified.

website/docs/d/notification_destination.html.markdown
```diff
@@ -33,18 +33,44 @@ resource "newrelic_notification_channel" "foo-channel" {
...
+  destination_id = data.newrelic_notification_destination.foo.id
+  product = "IINT"
+
+  property {
+    key = "payload"
+    value = "{\n\t\"name\": \"foo\"\n}"
+    label = "Payload Template"
+  }
+}
+```
```
pranav-new-relic, line 57: A small suggestion - can you please run `terraform fmt` on this code snippet and paste it here, just so it is well formatted? We're trying to do this whenever we make new changes to docs, so just wanted to suggest :)

website/docs/d/notification_destination.html.markdown
```diff
@@ -33,18 +33,44 @@ resource "newrelic_notification_channel" "foo-channel" {
...
+
 ## Argument Reference
 
 The following arguments are supported:
 
+Required **one of these two** arguments:
 * `id` - (Required) The id of the notification destination in New Relic.
+* `name` - (Required) The name of the notification destination.
```
pranav-new-relic, line 65: Just a thought - I think we can have (Optional) against **id** and **name** here, with the note at the top being "Either of the following two attributes are required, and not both:"

newrelic/data_source_newrelic_notifications_destination_test.go
```diff
@@ -52,6 +73,25 @@ data "newrelic_notification_destination" "foo" {
...
+
+	property {
+		key = "url"
+		value = "https://webhook.site/"
+	}
+}
+
+data "newrelic_notification_destination" "foo" {
+	name = newrelic_notification_destination.foo.name
+}
```
pranav-new-relic, line 91: Similar to [this comment](https://github.com/newrelic/terraform-provider-newrelic/pull/2395/files#r1230053944) - if possible, can you please format this via `terraform fmt`

newrelic/data_source_newrelic_notifications_destination.go
```diff
@@ -83,7 +93,7 @@ func dataSourceNewRelicNotificationDestinationRead(ctx context.Context, d *schem
 
 	if len(destinationResponse.Entities) == 0 {
 		d.SetId("")
-		return diag.FromErr(fmt.Errorf("the id '%s' does not match any New Relic notification destination", id))
+		return diag.FromErr(fmt.Errorf("the id or name you provided does not match any New Relic notification destination"))
```
pranav-new-relic, line 96: Another suggestion - do you think it would be better if we can conditionally vary the error message printed based on the ID or Name? Something similar to 

```go
		if !idOk || idValue == "" {
			return diag.FromErr(fmt.Errorf("the id provided does not match any New Relic notification destination"))
		}
		if !nameOk || nameValue == "" {
			return diag.FromErr(fmt.Errorf("the name provided does not match any New Relic notification destination"))
		}
```

(just a sample, doesn't necessarily have to be the right solution)
Also, discarding **you** from the message would be ideal.

pranav-new-relic (approved):

website/docs/d/notification_destination.html.markdown
lzaga-newrelic: Hi @pranav-new-relic , it doesn't look like this command affects this file. do you have another suggestion?

newrelic/data_source_newrelic_notifications_destination_test.go
lzaga-newrelic: also here - file wasn't affected by this command

newrelic/data_source_newrelic_notifications_destination.go
lzaga-newrelic: done

website/docs/d/notification_destination.html.markdown
pranav-new-relic: Hey @lzaga-newrelic - my bad. I should have been a little more clear - I think there's no straightforward procedure to run this on the file, apart from copying this block to a Terraform file and running `terraform fmt` on it :) I just pushed a commit with the formatting changes 👍

newrelic/data_source_newrelic_notifications_destination_test.go
pranav-new-relic: done: https://github.com/newrelic/terraform-provider-newrelic/pull/2395#discussion_r1230581080

newrelic/data_source_newrelic_notifications_destination.go
pranav-new-relic: A small question again, @lzaga-newrelic - have you been able to test this after you made the change needed to address this comment? I just want to be sure the condition we discussed here (and you added to the code) is actually right and we're not missing anything

website/docs/d/notification_destination.html.markdown
lzaga-newrelic: thanks :)

newrelic/data_source_newrelic_notifications_destination.go
lzaga-newrelic: just a small fix needed here - will push it right away

pranav-new-relic: @lzaga-newrelic maybe the new condition you've added needs to be the other way round 🤔 

`if !idOk || (idValue == "" && nameValue != "")`

and similarly for the name

(unless I'm missing something - please let me know)

lzaga-newrelic: nope, because there is an id for example but incorrect so 
`if !idOk || (idValue != "" && nameValue == "")` will be the right way to go

pranav-new-relic: Oh yeah, got you 👍

website/docs/d/notification_destination.html.markdown
```diff
@@ -37,14 +60,17 @@ resource "newrelic_notification_channel" "foo-channel" {
...
+* `name` - (Optional) The name of the notification destination.
+
+Optional:
 * `account_id` - (Optional) The New Relic account ID to operate on.  This allows you to override the `account_id` attribute set on the provider. Defaults to the environment variable `NEW_RELIC_ACCOUNT_ID`.
 
 ## Attributes Reference
 
 In addition to all arguments above, the following attributes are exported:
 
-* `name` - The name of the notification destination.
```
pranav-new-relic, line 47: One final thing I just found - since `name` is no longer a computed attribute (like it previously was), I think I understand `name` would be discarded from the list of exported attributes (we expect `name` to be supplied to the configuration of this data source and not be exported). However, I've a few thoughts 

1. If `name` is removed from the list of exported attributes (as per this documentation change), it needs to be removed from `flattenNotificationDestinationDataSource()` in `structures_newrelic_notification_destinations.go` as the `name` attribute is still accessible from the data source (I just tested it with an `output` block where I am able to print the `name` of the destination)

2. Else, we might want to revert this documentation change to have `name` remain as an attribute that may be exported

I am thinking of (2) over (1), since we currently could have users who could be using this data source to obtain the name, and removing it from the list of exported attributes could break their existing configuration. What do you think 🤔

lzaga-newrelic: I agree, I'll add it back to the list :)

Issue #2393, opened 2023-06-08
Command line init does not locate resource


Steve24by7:
Running terraform init for simple use case on MAC Ventura 
NRmac:azure$terraform version
Terraform v1.4.6
on darwin_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.15.0
+ provider registry.terraform.io/hashicorp/azurerm v3.0.0
+ provider registry.terraform.io/newrelic/newrelic v3.24.0

 for [this](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/guides/cloud_integrations_guide#azure) provider:
NRmac:azure$terraform init
Initializing modules...
Downloading git::https://github.com/newrelic/terraform-provider-newrelic.git for newrelic-azure-cloud-integrations...
╷
│ Error: subdir "tree/main/examples/modules/cloud-integrations/azure" not found
│ 
│ 
╵

My main
terraform {
  required_version = "~> 1.0"
  required_providers {
    newrelic = {
      source  = "newrelic/newrelic"
    }
    azurerm = {
      source = "hashicorp/azurerm"
    }
  }
}

provider "newrelic" {
  account_id = 123456
  api_key = "NRAK-xxxxx"
  region = "US"
}

provider "azurerm" {
  features {}

}

module "newrelic-azure-cloud-integrations" {
    source = "github.com/newrelic/terraform-provider-newrelic/tree/main/examples/modules/cloud-integrations/azure"

  newrelic_account_id     = 123456
  name                    = "Steve24by7Azure"
}

however if instead the source line reads as this:
  source = "github.com/newrelic/terraform-provider-newrelic/examples/modules/cloud-integrations/azure"

It works properly.
Clearly during the init, the downloaded files from git are placed in a directory tree that does not have the "tree/main" directories in the path.
NRmac:azure$ls  .terraform/modules/newrelic-azure-cloud-integrations/
CHANGELOG.md				build					scripts
CODE_OF_CONDUCT.md			examples				terraform-registry-manifest.json
CONTRIBUTING.md				go.mod					testing
GNUmakefile				go.sum					tools
LICENSE					main.go					website
README.md				newrelic

pranav-new-relic:
[...]

PR #2396, opened 2023-06-12
docs(guides): updates to getting started guide to remove deprecated resources

pranav-new-relic:


website/docs/guides/cloud_integrations_guide.html.markdown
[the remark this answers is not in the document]
pranav-new-relic: Hey @Steve24by7, thank you for reaching out here, on the PR 😄 Yes, I am aware the two other code snippets need this change too - but discarded them (temporarily) for a git conflict I had to fix, which is why this PR is still a draft :) The two other changes will be part of this PR soon. Thank you for your inputs :)

website/docs/guides/getting_started.html.markdown
```diff
@@ -136,76 +136,51 @@ New Relic alerts are great, but they're even better when combined with good noti
...
-This example will send an email to the specified recipients whenever the associated alert condition is triggered. If you would like to send notifications via different modalities, such as Slack, the [alert channel](/providers/newrelic/newrelic/latest/docs/resources/alert_channel) resource supports multiple types of channels. Refer to the [additional alert channel examples](/providers/newrelic/newrelic/latest/docs/resources/alert_channel#type) for more examples in configuring your different notification channels.
-
-## A Note About Secrets
-
-As part of a `newrelic` resource, there is often some amount of configuration
-that is required in order for a resource to reach its full potential.  In some
-cases, once a given entity is created, API results will obscure the values for
-items that are deemed to be secret.  As a result, Terraform is unable to make
-an accurate detection of a resource state, and so marks a resource as changed
-for every run.
```
pranav-new-relic, line 167: Note for the reviewer: The changes discard this section, because
- This section briefly states that sensitive credentials are sometimes not returned by the API, so running `terraform plan` after `terraform apply` (after the resource is created) might result in a drift for a few credentials, so such attributes need to be added to `ignore_changes`. However, this no longer seems to be the case. This was the case when Terraform first introduced "sensitive credentials" and this discrepancy was seen, but no longer exists now. I tried running the following example and others in the provider which contain sensitive fields, but found a drift in none.
- Upon discussing all of the above with Sander, it was inferred that this section in the docs is outdated, owing to the above.

website/docs/guides/getting_started.html.markdown
```diff
@@ -136,76 +136,51 @@ New Relic alerts are great, but they're even better when combined with good noti
...
-  config {
-    recipients              = "username@example.com"
-    include_json_attachment = "1"
+resource "newrelic_notification_destination" "sample_notification_destination" {
+  name = "Sample Notification Destination"
+  type = "EMAIL"
+  
+  property {
+    key   = "email"
+    value = "username@example.com"
```
pranav-new-relic, line 145: A note for @kidk: Based on your feedback, I did want to remove the `property` block, but it looks like both `notification_destination` and `notification_channel` need to have atleast one `property` block, failing which I see an error ` Error: Insufficient property blocks: At least 1 "property" blocks are required.`.

This is why the `property` block has been added in the new example.

website/docs/guides/getting_started.html.markdown
```diff
@@ -136,76 +136,51 @@ New Relic alerts are great, but they're even better when combined with good noti
...
+    type = "FILTER"
+    predicate {
+      attribute = "labels.policyIds"
+      operator  = "EXACTLY_MATCHES"
+      values    = [newrelic_alert_policy.alert_policy_name.id]
+    }
+  }
+  
+  destination {
+    channel_id = newrelic_notification_channel.sample_notification_channel.id
```
pranav-new-relic, line 176: A note for @kidk - based on your feedback, I did want to remove `issues_filter`, but I think this is necessary to link the workflow to the alert policy, to keep it similar to the older example in which an `alert_channel` is linked to the alert policy previously created.

Please let me know your thoughts

website/docs/guides/cloud_integrations_guide.html.markdown
```diff
@@ -28,7 +28,7 @@ Add the following module to your Terraform code, and set the variables to your d
 
 ```
 module "newrelic-aws-cloud-integrations" {
-  source = "github.com/newrelic/terraform-provider-newrelic/tree/main/examples/modules/cloud-integrations/aws"
+  source = "github.com/newrelic/terraform-provider-newrelic/examples/modules/cloud-integrations/aws"
```
pranav-new-relic, line 31: A note for @kidk: Discarding `/tree/main` from these code snippets as I've been able to identify the link to import modules works without `/tree/main`. This was also identified via [this GitHub issue](https://github.com/newrelic/terraform-provider-newrelic/issues/2393).

kidk (approved):

PR #2402, opened 2023-06-22
refactor(alerts): Amend validation for violation_close_timer

### LINKED ISSUE #2398: newrelic_alert_condition argument violation_close_timer rendering changes every time
- [X] Your New Relic `provider` [configuration](#terraform-configuration-files) (sensitive details redacted)
- [X] A list of [affected resources](#affected-resources) and/or data sources
- [X] The [configuration](#terraform-configuration-files) of the resources and/or data sources related to the bug report (i.e. from the list mentioned above)
- [X] Description of the [current behavior](#actual-behavior) (the bug)
- [X] Description of the [expected behavior](#expected-behavior)

### Terraform Version
Terraform v1.1.7

### Affected Resource(s)
Please list the resources as a list, for example:
- `newrelic_alert_condition`

```hcl
provider "newrelic" {
  region     = "US"
  api_key    = var.nr_api_key
  account_id = var.nr_account_id
}

resource "newrelic_alert_condition" "condition" {
  policy_id       = var.policy_id
  name            = var.alert_condition_name
  type            = var.alert_type
  entities        = var.entities
  metric          = var.metric
  runbook_url     = var.runbook_url
  condition_scope = var.condition_scope
  gc_metric       = var.gc_metric
  
  dynamic "term" {
    for_each = var.alert_content
    content {
      duration      = term.value["duration"]
      operator      = term.value["operator"]
      priority      = term.value["priority"]
      threshold     = term.value["threshold"]
      time_function = term.value["time_function"]
    }
  }
}
```

### Actual Behavior
When omitting optional argument `violation_close_timer`, the value in created resources is `72`. This means every time terraform plan/apply runs, it reverts this to a null value:

image

If you attempt to provide value `72` to match the current resource state, there is an exception that `72` is not a valid value which matches the documentation:

`Error: expected violation_close_timer to be one of [1 2 4 8 12 24], got 72`

### Expected Behavior
The value should be set to null in the remote resources or you should be able to provide the same value of `72` to prevent persistent differences in plan/apply.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform plan` or `terraform apply` when managing a `newrelic_alert_condition` with no value set for `violation_close_timer`, after it is created

founddrama:
# Description

Updates behavior of `violation_close_timer` field for `newrelic_alert_condition` to reflect the actual behavior. Validation looks for a number of hours between 1 and 720, and defaults to 72. This makes it consistent with the underlying API.

Fixes #2398 

## Type of change

## Checklist:

## How to test this change?

- Creating an `newrelic_alert_condition` with type of `apm_jvm_metric` _or_ type of `apm_app_metric` with `condition_scope` set to `instance`
- Can create the condition with `violation_close_timer` set to any integer between 1 and 720
- Can create the condition _without_ `violation_close_timer` being set, and should default to 72

### DIFF
--- a/newrelic/resource_newrelic_alert_condition.go
+++ b/newrelic/resource_newrelic_alert_condition.go
@@ -136,8 +136,8 @@ func resourceNewRelicAlertCondition() *schema.Resource {
 			"violation_close_timer": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				ValidateFunc: intInSlice([]int{1, 2, 4, 8, 12, 24}),
-				Description:  "Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be: 1, 2, 4, 8, 12 or 24.",
+				ValidateFunc: validation.IntBetween(1, 720),
+				Description:  "Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be between 1 and 720 hours.",
 			},
 			"gc_metric": {
 				Type:        schema.TypeString,
--- a/newrelic/resource_newrelic_alert_condition.go
+++ b/newrelic/resource_newrelic_alert_condition.go
@@ -137,7 +137,8 @@ func resourceNewRelicAlertCondition() *schema.Resource {
 				Type:         schema.TypeInt,
 				Optional:     true,
 				ValidateFunc: validation.IntBetween(1, 720),
-				Description:  "Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be between 1 and 720 hours.",
+				Default:      72,
+				Description:  "Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be between 1 and 720 hours. Defaults to 72.",
 			},
 			"gc_metric": {
 				Type:        schema.TypeString,
--- a/website/docs/r/alert_condition.html.markdown
+++ b/website/docs/r/alert_condition.html.markdown
@@ -99,7 +99,7 @@ The following arguments are supported:
   * `condition_scope` - (Required for some types) `application` or `instance`.  Choose `application` for most scenarios.  If you are using the JVM plugin in New Relic, the `instance` setting allows your condition to trigger [for specific app instances](https://docs.newrelic.com/docs/alerts/new-relic-alerts/defining-conditions/scope-alert-thresholds-specific-instances).
   * `enabled` - (Optional) Whether the condition is enabled or not. Defaults to true.
   * `gc_metric` - (Optional) A valid Garbage Collection metric e.g. `GC/G1 Young Generation`.
-  * `violation_close_timer` - (Optional) Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be: `1`, `2`, `4`, `8`, `12` or `24`.
+  * `violation_close_timer` - (Optional) Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be between 1 and 720 hours.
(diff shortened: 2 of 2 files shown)


newrelic/resource_newrelic_alert_condition.go
```diff
@@ -136,8 +136,9 @@ func resourceNewRelicAlertCondition() *schema.Resource {
 			"violation_close_timer": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				ValidateFunc: intInSlice([]int{1, 2, 4, 8, 12, 24}),
-				Description:  "Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be: 1, 2, 4, 8, 12 or 24.",
+				ValidateFunc: validation.IntBetween(1, 720),
+				Default:      72,
+				Description:  "Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be between 1 and 720 hours. Defaults to 72.",
```
pranav-new-relic, line 141: @founddrama a small question - does the API now expect any value between 1 and 720 and not distinct values such as 1, 2, 4, 8, 12, 24 (as it previously did - or am I mistaken in thinking this was only a validation done in Terraform and not in the API)?

I just wanted to post this question, as I see the UI wouldn't have support to have any random value as `violation_close_timer` - would like to hear your thoughts on this 😄 

image

founddrama: @pranav-new-relic The Terraform provider is performing validation above and beyond what's taking place in the underlying API. Ultimately, that's the source of this regression to begin with.

The change here will accept _any integer between 1 and 720_, representing the `violation_close_timer` as a value _in hours_.

`violation_close_timer` is mainly kept around for backward compatibility. Its previous allowable values (i.e., 1, 2, 4, 8, 12, 24) are a subset of the current valid rules. If you examine these conditions in the UI, you'll see that it accepts values as low as 5 minutes. However, to add support for minutes in Terraform for this condition type would be a breaking change for this field. The change in this PR allows existing conditions (i.e., those using one of the six previously restricted valid values) to continue working.

**As a side note:** the screenshot you're including appears to be from the NRQL condition UI and _not_ from the APM condition UI. NRQL conditions are configured using a different Terraform resource (i.e., `nrql_alert_condition`) which does not expose the `violation_close_timer` and instead uses `violation_time_limit_seconds`.

pranav-new-relic: Oops, so sorry - I realised just now that was out of an NRQL Alert Condition. Thank you for clarifying, and the detailed explanation, @founddrama - really appreciate it 😄 I'll get this merged soon.

pranav-new-relic:
@founddrama hi - I just have a few more thoughts about what we're doing here that I'd like to hear your opinion on (not exactly a question - would just like to discuss the following with you to understand what you think)

Since we're amending the `violation_close_timer` attribute at a schema level (making `Default` 72) and `violation_close_timer` is an attribute that can be added with any alert condition type (though it is mandatory only when the condition is of the type `apm_jvm_metric` and `apm_app_metric` (and `condition_scope` is `instance`)) - do you think it would be fine to amend this at a schema level?

Why I ask this is, consider a user who hasn't used `violation_close_timer` because their configuration does not use alert conditions of the aforementioned two types. Now, with the change made here, they would find 72 applied to their alert condition since it is default, which they would (probably) not expect. I don't see much of an issue here, as they would have to infer from the docs that this field has now been changed to default to 72 (and hence, I think this should be perfectly fine); but still, would like to hear your thoughts if you think this is fine, if you think there's any other edge case, etc. Sorry about the long question - and thank you 😄

pranav-new-relic (approved):

PR #2409, opened 2023-06-30
fix(dashboards): draft of the fix to prevent drift in line widgets

pranav-new-relic:
WIP, doesn't yet fix the issue

### DIFF
--- a/newrelic/structures_newrelic_one_dashboard.go
+++ b/newrelic/structures_newrelic_one_dashboard.go
@@ -836,6 +836,10 @@ func flattenVariableNRQLQuery(in *entities.DashboardVariableNRQLQuery) []interfa
 	return out
 }
 
+//func hashString(v interface{}) int {
+//	return schema.HashString(v.(map[string]interface{}))
+//}
+
 // return []interface{} because Page is a SetList
 func flattenDashboardPage(in *[]entities.DashboardPage) []interface{} {
 	out := make([]interface{}, len(*in))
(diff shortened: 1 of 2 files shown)


pranav-new-relic:
closing as this PR comprises only experimentation and does not need to be visible

newrelic/structures_newrelic_one_dashboard.go
```diff
@@ -287,6 +287,7 @@ func expandDashboardPageInput(d *schema.ResourceData, pages []interface{}, meta
 			}
 		}
 		if widgets, ok := p["widget_line"]; ok {
+			widgets = widgets.(*schema.Set).List()
```
pranav-new-relic, line 290: (a)
This change, paired with `schema.TypeSet` in `widget_line` fix the issue seen with the drift. If the configuration and the state differ in nothing apart from the order of line widgets, no drift is shown.

But now, I run into a new problem where a drift is shown across all line widgets, even if only one attribute's value in one line widget is changes.

newrelic/structures_newrelic_one_dashboard.go
```diff
@@ -835,6 +836,10 @@ func flattenVariableNRQLQuery(in *entities.DashboardVariableNRQLQuery) []interfa
 	return out
 }
 
+//func hashString(v interface{}) int {
```
pranav-new-relic, line 839: (b)

Guessing the cause of this new problem is that we don't convert the list of line_widgets back into a set in the flatten functions of read, I tried doing so multifold, but haven't been able to land at a solution and am still running into panics. The changes below show my most recent try to convert the list of line widgets into a set, and the error I see is 

`panic: interface conversion: interface {} is map[string]interface {}, not string`

PR #2417, opened 2023-07-11
fix(service_level_alert_helper): Add bad events support

NilVentosa:
# Description

Service Levels can be set up using good or bad events. Until now the newrelic_service_level_alert_helper only considered good events. We added the boolean field `is_bad_events` as a optional field that defaults to false, as the most common case is good events.

## Type of change

## Checklist:

## How to test this change?
Use the helper with bad events, for example using the updated example in the documentation

### DIFF
--- a/website/docs/d/service_level_alert_helper.html.markdown
+++ b/website/docs/d/service_level_alert_helper.html.markdown
@@ -12,7 +12,7 @@ Use this data source to obtain the necessary fields to set up alerts on your ser
 
 ## Example Usage
 
-Firstly set up your service level objective, we recommend to use local variables for the `target` and `time_window.rolling.count`, as they are also necessary for the helper.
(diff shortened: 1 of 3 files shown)


website/docs/d/service_level_alert_helper.html.markdown
```diff
@@ -31,9 +31,9 @@ resource "newrelic_service_level" "foo" {
             from = "Transaction"
             where = "appName = 'Example application' AND (transactionType='Web')"
         }
-        good_events {
+        bad_events {
```
sanderblue, line 34: Instead of replacing `bad_events` in the example,  can we include examples using  `good_events` and `bad_events`? The more useful examples we have, the better for our customers 🙂

sanderblue (approved):

sanderblue (approved):

PR #2420, opened 2023-07-11
fix(alert_condition): reverting default violation_close_timer 72 to prevent failures

### LINKED ISSUE #2418: Regression after releasing NewRelic Teraform provider 3.25.1
Regression was detected after updating Terraform NewRelic provider 3.25.0 -> 3.25.1. I suspect the problem is the following commit: https://github.com/newrelic/terraform-provider-newrelic/pull/2402/files

# What is the behaviour/problem?
There are NewRelic Terraform resources for which  `violation_close_timer` is optional. In case of specific configurations using it is even not allowed (validation preventing defining such resources exists). Example `newrelic_alert_condition` with `type`: `apm_app_metric` and `condition_scope`: `application`. During trial to `terraform apply` such resources when user will not provide resource at all in Terraform code or will set `violation_close_timer` to null then `default=72` is applied which cause resource validation is failing

# What kind of resource is affected. In my case:
* `newrelic_alert_condition`

# What is error message
```
╷
│ Error: violation_close_timer only supported for apm_app_metric when condition_scope = 'instance'
│ 
│   with module.newrelic.module.app_alerts.newrelic_alert_condition.response_time_web["blahblah"],
│   on modules/newrelic/modules/app_alerts/main.tf line 170, in resource "newrelic_alert_condition" "response_time_web":
│  170: resource "newrelic_alert_condition" "response_time_web" {
```

# How resources are defined
```
resource "newrelic_alert_condition" "appdex" {
  for_each  =  toset(['<MY_APP_1>'])
  policy_id = '<POLICY_ID>'

  name            = "<ALERT_NAME>"
  type            = "apm_app_metric"
  entities        = [<ENTITIES>]
  metric          = "apdex"
  condition_scope = "application"

  term {
    duration      = 5
    operator      = "below"
    priority      = "warning"
    threshold     = "0.85"
    time_function = "all"
  }

  term {
    duration      = 5
    operator      = "below"
    priority      = "critical"
    threshold     = "0.7"
    time_function = "all"
  }
}
```

```
resource "newrelic_alert_condition" "response_time_web" {
  for_each = toset([<MY_APP_1>])
  policy_id = <POLICY_ID>

  name            = "<ALERT_NAME>"
  type            = "apm_app_metric"
  entities        = [<ENTITY>]
  metric          = "response_time_web"
  condition_scope = "application"

  term {
    duration      = 15
    operator      = "above"
    priority      = "critical"
    threshold     = "1.0"
    time_function = "all"
  }
}
```

# Additional info
When using old pipelines with old version of NewRelic Terraform provider (`3.25.0`) everything is working as expected

pranav-new-relic:
# Description

This PR fixes the following linked issue that was caused by a PR merged earlier (https://github.com/newrelic/terraform-provider-newrelic/pull/2402/files), which is now leading to an issue in which the attribute `violation_close_timer` is defaulting to 72 for all types of conditions, which is not expected, and is hence, leading to an API validation issue.

Since the approach towards conditionally varying this default value only in two cases (when `type` is `apm_jvm_metric` (OR) when `type` is `apm_app_metric` and `condition_scope` is `instance`) is not straightforward yet (this might need modifications directly in the create/update functions to set defaults only when these conditions are true, and proper state management upon read, to infer that 72 is the default value and is expected to cause no drift), this PR fixes the issue in the interim by reverting the default value set.

Fixes #2418

### DIFF
--- a/newrelic/resource_newrelic_alert_condition.go
+++ b/newrelic/resource_newrelic_alert_condition.go
@@ -137,8 +137,7 @@ func resourceNewRelicAlertCondition() *schema.Resource {
 				Type:         schema.TypeInt,
 				Optional:     true,
 				ValidateFunc: validation.IntBetween(1, 720),
-				Default:      72,
-				Description:  "Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be between 1 and 720 hours. Defaults to 72.",
+				Description:  "Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be between 1 and 720 hours.",
 			},
 			"gc_metric": {
 				Type:        schema.TypeString,
--- a/website/docs/r/alert_condition.html.markdown
+++ b/website/docs/r/alert_condition.html.markdown
@@ -99,7 +99,9 @@ The following arguments are supported:
   * `condition_scope` - (Required for some types) `application` or `instance`.  Choose `application` for most scenarios.  If you are using the JVM plugin in New Relic, the `instance` setting allows your condition to trigger [for specific app instances](https://docs.newrelic.com/docs/alerts/new-relic-alerts/defining-conditions/scope-alert-thresholds-specific-instances).
   * `enabled` - (Optional) Whether the condition is enabled or not. Defaults to true.
   * `gc_metric` - (Optional) A valid Garbage Collection metric e.g. `GC/G1 Young Generation`.
-  * `violation_close_timer` - (Optional) Automatically close instance-based incidents, including JVM health metric incidents, after the number of hours specified. Must be between 1 and 720 hours. Defaults to 72.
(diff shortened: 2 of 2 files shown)


NSSPKrishna (approved):

pranav-new-relic:
**FYI:** This PR is now being merged and released with v3.25.2 of the Terraform Provider to revert the change released with the previous version (`Default: 72` in `violation_close_timer` in this resource), in order to immediately unblock customers using other alert conditions. Additional implementation to conditionally default `violation_close_timer` will be picked up as part of a future ticket that has been created, soon. 

Changes made in this PR (rollback of https://github.com/newrelic/terraform-provider-newrelic/pull/2402/files) have been tested - tests show that the behaviour of this resource is now as in v3.25.0.

PR #2421, opened 2023-07-13
fix(browser_application): addition of browser properties

pranav-new-relic:
### Description

This PR adds a new computed (read-only) attribute `js_config` to the `newrelic_browser_application` resource, which would export the JavaScript configuration of the browser application created, which users would need to automate creating browser agents using these attributes exported from the resource.

### Background

When a BrowserApplicationEntity is queried, the response comprises `jsConfig` inside `browserProperties`, that contains the following information (sensitive details redacted)

**Request** 
```gql
{
  actor {
    entity(guid: "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX") {
      ... on BrowserApplicationEntity {
        guid
        name
        browserProperties {
          jsConfig
          jsConfigScript
          jsLoaderScript
        }
      }
    }
  }
}

```

**Response**
```gql
{
  "data": {
    "actor": {
      "entity": {
        "browserProperties": {
          "jsConfig": {
            "info": {
              "applicationID": XXXXXXXXX,
              "beacon": "xxx.nr-xxx.net",
              "errorBeacon": "xxx.nr-xxx.net",
              "licenseKey": "NRBR-XXXXXXXXXXXXXX",
              "sa": 1
            },
            "init": {
              "ajax": {
                "deny_list": [
                  "xxx.nr-xxx.net"
                ]
              },
              "distributed_tracing": {
                "enabled": true
              },
              "privacy": {
                "cookies_enabled": true
              }
            },
            "loader_config": {
              "accountID": XXXXXXX,
              "agentID": XXXXXXXXXX,
              "applicationID": XXXXXXXXXXX,
              "licenseKey": "NRBR-XXXXXXXXXXXXXXXX",
              "trustKey": XXXXXXX
            }
          },
    "jsLoaderScript" : "<Huge JS code snippet in a string>",
    "jsConfigScript" : "<Huge JS code snippet in a string>",
    ..
   ..
  ..
}
```

Users who would like to automate the procedure of setting up an NR browser agent via attributes exported from this Terraform Resource would primarily need `info`, `init` and `loader_config` in `jsConfig`. These are [inputs needed to be given to the NR browser agent](https://github.com/newrelic/newrelic-browser-agent).

However, the mutation to create a browser agent via TF does not return these attributes (as it isn't an entity that is returned by the mutation), and this happens only with the read query. Hence,

#### [1. Changes in Client Go](https://github.com/newrelic/newrelic-client-go/pull/1035)
- Update to API via Tutone to add these values to the `getEntities` and `getEntity` queries, one of which is used by the **Read** function of this Terraform resource

#### 2. Changes in The Terraform Provider (This PR): 
- Addition of `js_config` as a computed attribute that would return the aforementioned `jsConfig` as a marshalled JSON string (see usage in the screenshot below)
- Updates to the **create** and **update** functions of the resource to call the **read** method at the end, so as to have the `jsConfig` values obtained in the read function set to the state upon create/update. This would allow users to export these values upon creating/updating the resource.

### How do users use this?

image

```hcl
resource "newrelic_browser_application" "foo" {
  name = "Random Browser Application"
  cookies_enabled = true
  distributed_tracing_enabled = true
  loader_type = "SPA"
}

output "bar" {
  value = jsondecode(newrelic_browser_application.foo.js_config)
}
```


go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/newrelic/terraform-provider-newrelic/v2
 
 go 1.19
 
+// DISCARD THIS BEFORE MERGE!
+replace github.com/newrelic/newrelic-client-go/v2 => github.com/newrelic/newrelic-client-go/v2 v2.19.5-0.20230713153644-95afcde1f094
+
```
pranav-new-relic, line 7: @pranav-new-relic **DO NOT MERGE WITHOUT DISCARDING THIS**

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/newrelic/go-agent/v3 v3.20.3
 	github.com/newrelic/go-insights v1.0.3
-	github.com/newrelic/newrelic-client-go/v2 v2.21.1
+	github.com/newrelic/newrelic-client-go/v2 v2.21.2-0.20231011082018-047ffcefa229
```
pranav-new-relic, line 10: @pranav-new-relic **DO NOT MERGE WITHOUT DISCARDING/REVERTING THIS**

go.sum
```diff
@@ -273,8 +273,8 @@ github.com/newrelic/go-agent/v3 v3.20.3 h1:hUBAMq/Y2Y9as5/yxQbf0zNde/X7w58cWZkm2
 github.com/newrelic/go-agent/v3 v3.20.3/go.mod h1:rT6ZUxJc5rQbWLyCtjqQCOcfb01lKRFbc1yMQkcboWM=
 github.com/newrelic/go-insights v1.0.3 h1:zSNp1CEZnXktzSIEsbHJk8v6ZihdPFP2WsO/fzau3OQ=
 github.com/newrelic/go-insights v1.0.3/go.mod h1:A20BoT8TNkqPGX2nS/Z2fYmKl3Cqa3iKZd4whzedCY4=
-github.com/newrelic/newrelic-client-go/v2 v2.21.1 h1:F00Sv41LTSm8edt1Ld2qnV7Jw0lRee8t921odyaK3nc=
-github.com/newrelic/newrelic-client-go/v2 v2.21.1/go.mod h1:VPWTvEfKvnTZLunAC7fiW33y4e0srznNfN5HJH2cOp8=
+github.com/newrelic/newrelic-client-go/v2 v2.21.2-0.20231011082018-047ffcefa229 h1:6IiwH3IPjeEYyctPuQW3HDr7yee38yO9zS9TnLSZg/Y=
+github.com/newrelic/newrelic-client-go/v2 v2.21.2-0.20231011082018-047ffcefa229/go.mod h1:VPWTvEfKvnTZLunAC7fiW33y4e0srznNfN5HJH2cOp8=
```
pranav-new-relic, line 277: @pranav-new-relic **DO NOT MERGE WITHOUT DISCARDING/REVERTING THIS**

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/newrelic/go-agent/v3 v3.20.3
 	github.com/newrelic/go-insights v1.0.3
-	github.com/newrelic/newrelic-client-go/v2 v2.21.1
+	github.com/newrelic/newrelic-client-go/v2 v2.21.2-0.20231011094303-bbae780fcced
```
pranav-new-relic, line 10: Please revert this change before you merge the PR, @pranav-new-relic 
This change was added to allow compilation/integration tests to succeed as they would need to point to the changes in the Go-Client counterpart of this PR.

go.sum
```diff
@@ -273,8 +273,8 @@ github.com/newrelic/go-agent/v3 v3.20.3 h1:hUBAMq/Y2Y9as5/yxQbf0zNde/X7w58cWZkm2
 github.com/newrelic/go-agent/v3 v3.20.3/go.mod h1:rT6ZUxJc5rQbWLyCtjqQCOcfb01lKRFbc1yMQkcboWM=
 github.com/newrelic/go-insights v1.0.3 h1:zSNp1CEZnXktzSIEsbHJk8v6ZihdPFP2WsO/fzau3OQ=
 github.com/newrelic/go-insights v1.0.3/go.mod h1:A20BoT8TNkqPGX2nS/Z2fYmKl3Cqa3iKZd4whzedCY4=
-github.com/newrelic/newrelic-client-go/v2 v2.21.1 h1:F00Sv41LTSm8edt1Ld2qnV7Jw0lRee8t921odyaK3nc=
-github.com/newrelic/newrelic-client-go/v2 v2.21.1/go.mod h1:VPWTvEfKvnTZLunAC7fiW33y4e0srznNfN5HJH2cOp8=
+github.com/newrelic/newrelic-client-go/v2 v2.21.2-0.20231011094303-bbae780fcced h1:Qk2TPqUXjQEoZGFNvgHFD03riHzYUTb+jZjQq8gsPbI=
+github.com/newrelic/newrelic-client-go/v2 v2.21.2-0.20231011094303-bbae780fcced/go.mod h1:VPWTvEfKvnTZLunAC7fiW33y4e0srznNfN5HJH2cOp8=
```
pranav-new-relic, line 277: Please revert this change before you merge the PR, @pranav-new-relic 
This change was added to allow compilation/integration tests to succeed as they would need to point to the changes in the Go-Client counterpart of this PR.

mbazhlekova (approved):
Looks good! Do you think we need to update the integration test?

newrelic/resource_newrelic_agent_application_browser_integration_test.go
```diff
@@ -38,6 +39,8 @@ func TestAccNewRelicAgentApplicationBrowser(t *testing.T) {
 				),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckNewRelicAgentApplicationBrowserExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "js_config"),
+					testAccCheckJsConfigNestedAttributes(resourceName, "js_config", []string{"init", "info", "loader_config"}),
```
pranav-new-relic, line 43: Hi @mbazhlekova - these are the changes I've made to the current test case which does two things - 
- check if the resource comprises (exports) an attribute named `js_config`
- since `js_config` is exported as a string that is converted into a JSON using the `jsondecode()` function in HCL, apply a similar mechanism in the test to convert the string into a JSON and iterate through the keys of the JSON to see if `init`, `info` and `loader_config` are present.

I also felt it might be a good idea the `PreCheck` function to `testAccBrowserApplicationsCleanup` as I found a large number of redundant browser applications in our test New Relic account (created by this test) which would be good to be cleaned up before we run this test.

With both of these changes, upon local tests, I find that the test passes.

Please let me know if this works - I'm also wondering if the update made to the current test case is fine, or this would be better as a new test case. Thank you 🙂

mbazhlekova: Adding a cleanup function is a great idea! 👍

PR #2429, opened 2023-07-19
fix(cloud_aws_link_account): refactored timeout to discard limit on create

pranav-new-relic:
### The Problem

As observed in a recent GTSE, the `newrelic_cloud_aws_link_account` resource usually works fine when the ARN it takes is hardcoded; however, when the ARN is generated by other AWS resources, such as in this sample configuration

https://github.com/newrelic/terraform-provider-newrelic/blob/refactor/update-aws-example/examples/modules/cloud-integrations/aws/main.tf#L71-L80

because AWS takes time to validate the ARN created, the API that is called when a `newrelic_cloud_aws_link_account` resource needs to be created returns a **failed** status for the first few times, suggesting the ARN is invalid (because AWS hasn't yet validated this ARN, as it has just been created via some other AWS resource). Because the method to create is wrapped in `RetryContext`, the API is called until it succeeds, but this times out after 10 seconds because of the timeout we currently have in place.

https://github.com/newrelic/terraform-provider-newrelic/blob/refactor/update-aws-example/newrelic/resource_newrelic_cloud_aws_link_account.go#L50-L54

As a result, even if the resource succeeds to be created (with the API showing a successful status after four failed retries) and these linked accounts are shown in the UI, `terraform apply` does not save this resource to the state, because it throws the error seen upon failure (because it took >10 seconds to obtain a successful status from the API).

```
╷
│ Error: ERR_INVALID_DATA : Validation failed: The ARN you entered does not permit the correct access to your AWS account - please check the ARN and try again.
│ 
│   with newrelic_cloud_aws_link_account.newrelic_cloud_integration_push,
│   on main.tf line 70, in resource "newrelic_cloud_aws_link_account" "newrelic_cloud_integration_push":
│   70: resource "newrelic_cloud_aws_link_account" "newrelic_cloud_integration_push" {
│ 
╵
```

### The Solution

Remove the timeout and allow the resource to use Terraform's default timeout of 20 minutes. (as specified [here in the docs](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts#default-timeouts-and-deadline-exceeded-errors))

### Testing

This was tested in the following ways. 
- The user who reported the GTSE was asked to add a `timeout` in this resource's configuration, which would help override the 10-second timeout defined in the resource. I tested this myself and saw this to be working, and it worked for the user too.
- I then tested the same with this code change, and observed the same results.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/newrelic/go-agent/v3 v3.20.3
 	github.com/newrelic/go-insights v1.0.3
-	github.com/newrelic/newrelic-client-go/v2 v2.19.4
+	github.com/newrelic/newrelic-client-go/v2 v2.19.6
 	github.com/stretchr/testify v1.8.2
 )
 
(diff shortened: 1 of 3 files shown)


pranav-new-relic:
Merging this PR as 
- this is identical to the fix made in #2423 
- this helped close the GTSE and has been tested thoroughly

PR #2462, opened 2023-09-08
chore(aws_cloud_integrations): fix to aws cloud integration tests

pranav-new-relic:
WIP (to be updated after this is finalized)
In short, this PR is made to experiment fixes to be made to AWS-integration integration tests, because it appears that the behaviour of the API has changed in recent times, when a large number of integrations are applied/created in one apply.


pranav-new-relic:
Can be closed, invalid PR as the API has now been fixed to match expected behaviour, and the integration tests need no modification.

PR #2466, opened 2023-09-13
fix(aws_cloud_integrations): WIP fix to removal of arguments from integrations

pranav-new-relic:
WIP fix for https://issues.newrelic.com/browse/NR-158475 / https://issues.newrelic.com/browse/NR-160342
As on Sep 13, this works for the scenario in the bug (when arguments are commented) but breaks expected behaviour (for instance, when a new integration is added to update Terraform configuration) and hence, needs more investigation for the right fix

### DIFF
--- a/newrelic/structures_newrelic_cloud_aws_integrations.go
+++ b/newrelic/structures_newrelic_cloud_aws_integrations.go
@@ -1,6 +1,8 @@
 package newrelic
 
 import (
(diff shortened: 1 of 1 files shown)


pranav-new-relic:
Closing this PR as it does not appear feasible at the moment (this is causing other breaking changes). Can be revisited at a later point of time.

PR #2487, opened 2023-10-13
fix(muting_rule): error handling update to throw non-GraphQL errors too

pranav-new-relic:
#### TL;DR
The error handling mechanism of `newrelic_alert_muting_rule` has a small issue; it currently doesn't throw errors which are not of the type `alerts.GraphQLErrorResponse`.

The changes in this PR wrap the block of code throwing the error around another condition which throws other errors too. (with an `else`

#### Summary 

An issue identified by a customer points that the error handling mechanism of the resource `newrelic_alert_muting_rule` is broken. When a few types of errors are returned by the step in the code that creates a muting rule via NerdGraph, the error is not thrown; instead, the provider runs into a panic, and crashes.

```
Stack trace from the terraform-provider-newrelic_v3.21.3 plugin:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x18086d2]
goroutine 65 [running]:
github.com/newrelic/terraform-provider-newrelic/v2/newrelic.resourceNewRelicAlertMutingRuleCreate(
{0x1ca0a68, 0xc00079fbc0}
, 0x0?,
{0x19a5bc0?, 0xc0002cbc40?}
)
/home/runner/work/terraform-provider-newrelic/terraform-provider-newrelic/newrelic/resource_newrelic_alert_muting_rule.go:234 +0x232
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).create(0xc0000abce0,
..
..
```

#### Cause

It is evident from the above stack trace that line 234 in the resource is the point of origin of this error.

`/home/runner/work/terraform-provider-newrelic/terraform-provider-newrelic/newrelic/resource_newrelic_alert_muting_rule.go:234 +0x23`

https://github.com/newrelic/terraform-provider-newrelic/blob/660a2b00ebd18c454715d11736ebec6500f9db64/newrelic/resource_newrelic_alert_muting_rule.go#L234

This line in the code sets the ID of the resource created by concatenating the account ID with the ID of the muting rule.
The reason why this line is seen to throw the following error

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x18086d2]
```
is because it is trying to access **created.ID** when created has probably not been returned (or has been returned as nil from NerdGraph) in one of the previous steps.

```
created, err := client.Alerts.CreateMutingRuleWithContext(ctx, accountID, createInput)
```
and hence, a nil pointer exception is thrown.

This is happening because when in the aforementioned line **created** is nil, **err** would not be nil, as an non null error would lead to **created** being nil.

```
created, err := client.Alerts.CreateMutingRuleWithContext(ctx, accountID, createInput)
```

In such a case, we're only throwing the error if it is of the type 

```
if graphQLError, ok := err.(*alerts.GraphQLErrorResponse); ok {
```
and not if it is an error of any other class; and hence, the provider crashes, since no error is thrown when **created** is nil and err is not nil, unless the error is of a certain type.

Hence, this PR comprises the change discussed in the **TL;DR** section above.

#### Tests Done

I've tested these changes with both validation errors being thrown and other errors
- **Validation Errors**: It doesn't look like validation errors can be reproduced as this resource has too many `ValidateFunc`s in place :) however, to reproduce this, I temporarily removed the validation function on operators in conditions, and tried applying a Terraform configuration with one of the operators as "FAKE". The validation error is thrown, and remains unaffected by the changes in this PR (the same error is thrown without these changes too)
image

- **Other Errors**: By performing a terraform apply with a command that re-routes the request via a proxy, the following error was thrown, which earlier, was not thrown by the resource. 
image

### DIFF
--- a/newrelic/structures_newrelic_alert_muting_rule.go
+++ b/newrelic/structures_newrelic_alert_muting_rule.go
@@ -1,10 +1,12 @@
 package newrelic
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"time"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/newrelic/newrelic-client-go/v2/pkg/alerts"
 )
@@ -385,4 +387,38 @@ func flattenWeeklyRepeatDays(daysOfWeek []alerts.DayOfWeek) []string {
 	}
 
 	return out
+}
+
+func handleMutingRuleCreateUpdateError(err error) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	var graphQLError *alerts.GraphQLErrorResponse
+	if errors.As(err, &graphQLError) {
+		for _, e := range graphQLError.Errors {
+			var message string = e.Message
+			var errorClass string = e.Extensions.ErrorClass
+			var validationErrors = e.Extensions.ValidationErrors
+
+			if len(validationErrors) == 0 {
+				diags = append(diags, diag.Diagnostic{
+					Severity: diag.Error,
+					Summary:  message + ": " + errorClass,
+				})
+			} else {
+				for _, validationError := range validationErrors {
+					diags = append(diags, diag.Diagnostic{
+						Severity: diag.Error,
+						Summary:  message + ": " + errorClass,
+						Detail:   validationError.Name + ": " + validationError.Reason,
+					})
+				}
+			}
+		}
+	} else {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  err.Error(),
+		})
+	}
+	return diags
 }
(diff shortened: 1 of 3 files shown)


newrelic/structures_newrelic_alert_muting_rule.go
```diff
@@ -386,3 +387,36 @@ func flattenWeeklyRepeatDays(daysOfWeek []alerts.DayOfWeek) []string {
 
 	return out
 }
+
+func handleMutingRuleCreateUpdateError(err error) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	if graphQLError, ok := err.(*alerts.GraphQLErrorResponse); ok {
```
NSSPKrishna, line 394:
```suggestion
        var graphQLError *alerts.GraphQLErrorResponse
	if errors.As(err, &graphQLError)
```
Using the following code would be a better approach:

pranav-new-relic: Thank you, was a very valuable suggestion 🙂 changed this in the latest commit.

NSSPKrishna (approved):
The code looks good to me.

PR #2482, opened 2023-10-05
Update validation on violation_time_limit_seconds

rsdoherty:
# Description

As per the docs, this field accepts up to 30 days: https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/advanced-alerts/rest-api-alerts/alerts-conditions-api-field-names/

Fixes #2475 

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

Please delete options that are not relevant.

- [x] My commit message follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] My code is formatted to [Go standards](https://go.dev/blog/gofmt)
- [x] I have performed a self-review of my own code
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes. Go [here](https://github.com/newrelic/terraform-provider-newrelic/blob/main/CONTRIBUTING.md#testing) for instructions on running tests locally.

## How to test this change?

Run `terraform plan` using the configuration documented in the issue: https://github.com/newrelic/terraform-provider-newrelic/issues/2475

PR #2493, opened 2023-10-23
fix(multilocation_alert_condition): make violation_time_limit_seconds nearly consistent with its usage in nrql_alert_condition

pranav-new-relic:
TBD
(adapts changes in #2482)

### DIFF
--- a/newrelic/resource_newrelic_synthetics_multilocation_alert_condition.go
+++ b/newrelic/resource_newrelic_synthetics_multilocation_alert_condition.go
@@ -11,7 +11,7 @@ import (
 	"github.com/newrelic/newrelic-client-go/v2/pkg/errors"
 )
 
-// syntheticsMultiLocationConditionTermSchema returns the schema used for a critial or warning term priority.
+// syntheticsMultiLocationConditionTermSchema returns the schema used for a critical or warning term priority.
 func syntheticsMultiLocationConditionTermSchema() *schema.Resource {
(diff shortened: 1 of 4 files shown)


newrelic/resource_newrelic_synthetics_multilocation_alert_condition.go
```diff
@@ -81,9 +81,18 @@ func resourceNewRelicSyntheticsMultiLocationAlertCondition() *schema.Resource {
 			},
 			"violation_time_limit_seconds": {
 				Type:         schema.TypeInt,
-				Required:     true,
-				ValidateFunc: validation.IntInSlice([]int{0, 3600, 7200, 14400, 28800, 43200, 86400}),
-				Description:  "The maximum number of seconds an incident can remain open before being closed by the system.  Must be one of: 0, 3600, 7200, 14400, 28800, 43200, 86400",
+				Optional:     true,
+				ValidateFunc: validation.IntBetween(0, violationTimeLimitSecondsMax),
```
pranav-new-relic, line 85: This is supposed to be 300 - 2592000 as also specified in New Relic docs, but I've added 0 as the minimum as this resource is already in use with 0 as a valid value of `violation_time_limit_seconds`.

However, the current behaviour of the API is such that if `0` is given as a value against this argument, the API defaults it to 259200 seconds. This is why the below `DiffSuppressFunc` has been added.

PR #2507, opened 2023-11-07
chore(synthetics): refactor validate method to throw a deprecation warning with MUTED status

pranav-new-relic:
### Summary of Changes
This PR addresses the following - 
- modifying the validate function of the status attribute of all Synthetic Monitor resources to a new function, `validateSyntheticMonitorStatus`. This function checks if the status is valid (similar to current behaviour) but also helps throw a warning if the status is `MUTED`. Examples below.
- making consistent the statement talking about the deprecation across docs, the warning message and the description of the `status` attribute in the code

### Examples

Case 1: If one tries creating a monitor with status `MUTED` with these new changes

image

image

Case 2: If one tries modifying a monitor with status `MUTED` (with `MUTED` in the configuration)

image

However, this is shown only if the current status in the configuration is `MUTED` and not if we try changing it to `ENABLED` or `DISABLED`.

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/newrelic/go-insights v1.0.3
 	github.com/newrelic/newrelic-client-go/v2 v2.21.2
 	github.com/stretchr/testify v1.8.4
+	golang.org/x/exp v0.0.0-20231006140011-7918f672742d
 )
 
 require (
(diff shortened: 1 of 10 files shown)


pranav-new-relic:
Unrelated: The only failing integration tests are from Azure. I am currently in conversation with Maria on this, to see how we can get them updated.

pranav-new-relic:
Good to be merged - tested this with all five Synthetic monitor resources, and this works as expected.

PR #2580, opened 2024-02-15
docs: Added import information to notification channel and destinatio…

amaor-newrelic:
# Description
Added Import documentation to newrelic_notification_destination and newrelic_notification_channel

### DIFF
--- a/website/docs/r/notification_channel.html.markdown
+++ b/website/docs/r/notification_channel.html.markdown
@@ -181,7 +181,7 @@ resource "newrelic_notification_channel" "foo" {
 
   property {
     key = "service"
-    label = "Service Name" // Name of your service. If left out will be blank in UI.
+    label = "Service Name" // Label of service tab in channel builder. This will show as blank if left empty.
     value = "PTQK3FM"
   }
 
--- a/website/docs/r/notification_channel.html.markdown
+++ b/website/docs/r/notification_channel.html.markdown
@@ -181,7 +181,7 @@ resource "newrelic_notification_channel" "foo" {
 
   property {
     key = "service"
-    label = "Service Name" // Label of service tab in channel builder. This will show as blank if left empty.
+    label = "Service Name" // Label of service drop-down value in the channel builder. This will show as blank if left empty.
     value = "PTQK3FM"
   }
 
--- a/website/docs/r/notification_channel.html.markdown
+++ b/website/docs/r/notification_channel.html.markdown
@@ -181,6 +181,7 @@ resource "newrelic_notification_channel" "foo" {
 
   property {
     key = "service"
+    label = "Service Name" // Name of your service. If left out will be blank in UI.
     value = "PTQK3FM"
   }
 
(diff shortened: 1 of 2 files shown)


website/docs/r/notification_channel.html.markdown
```diff
@@ -181,6 +181,7 @@ resource "newrelic_notification_channel" "foo" {
 
   property {
     key = "service"
+    label = "Service Name" // Name of your service. If left out will be blank in UI.
```
amaor-newrelic, line 184: Unrelated, but could lower some complaints about it being blank in the future 😅

website/docs/r/notification_destination.html.markdown
```diff
@@ -223,16 +223,21 @@ resource "newrelic_notification_destination" "foo" {
 
 #### [Slack](https://docs.newrelic.com/docs/alerts-applied-intelligence/notifications/notification-integrations/#slack)
 
-Slack destinations are created by actively interacting with the UI. This is how OAuth authentication works and there is no way to authenticate otherwise. Because of this, there is no way to create a slack destination via terraform, and must be created in the UI and imported into terraform.
+Slack destinations are created by actively interacting with the UI. This is how OAuth authentication works and there is no way to authenticate otherwise. Because of this, there is no way to create a slack destination via terraform, and must be created in the UI and imported into terraform. More information in the [Import](#import) section below.
```
amaor-newrelic, line 226: I'm thinking of also stating that slack destinations are best used as data sources and not as resources, wdyt?

website/docs/r/notification_channel.html.markdown
IliaShurygin: I don't really understand what this comment means, so I am not really sure it would help customers...
Could elaborate what "service" is in this context? PD service? Also, what will be blank in the UI? Does it have to be the name of your PD services or would any name work?

Honestly, I am not really sure if this field affects UI at all: there's no separate UI for channels, and workflows do not rely on channel labels. Maybe the workflow builder UI does fetch channel name in the background, but I am not sure about this either...

website/docs/r/notification_destination.html.markdown
IliaShurygin: I know that you did not change this part, but there's a typo in "and must be created" (should be "and it must be created")

Frankly, I would rephrase this whole part to something like. OAuth part does not really matter.

> In order to create a Slack destinations, you have to go grant our application access to your workspace. This process is [based on OAuth](https://api.slack.com/authentication/oauth-v2) can only be done through a browser. 
As a result, you cannot set up a Slack destination purely with Terraform code. 
You can either create a Slack destination in the interface and [import](#import) it to Terraform, or use destination data source.

TODO: add a link to the destination data source page

amaor-newrelic: 👍

website/docs/r/notification_channel.html.markdown
amaor-newrelic: After speaking, write explicitly what this label is and why they should fill it.

IliaShurygin (approved):

PR #2590, opened 2024-02-28
feat(destinations): expose destination entity guid

nzur-newrelic:
# Description

Exposing the `guid` field on notification destination in the resource and data source. 

## Type of change

## Checklist:

## How to test this change?

- Destination should now have a computed GUID field that can be used in the script


pranav-new-relic:
@nzur-newrelic 👋 

I just found that two unit tests were failing as the datatype to be used was supposed to be `notifications.EntityGUID` and not `destinations.EntityGUID` 🙂 so I just made that change. 

However, the unit test `TestFlattenNotificationDestination` seems to fail with your changes to add `guid` to the test, which is because I find changes to add `guid` as a computed argument to the `newrelic_notification_destination` resource haven't been made, in addition to which the change to set the `guid` attribute in the read function of the `newrelic_notification_destination` resource (i.e., in the function `flattenNotificationDestination`) would need to be made. In short, the changes you've added to export `guid` currently seem to be added to tests of both the data source and the resource, but only to the implementation of the data source - please add them to the implementation of the resource too, so we can see this test pass 🙂 

P.S. (just thinking aloud) this should have been captured by the integration test you modified too, as the resource isn't currently exporting `guid` but the test includes this change; however, the integration test CI task hasn't run on this PR at all; not sure why, probably because of the semantics of a one line change 🙂

nzur-newrelic:
>  this should have been captured by the integration test you modified too, as the resource isn't currently exporting `guid` but the test includes this change; however, the integration test CI task hasn't run on this PR at all; not sure why, probably because of the semantics of a one line change.. not sure why this happened.
My hunch is that they didn't run because the unit tests failed but again it's just a hunch

pranav-new-relic:
@nzur-newrelic I find the resource `newrelic_notification_destination` still doesn't have the following 
```
			"guid": {
				Type:        schema.TypeString,
				Computed:    true,
				Description: "Destination entity GUID",
			},
```

which is why the unit test is failing again, as the argument `guid` it's trying to set the GUID to doesn't exist in the schema of the resource.

pranav-new-relic:
The integration tests are still not running; I'll try finding out why. Congrats on successful unit tests :) 

I have one (possibly final) comment, however - I guess a tiny documentation change is warranted in
- the data source: https://github.com/newrelic/terraform-provider-newrelic/blob/v3.32.0/website/docs/d/notification_destination.html.markdown#attributes-reference
- the resource: https://github.com/newrelic/terraform-provider-newrelic/blob/v3.32.0/website/docs/r/notification_destination.html.markdown#attributes-reference

to make clear in the docs, that `guid` is now going to be exported too. Can you please make this change @nzur-newrelic?

pranav-new-relic (approved):

PR #2599, opened 2024-03-06
newrelic_entity - Ignore not found parameter

gabrielqs:
# Description

Expiration for lambda entities is set to 1 day. If a lambda is idle for a couple of days, AWS won't emit any metrics for it, and new relic will expire the entity and remove the lambda from the database. Whenever the lambda is triggered again, metrics will be sent over to NR and then the lambda entity will again exist. This happens often in our stage environment during weekends. What happens then whenever we run terraform apply Monday is that this resource will try to load the lambdas that existed friday and will throw an error until the lambda is triggered again.

This PR adds a new parameter to the `newrelic_entity` data source called `ignore_not_found`. This parameter defaults to false. When explicitly set to `true` it will suppress the exception. In this case the entity is returned but with null values set to all of its attribute

PS: This is my first PR, so please take a breath. I'll address any comments/tests/docs/concerns in follow up commits. Just want to have an idea of how much this needs to change and if we can even consider merging something like this.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] This change requires a documentation update

## Checklist:

- [X] My commit message follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] My code is formatted to [Go standards](https://go.dev/blog/gofmt)
- [X] I have performed a self-review of my own code
- [X] New and existing unit tests pass locally with my changes. Go [here](https://github.com/newrelic/terraform-provider-newrelic/blob/main/CONTRIBUTING.md#testing) for instructions on running tests locally.

## How to test this change?

- Load an entity which doesn't exist using the `newrelic_entity` data source, don't set `ignore_not_found`, you will see an exception saying the entity wasn't found and the plan execution will halt
- Load an entity which doesn't exist using the `newrelic_entity` data source, set `ignore_not_found` to `true`, you will be able to proceed with your plan execution, adding conditionals do your code where you expect entities can not exist within your code.

### DIFF
--- a/newrelic/data_source_newrelic_entity.go
+++ b/newrelic/data_source_newrelic_entity.go
@@ -86,6 +86,12 @@ func dataSourceNewRelicEntity() *schema.Resource {
 				Computed:    true,
 				Description: "A unique entity identifier.",
 			},
+			"ignore_not_found": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Do not thrown an error when the entity isn't found.",
+			},
 		},
 	}
 }
@@ -144,7 +150,13 @@ func dataSourceNewRelicEntityRead(ctx context.Context, d *schema.ResourceData, m
 	}
 
 	if entity == nil {
-		return diag.FromErr(fmt.Errorf("no entities found for the provided search parameters, please ensure your schema attributes are valid"))
+		if d.Get("ignore_not_found").(bool) {
+			log.Printf("[INFO] Entity not found, ignoring error")
+			d.SetId("")
+			return nil
+		} else {
+			return diag.FromErr(fmt.Errorf("no entities found for the provided search parameters, please ensure your schema attributes are valid"))
+		}
 	}
 
 	return diag.FromErr(flattenEntityData(entity, d))


newrelic/data_source_newrelic_entity.go
```diff
@@ -144,7 +150,13 @@ func dataSourceNewRelicEntityRead(ctx context.Context, d *schema.ResourceData, m
 	}
 
 	if entity == nil {
-		return diag.FromErr(fmt.Errorf("no entities found for the provided search parameters, please ensure your schema attributes are valid"))
+		if d.Get("ignore_not_found").(bool) {
+			log.Printf("[INFO] Entity not found, ignoring error")
+			d.SetId("")
+			return nil
```
pranav-new-relic, line 156:
```suggestion
			var diags diag.Diagnostics
			diags = append(diags, diag.Diagnostic{
				Severity: diag.Warning,
				Summary:  "no entities found for the provided search parameters, please ensure your schema attributes are valid",
			})
			return diags
```

(please **do not** commit this already - this is a suggestion I have, in order to return a warning instead of an error when an entity is not found and the `ignore_not_found` attribute is used, to avoid halting execution caused by an error; but I'd like more in my team to take a look at this and judge if this change makes sense. Cheers!)

pranav-new-relic:
Hi @gabrielqs, congratulations on your first PR :) it looks great, and aims at solving a real problem you've quoted in the description. The changes look good to me so far; however, I'd like more on my team to take a look at this PR to _see if we can even merge something like this_, like you mentioned, to understand how appropriate this change would be, from a technical standpoint. I shall be forwarding this PR for further review in the next couple of days - please keep an eye as we continue to do so :) 

Thanks for the contribution again!

newrelic/data_source_newrelic_entity.go
```diff
@@ -86,6 +86,12 @@ func dataSourceNewRelicEntity() *schema.Resource {
 				Computed:    true,
 				Description: "A unique entity identifier.",
 			},
+			"ignore_not_found": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Do not thrown an error when the entity isn't found.",
```
pranav-new-relic, line 93:
```suggestion
				Description: "Do not throw an error when the entity isn't found.",
```

newrelic/data_source_newrelic_entity.go
gabrielqs: ok, that makes sense. thank you

gabrielqs:
hello,

any new on this PR being reviewed? thank you in advance

pranav-new-relic:
@gabrielqs here's some good news - this change is now out with v3.34.1 of the New Relic Terraform Provider, released just a couple of minutes ago. Please give the `ignore_not_found` argument a try :)

pranav-new-relic:
FYI for the New Relic team (future reference): though the PR was closed earlier, an identical solution has recently been merged via #2628 and released (see the above comment).

gabrielqs:
thank you @pranav-new-relic!

Issue #2666, opened 2024-05-21
`newrelic_entity` Data Source: The data source does not return entity tags


Hasgaroth:
Hi,
We are having an issue with the NewRelic data source `newrelic_entity` which is not returning any tags associated with the resources that are found through the use of the data source.

## Affected Resource

- `newrelic_entity` data source

## Version information

```
Terraform v1.5.5
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v5.50.0
+ provider registry.terraform.io/newrelic/newrelic v3.36.0
```

## Current Behaviour

- The `newrelic_entity` data source does `not` return the tags associated with the found entity

## Expected Behaviour

- The `newrelic_entity` data source returns a data map containing the tags associated with the found entity

## Full Description

The [documentation](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/data-sources/entity.html) for this data source gives the option of providing tags as part of the search criteria, and for the output attributes that are exported, it states `In addition to all arguments above, the following attributes are exported`. So we would expect that all the tags that are assigned to the discovered entity will be exported with the rest of the attributes.

Unfortunately, there are no tags exported as part of the object attribute data.

```
newrelic_entity = {
  account_id                 = 1234567
  application_id             = null
  domain                     = "AIOPS"
  guid                       = "redacted_guid_value"
  id                         = "redacted_guid_value"
  ignore_case                = false
  ignore_not_found           = true
  name                       = "Entity name"
  serving_apm_application_id = null
  tag                        = null
  type                       = "CONDITION"
}
```

This is now causing a perpetual `diff` in our Terraform `plan` results, as we have an automated process that adds tags to resources based on other tag information that is present. The NewRelic provider does not give us the ability to ignore a set of these added tags, so decides that the tags need to be removed.

Running the `apply` for the resources then removes these tags, but the automated process re-adds them the next time it runs, meaning that the Terraform `plan` output will again show the drift in the tags.

The AWS provider gives the ability to provide a set of default tags for all resources, but also the ability to ignore certain tags so that they do not conflict with the Terraform controlled tags.

```
provider "aws" {
  # Other provider config items

  ignore_tags {
    key_prefixes = ["prefix_to_ignore"]
  }

  default_tags {
    tags = { set of tags to add to all resources }
  }
}
```

[...]

PR #2682, opened 2024-05-31
feat: add support for tags to the `newrelic_entity` data source

abasha1234:
# data_source_newrelic_entity

Fixes # [issue-2666](https://github.com/newrelic/terraform-provider-newrelic/issues/2666)

In the GitHub issue linked above, it has been reported as feedback from a customer that the `newrelic_entity` data source in the New Relic Terraform Provider currently does not return the tags of the entity fetched. This PR will add support for `tags` to the `data_source_newrelic_entity`


website/docs/d/entity.html.markdown
```diff
@@ -165,12 +165,65 @@ In addition to all arguments above, the following attributes are exported:
 * `guid` - The unique GUID of the entity.
 * `application_id` - The domain-specific application ID of the entity. Only returned for APM and Browser applications.
 * `serving_apm_application_id` - The browser-specific ID of the backing APM entity. Only returned for Browser applications.
+* `entity_tags` - The entity_tags attribute in the New Relic Terraform provider is used to store the tags associated with a New Relic entity in a JSON string format within the Terraform state.
```
pranav-new-relic, line 168:
```suggestion
* `entity_tags` - A JSON-encoded string, comprising tags associated with the entity fetched. 
   * See the **Additional Examples** section below, for an illustration depicting the usage of `jsondecode` with the attribute `entity_tags`, to get the tags associated with the entity fetched.
```

website/docs/d/entity.html.markdown
```diff
@@ -165,12 +165,66 @@ In addition to all arguments above, the following attributes are exported:
...
+
+The `local.key_value_map` would be:
+
+```hcl
+{
+  env = ["production"]
+  team = ["ops"]
+}
+```
+
```
pranav-new-relic, line 227:
```suggestion
### Using the `entity_tags` Attribute to Fetch Tags Associated with the Entity

As stated above in the **Attribute Reference** section, while the attribute `entity_tags` helps retrieve tags associated with the entity fetched by the data source, the tags are returned as a JSON-encoded string and not a conventional list or a map, owing to a couple of design considerations; which is why one would need to use the Terraform function `jsondecode()`, along with the attribute `entity_tags` in order to convert the JSON-encoded string into a map with key-value pairs. 

The following is an illustration of the aforementioned scenario. It may be observed that a key-value pair version of the JSON-encoded string exported by `entity_tags` is written to the variable `key_value_maps` , using the `jsondecode()` function.

```hcl
data "newrelic_entity" "foo" {
  name = "Sample Searchable Entity"
  domain = "EXT"
  type = "SERVICE_LEVEL"
}

locals {
  key_value_map = { for pair in jsondecode(data.newrelic_entity.foo.entity_tags) : pair.key => pair.values }
}

output "key_value_map" {
  value = local.key_value_map
}
```
The value of `local.key_value_map`  would look like the following.
```hcl
{
  env = ["production"]
  team = ["ops"]
}
```

In order to prevent an error being thrown by the variable added if no tags are associated with the entity (consequently, the value of `entity_tags` is an empty string), the value of the variable may be conditionally defined in order to not throw an error, as shown below.

```hcl
locals {
  key_value_map = data.newrelic_entity.foo.entity_tags != null ? { for pair in jsondecode(data.newrelic_entity.foo.entity_tags) : pair.key => pair.values } : null
}
```

```

@abasha1234 can you please take a look at this change, commit this, and shift this section below all of the sections we have under **Additional Examples**?

Fantastic work on the explanation you've written, but this would be a bit too lengthy/elaborate, and we might want to keep this concise to just help the customer understand they would need to use `jsondecode()` with `entity_tags`; hence these changes :)

abasha1234: Hey @pranav-new-relic thanks for feedback.

pranav-new-relic:
All other failures in integration tests seem to be expected because of latency, except 
- `TestAccNewRelicEntityData_RetrieveSubAccountEntity`
- `TestAccNewRelicEntityData_EntityInSubAccount`
and these aren't working because the entity they're using in the referenced account no longer exists. I've just replenished the entity, and both of these tests shouldn't be failing from the next run.

pranav-new-relic:
Replenished the dead entity _again_; rerunning integration tests

pranav-new-relic (approved):

PR #2697, opened 2024-06-24
chore: make necessary adjustments to support entities package updates

sanderblue:
Requires https://github.com/newrelic/newrelic-client-go/pull/1171

PR #2708, opened 2024-07-03
chore: fix failing tests and more caused by the entities cleanup in client-go

pranav-new-relic:
#2697 ++


pranav-new-relic:
image

the three failing are intermittent failures. Good to go

PR #2727, opened 2024-08-09
feat(synthetics): changes to disallow usage of the legacy runtime with new monitors starting August 26, 2024

pranav-new-relic:
TBD

### DIFF
--- a/newrelic/structures_newrelic_synthetics_all_monitors_validation_helpers.go
+++ b/newrelic/structures_newrelic_synthetics_all_monitors_validation_helpers.go
@@ -1,0 +1,130 @@
+package newrelic
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func validateSyntheticMonitorRuntimeAttributes(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
+	var errorsList []error
+
+	err := validateSyntheticMonitorLegacyRuntimeAttributesUponCreate(d)
+	if err != nil {
+		errorsList = append(errorsList, err...)
+	}
+
+	if len(errorsList) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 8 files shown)


newrelic/structures_newrelic_synthetics_all_monitors_validation_helpers.go
```diff
@@ -0,0 +1,153 @@
...
+	// this would return true only if runtime_type_version is not specified in the configuration at all
+	// and false, if runtime_type_version is specified either as an empty string "", or as any other non nil value (non-empty string)
+	isRuntimeTypeVersionNotSpecifiedInConfiguration := rawConfiguration.GetAttr(SyntheticsRuntimeTypeVersionAttributeLabel).IsNull()
+	_, runtimeTypeVersionInConfig := d.GetChange(SyntheticsRuntimeTypeVersionAttributeLabel)
+
+	// this would return true both when `runtime_type_version` is not specified in the config and when `runtime_type_version` is specified as "" in the config
+	// and false, if `runtime_type_version` has a non nil value (a non-empty string) as its value
+	isRuntimeTypeVersionNullValue := runtimeTypeVersionInConfig == ""
+
+	if !isSyntheticMonitorAlreadyCreated &&
```
pranav-new-relic, line 59:
```suggestion
	if !isSyntheticMonitorCreated &&
```

@pranav-new-relic

pranav-new-relic:
based on latest updates, Synthetics shall be disallowing all requests which do not include runtime attributes in the configuration; so we no longer need to differentiate between the scenario of runtime attributes being absent, and runtime attributes being present, but as empty strings; as the API would not be defaulting monitors created without runtime attributes to the new runtime.

This eliminates the need of this PR, though this shall be useful as a reference, to differentiate between similar complex scenarios as the one stated above, using `d.GetRawConfig()`.

Closing this PR. The new PR which would comprise changes for the Synthetics Legacy Runtime EOL (for the phase on August 26, 2024) would be #2731.

PR #2728, opened 2024-08-16
chore(modules): Adding module to retrieve api keys using a external graphql provider

vinay-newrelic:
# Description
In light of the change to be brought about starting September 25, 2024; existing API Keys can no longer be copied from the UI. So we have designed a module using the sullivtr/graphql provider that can directly be plugged in by the customer and used to fetch the API key.

## Type of change

## Checklist:

## How to test this change?

- Step 1
- Step 2
- etc


examples/modules/sulllivtr-graphql/main.tf
```diff
@@ -0,0 +1,24 @@
...
+}
+
+data "graphql_query" "basic_query" {
+  query_variables = {
+    "id" = var.service.key_id
+    "key_type" = var.service.key_type
+  }
+  query  = file("query.gql")
+}
+
```
pranav-new-relic, line 21: @vinay-newrelic it might be a good idea for us to add the `newrelic_api_access_key` resource also to this module. This module would anyway be used by a customer after they create an API key via the `newrelic_api_access_key` resource (using the ID of the created API key with this module), so it would be a good idea to integrate this resource into the module too.

This would allow customers to be able to actually get the API key directly out of the `newrelic_api_access_key` resource; the only difference it would make to the functioning of this module would be that the query variable "id" would be `newrelic_api_access_key.foo.id` (the ID exported by the resource).

pranav-new-relic:
@vinay-newrelic please rename this module from `sulllivtr-graphql` to `fetch-newrelic-api-access-key`

examples/modules/fetch-newrelic-api-access-key/providers.tf
```diff
@@ -0,0 +1,17 @@
+terraform {
+  required_providers {
+    newrelic = {
+      source = "newrelic/newrelic"
+    }
+    graphql = {
+      source = "sullivtr/graphql"
+       version = "2.5.4"
```
pranav-new-relic, line 8: @vinay-newrelic We can remove this version constraint, as this would not consumers of this module use the latest version of `sullivtr/graphql`

examples/modules/fetch-newrelic-api-access-key/main.tf
```diff
@@ -0,0 +1,41 @@
...
+  name        = "APM ${var.create_access_keys_service.key_type} License Key for ${var.create_access_keys_service.name}"
+  notes       = var.create_access_keys_service.notes
+}
+
+data "graphql_query" "query_with_id" {
+  query_variables = {
+    "id"        = newrelic_api_access_key.api_access_key[0].id
+    "key_type"  = var.create_access_keys_service.key_type
+  }
+  query = file("query.gql")
```
pranav-new-relic, line 33: @vinay-newrelic we seem to be running into a problem with using this module because of this reference of `query.gql`. I see the following error when I try using the sample configuration you've attached in the docs to use this module:

image


Not sure why this is a problem, it could be that modules don't support using files this way unless a path to the file is specified in the call of the module, maybe 🤔 

Anyway; to overcome this issue for now, I would suggest -
- placing the query in a multiline string (using the <<EOF heredoc syntax; a Google search can help explain how this can be used)
- put it inside `terraform.tfvars` as the value of the variable `query_string`
- declare a new variable `query_string` in variables.tf which can then be used as `var.query_string` instead of `file("query.gql")` everywhere else in the `main.tf` references.

Alternatively, you can also add `query_string` as a variable to `variables.tf`, use the multiline (heredoc) syntax to add the default value of the query to the variable (as the value of the argument default) and then use `var.query_string` everywhere `file("query.gql")` is currently being used.

Before you add the GraphQL query to the multiline string, please format it using any online GraphQL formatter and paste it back in the multiline query.

examples/modules/fetch-newrelic-api-access-key/main.tf
```diff
@@ -0,0 +1,41 @@
...
+  query  = file("query.gql")
+  count = local.is_resource_created ? 0 : 1
+}
+
+resource "newrelic_api_access_key" "api_access_key" {
+  count  = var.create_access_keys_service.newrelic_account_id != "" ? 1 : 0
+  account_id  = var.create_access_keys_service.newrelic_account_id
+  key_type    = var.create_access_keys_service.key_type
+  ingest_type = var.create_access_keys_service.ingest_type
+  name        = "APM ${var.create_access_keys_service.key_type} License Key for ${var.create_access_keys_service.name}"
```
pranav-new-relic, line 24: @vinay-newrelic this statement needs a change as the `ingest_type` can be `BROWSER` or `INGEST` and this ingest type would need to be printed accordingly. In addition, it might also be useful to add the ingest type to the output attributes printed.

examples/modules/fetch-newrelic-api-access-key/main.tf
```diff
@@ -0,0 +1,41 @@
...
+    "key_type"  = var.fetch_access_keys_service.key_type
+  }
+  query  = file("query.gql")
+  count = local.is_resource_created ? 0 : 1
+}
+
+resource "newrelic_api_access_key" "api_access_key" {
+  count  = var.create_access_keys_service.newrelic_account_id != "" ? 1 : 0
+  account_id  = var.create_access_keys_service.newrelic_account_id
+  key_type    = var.create_access_keys_service.key_type
```
pranav-new-relic, line 22: @vinay-newrelic upon testing I find that this module works fine to create `INGEST` - `LICENSE` and `INGEST` - `BROWSER` keys but not USER keys. 

image

This is because we need the attribute `user_id` here (coming in as an input from the module) as the attribute `user_id` is required to create a USER API Key. You might want to add `user_id` as an optional attribute in the create module and pass it to the `newrelic_api_access_key` resource.

website/docs/r/api_access_key.html.markdown
```diff
@@ -64,3 +64,76 @@ For example:
...
+    newrelic_account_id        = "12345678"
+    name                       = "key for CLI"
+    key_type                   = "INGEST"
+  }
+}
+
+output "required_attributes" {
+  value = module.create_access_keys.required_attributes
+}
+```
```
pranav-new-relic, line 104: @vinay-newrelic please format all examples used here, using the `terraform fmt` command. 

Additionally 
- please add examples for the three kinds of keys (USER, INGEST LICENSE and INGEST BROWSER)
- we do not need to explain what each argument does, so the Input Variables section is not required
- the `Outputs` section may be updated based on a previous comment of mine, if the `ingest_type` is also included
- with six examples added, since this file would turn out to be too large, I feel it would be a better idea to add all of this content in a `README.md` file in the module (on GitHub) and simply add a link to this in the docs of the `newrelic_api_access_key` resource (in alignment with one of your previous suggestions) :)

pranav-new-relic:
@vinay-newrelic please rename this module to something like "newrelic_api_access_key_extended" instead of the current "fetch-..." name that exists, as we wouldn't want to make it look like the module can only fetch an API Access Key

website/docs/r/api_access_key.html.markdown
```diff
@@ -64,3 +64,76 @@ For example:
 ```
 $ terraform import newrelic_api_access_key.foobar "1234567:INGEST"
 ```
+## Overview
+This module can be used for creation of api keys
```
pranav-new-relic, line 68:
```suggestion
This module may be used to create a user or ingest key using the `newrelic_api_access_key` resource, and fetch the created key, by performing a NerdGraph query under the hood, using the ID of the key created via the resource to fetch the created key.
```

examples/modules/fetch-newrelic-api-access-key/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+    key_type = "XXXX"
+  }
+}
+
+variable "create_access_keys_service" {
+  description = "The service is to create api keys"
+  type = object({
+    api_key                    = string
+    newrelic_account_id        = string
+    name                       = optional(string,"production")
```
pranav-new-relic, line 29:
```suggestion
    name                       = optional(string,"New API Key")
```

examples/modules/fetch-newrelic-api-access-key/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+    key_type                   = string
+    ingest_type                = optional(string,"LICENSE")
+    notes                      = optional(string,"To be used with service XXXX")
```
pranav-new-relic, line 32:
```suggestion
    notes                      = optional(string,"API Key created using the newrelic_api_access_key Terraform resource")
```

examples/modules/fetch-newrelic-api-access-key/main.tf
vinay-newrelic: @pranav-new-relic my bad i missed this key while creating api_access key ..I have now added this an conditional field.So it would be required when key_type == USER and optional in other cases.

website/docs/r/api_access_key.html.markdown
vinay-newrelic: @pranav-new-relic sounds good .Made the necessary changes as per our agreement then

examples/modules/fetch-newrelic-api-access-key/main.tf
vinay-newrelic: @pranav-new-relic i have added ingest type also as an output variable but the ingest type key would ouput null in case of USER keys ...I tried to make it a conditional output but did not find an appropriate way .

pranav-new-relic (approved):

rtizzy:
@pranav-new-relic 

As an NR customer I am pretty appalled not only by the implementation of this provider but by the documentation for it. 

https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/api_access_key

Here is what the docs say

```
[key](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/api_access_key#key-1) - The actual API key. This attribute is masked and not be visible in your terminal, CI, etc.
```

I created an access key with the resource, the docs imply you should be able to fetch it. Nope!

It makes no mention of needing to jump through 5 different hoops to fetch the key I literally just created. 

This is not something TF users care about. This should be **in the provider itself**.

PR #2734, opened 2024-08-22
fix(dashboard): Adding a proper error message to the default values f…

vinay-newrelic:
# Description

if the  element in the default_values list specified by the customer is an empty string "", the provider's unable to decipher this to send to the API. This should not be the case however; a better error should be thrown by the provider stating the right cause. So we need to improve the error handling with this case; 

Fixes # (issue)
Added error handling scenarios with nil and empty checks for default_values and a proper error message for the customer to understand

#### Test Configutation

```hcl
resource "newrelic_one_dashboard" "test_2734" {
  name        = "Sample Dashboard Test PR #2734"
  permissions = "public_read_only"

  page {
    name = "Sample Dashboard Test PR #2734"
    widget_markdown {
      title  = "Sample Dashboard Test PR #2734"
      row    = 7
      column = 1
      width  = 12
      height = 3
      text   = "Sample Dashboard Test PR #2734"
    }

  }

  variable {
    #     default_values = ["", "*"]
    #     default_values = []
    #     default_values = [""]
    #     default_values = ["*"]
    is_multi_selection = true
    item {
      title = "item"
      value = "ITEM"
    }
    name = "variable"
    nrql_query {
      account_ids = [XXXXXX]
      query       = "FROM Transaction SELECT average(duration) FACET appName"
    }
    replacement_strategy = "default"
    title                = "title"
    type                 = "nrql"
  }
}
```

## Type of change

## Checklist:

## How to test this change?

- Step 1
- Step 2
- etc


newrelic/structures_newrelic_one_dashboard.go
```diff
@@ -60,8 +63,22 @@ func expandDashboardVariablesInput(variables []interface{}) []dashboards.Dashboa
...
+					break
+				}
+			}
+			if len(d) > 0 && hasNonNil {
+				variable.DefaultValues = expandVariableDefaultValues(d)
+			} else {
+				return []dashboards.DashboardVariableInput{}, fmt.Errorf("default_values is an empty slice or contains only nil values")
+			}
+		} else {
+			return []dashboards.DashboardVariableInput{}, fmt.Errorf("default_values is either nil, not a slice, or an empty slice")
```
pranav-new-relic, line 81: @vinay-newrelic looks like there are a couple of issues with the code changes made.

https://github.com/user-attachments/assets/6ba46281-d1a7-4f09-91b4-4dfa42454696

- in the first test, if `default_values = [""]`, the error is thrown, which is ✅ 
- in the second test, if `default_values = []`, the error is thrown, which is ✅
- in the third test, if `default_values` does not exist in the `variable{}` block, the error is thrown, which is not supposed to happen, as `default_values` is an optional argument in `variable{}`. This would amount to a breaking change; so we need to make sure this does not happen because of the changes.
- in the fourth test, if `default_values = ["", "*"]` (the `*` can be any other non empty string), a panic is still thrown, so we might need to change the logic accordingly

Please take a look and add changes to make sure these cases are covered :)

vinay-newrelic: @pranav-new-relic I agree upon the third test case and  removing the mandatory check on default values would solve the issue but regarding the fourth test case ...will add **`_has atleast one nill_`** check instead of **_non nil check_** which is being done currently

pranav-new-relic:
@vinay-newrelic everything looks good now, one final set of comments 
- you're probably seeing the lint error that says the cyclomatic complexity of the function has exceeded 21; one quick fix to that would be to split all the functionality dealing with default values (in variables) into a different function, so the if conditions are split into two functions, and the cyclomatic complexity would go down
- of the four test cases we discussed above, if `default_values = []`, we're able to continue to successfully apply without throwing an error with the latest changes. A simpler method to fix this (without handling it in the code) would be to add the condition `MinItems: 1` to the `legend_enabled` attribute in the schema of the resource.

pranav-new-relic (approved):

PR #2740, opened 2024-08-30
fix(synthetics): allow new -> legacy runtime downgrade until Oct 22 for simple browser, scripted monitors

pranav-new-relic:
The "update" functions of resources which deal in Simple Browser, Scripted Browser and Scripted API monitors in our Terraform Provider do not seem to default to an empty runtime, if no runtime attributes are specified in the configuration, or if they are empty strings. This would cause customers to not be able to downgrade monitors from the new runtime to the legacy runtime (though this is not encouraged, given the context around the Legacy Runtime EOL, customers should be able to do this until the final EOL for new and existing monitors on October 22, 2024; hence this change).

This PR includes a tiny change for update functions to default `runtimeType` and `runtimeTypeVersion` to default to empty strings, if runtime attributes are absent in the configuration, or if they have empty strings as values in the configuration. This would allow the request sent to NerdGraph to pick these attributes up with values as empty strings, as [these attributes do not have an `omitempty` against them in the datatype in Go Client](https://github.com/newrelic/newrelic-client-go/blob/ad6a16dea269ab26a057aab1d8b2c5e87e80d0e3/pkg/synthetics/types.go#L1367-L1374) (since they're mandatory on NerdGraph if the parent field `runtime` is specified).

It may be noted that changes have only been made to
- Scripted Browser
- Scripted API

monitors, and not to
- Broken Links
- Cert Check
- Step

monitors, as they already handle this scenario 

https://github.com/newrelic/terraform-provider-newrelic/blob/2a29b5bf5fca2e67f36ad3d6ce937f4732ef9ae0/newrelic/resource_newrelic_synthetics_cert_check_monitor.go#L395-L398

https://github.com/newrelic/terraform-provider-newrelic/blob/d415d1d057a86efc030245674ff508e4dedc32c9/newrelic/structures_newrelic_synthetics_broken_links_monitor.go#L83-L86

https://github.com/newrelic/terraform-provider-newrelic/blob/d415d1d057a86efc030245674ff508e4dedc32c9/newrelic/structures_newrelic_synthetics_step_monitor.go#L92-L95

and also not to the Simple Browser Monitor, owing to a glitch (see the first resolved comment for more).

### DIFF
--- a/newrelic/structures_newrelic_synthetics_script_api_monitor.go
+++ b/newrelic/structures_newrelic_synthetics_script_api_monitor.go
@@ -27,9 +27,8 @@ func buildSyntheticsScriptAPIMonitorInput(d *schema.ResourceData) synthetics.Syn
 	_, runtimeTypeOk := d.GetOk("runtime_type")
 	_, runtimeVersionOk := d.GetOk("runtime_type_version")
 
+	input.Runtime = &synthetics.SyntheticsRuntimeInput{}
 	if scriptLangOk || runtimeTypeOk || runtimeVersionOk {
-		input.Runtime = &synthetics.SyntheticsRuntimeInput{}
-
 		if v, ok := d.GetOk("script_language"); ok {
 			input.Runtime.ScriptLanguage = v.(string)
 		}
@@ -39,6 +38,9 @@ func buildSyntheticsScriptAPIMonitorInput(d *schema.ResourceData) synthetics.Syn
 		if v, ok := d.GetOk("runtime_type_version"); ok {
 			input.Runtime.RuntimeTypeVersion = synthetics.SemVer(v.(string))
 		}
+	} else {
+		input.Runtime.RuntimeType = ""
+		input.Runtime.RuntimeTypeVersion = ""
 	}
 
 	return input
@@ -66,9 +68,8 @@ func buildSyntheticsScriptAPIMonitorUpdateInput(d *schema.ResourceData) syntheti
 	_, runtimeTypeOk := d.GetOk("runtime_type")
 	_, runtimeVersionOk := d.GetOk("runtime_type_version")
 
+	input.Runtime = &synthetics.SyntheticsRuntimeInput{}
 	if scriptLangOk || runtimeTypeOk || runtimeVersionOk {
-		input.Runtime = &synthetics.SyntheticsRuntimeInput{}
-
 		if v, ok := d.GetOk("script_language"); ok {
 			input.Runtime.ScriptLanguage = v.(string)
 		}
@@ -78,6 +79,9 @@ func buildSyntheticsScriptAPIMonitorUpdateInput(d *schema.ResourceData) syntheti
 		if v, ok := d.GetOk("runtime_type_version"); ok {
 			input.Runtime.RuntimeTypeVersion = synthetics.SemVer(v.(string))
 		}
+	} else {
+		input.Runtime.RuntimeType = ""
+		input.Runtime.RuntimeTypeVersion = ""
 	}
 
 	return input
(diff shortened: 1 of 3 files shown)


newrelic/structures_newrelic_synthetics_simple_browser_monitor.go
```diff
@@ -164,6 +165,9 @@ func buildSyntheticsSimpleBrowserMonitorRuntimeAndDeviceEmulationUpdateStruct(d
 		if runtimeTypeVersionOk {
 			simpleBrowserMonitorUpdateInput.Runtime.RuntimeTypeVersion = synthetics.SemVer(runtimeTypeVersion.(string))
 		}
+	} else {
+		simpleBrowserMonitorUpdateInput.Runtime.RuntimeType = ""
+		simpleBrowserMonitorUpdateInput.Runtime.RuntimeTypeVersion = ""
```
pranav-new-relic, line 170: @pranav-new-relic DO NOT add this change to the update function for Simple Browser Monitors. Why?

- Technically, a simple browser monitor cannot exist in the legacy runtime. Today, if such a monitor is created with legacy runtime attributes too from Terraform, a drift is shown, indicating runtime values of the new runtime.
- Because of this behaviour, in the update function, if we provision downgrading from the new runtime to the legacy runtime, the update request accepts runtime values as empty strings and sets them in the UI as well, against the "Legacy Runtime" checkbox, with the `terraform apply` for update. However, because of the point stated above, if the customer runs a `terraform plan` on their configuration after this, they would still see that a drift is shown showing new runtime values; and upon trying to fix the drift and running a `terraform apply`, they would not be able to, as there would be no changes matched against the API :) however, the UI would continue to display the mode as "Legacy Runtime" though the monitor exists in the new runtime.
- To overcome this limitation, it would be ideal to leave this part of the code unchanged (i.e. revert these changes). This would ensure an empty runtime is not populated in the update request when the runtime attributes in the configuration of the customer are empty (or do not exist), thereby keeping the mode of the monitor "New Runtime" in the UI. This would still not fix the problem with the drift, as a drift would be shown asking the customer to move to the values of the new runtime as fetched by the read query, but this would atleast not lead to a false assumption by the customer that the monitor is actually going into the legacy runtime.

newrelic/structures_newrelic_synthetics_script_api_monitor.go
```diff
@@ -78,6 +79,9 @@ func buildSyntheticsScriptAPIMonitorUpdateInput(d *schema.ResourceData) syntheti
 		if v, ok := d.GetOk("runtime_type_version"); ok {
 			input.Runtime.RuntimeTypeVersion = synthetics.SemVer(v.(string))
 		}
+	} else {
+		input.Runtime.RuntimeType = ""
+		input.Runtime.RuntimeTypeVersion = ""
```
pranav-new-relic, line 84: Test - with a created Scripted API Monitor, we can now apply to shift it to the legacy runtime and back to the new runtime too.

Uploading Screen Recording 2024-08-30 at 2.02.36 PM.mov…

newrelic/structures_newrelic_synthetics_script_browser_monitor.go
```diff
@@ -104,6 +105,9 @@ func buildSyntheticsScriptBrowserUpdateInput(d *schema.ResourceData) synthetics.
 		if runtimeTypeVersionOk {
 			input.Runtime.RuntimeTypeVersion = synthetics.SemVer(runtimeTypeVersion.(string))
 		}
+	} else {
+		input.Runtime.RuntimeType = ""
+		input.Runtime.RuntimeTypeVersion = ""
```
pranav-new-relic, line 110: Test - with a Scripted Browser monitor already created, it can be moved from the new runtime to the legacy runtime and vice versa.

https://github.com/user-attachments/assets/f5e083cb-d3f0-436d-9d85-d1242157e1b6

gmanandhar-nr (approved):
lgtm

PR #2743, opened 2024-09-04
feat(synthetics): adds support for multiple browsers and devices with synthetic monitors

shashank-reddy-nr:
### Background
Recently, Synthetics added a new feature for scripted, step, and simple browser monitors to support multiple browsers and devices, allowing users to choose from various options when running monitors. Until now, we only had the option of choosing one device combination. As per customer requests and Synthetics' requirements to support the same multi-browser/devices functionality in Terraform, changes were made in this PR.

### Changes
1: We have added support to choose multiple browsers and devices for scripted, step, and simple browser monitors.
2: Two new attributes, called "browsers" and "devices", of list types, have been added.
3: The integration test cases have been updated to include these two new fields.
4: The documentation has been updated with these two new fields.


newrelic/structures_newrelic_synthetics_all_monitors_validation_helpers.go
```diff
@@ -4,18 +4,29 @@ import (
 	"context"
 	"errors"
 	"fmt"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"strings"
 )
 
-func validateSyntheticMonitorRuntimeAttributes(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
+func validateSyntheticMonitorAttributes(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
```
pranav-new-relic, line 11: thanks for renaming this 👍

newrelic/structures_newrelic_synthetics_simple_browser_monitor.go
```diff
@@ -47,6 +47,12 @@ func buildSyntheticsSimpleBrowserMonitor(d *schema.ResourceData) (synthetics.Syn
 		simpleBrowserMonitorInput.AdvancedOptions.UseTlsValidation = &vs
 	}
 
+	typedBrowsers := expandSyntheticsBrowsers(d.Get("browsers").(*schema.Set).List())
```
pranav-new-relic, line 50: maybe `typedBrowsers` / `typedDevices` might instead be better referred to as `browsersValuesAsList` / `devicesValuesAsList`? Just sharing a thought for us to think about 😄 not to have this changed right away - let's just think about it

shashank-reddy-nr: I feel typedBrowsers is more appropriate, because we are type casting a already list type data to user defined data type.

No Offence 😅
Screenshot 2024-09-10 at 7 59 51 PM

pranav-new-relic:
looks like some integration tests linked to `device_type` and `device_emulation` have gone off because of our `RequiredWith` change - please take a look tomorrow 👍

newrelic/resource_newrelic_synthetics_monitor_test.go
```diff
@@ -30,7 +30,7 @@ func TestAccNewRelicSyntheticsBrowserMonitor_DeviceEmulationError(t *testing.T)
 			// Test: Create
 			{
 				Config:      testAccNewRelicSyntheticsBrowserMonitorConfig_DeviceEmulationError(rName, string(SyntheticsMonitorTypes.BROWSER)),
-				ExpectError: regexp.MustCompile(`you need to specify both 'device_type' and 'device_orientation' fields`),
+				ExpectError: regexp.MustCompile("all of `device_orientation,device_type` must be\nspecified"),
```
pranav-new-relic, line 33: is the `\n` intended 🤔

shashank-reddy-nr: yes, this is intended, due to the error message format is quite weird as shown below
Screenshot 2024-09-11 at 11 16 00 AM

PR #2761, opened 2024-10-23
docs(group):  Documentation of solution to handle race condition in n…

vinay-newrelic:
Documentation of solution to handle race condition in newrelic_group resource

### DIFF
--- a/website/docs/r/group.html.markdown
+++ b/website/docs/r/group.html.markdown
@@ -44,6 +44,28 @@ In addition to the attributes listed above, the following attribute is also expo
 
 ## Additional Examples
 
+### Updating User Group Membership Management in Terraform
+
+### Overview
+There is a potential race condition within Terraform when managing user accounts and their respective group memberships. A user might be deleted before Terraform disassociates them from a user group. This can lead to an error during terraform apply because the user ID no longer exists when the group resource is being updated.
+
+### Recommended Solution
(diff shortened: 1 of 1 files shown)


website/docs/r/group.html.markdown
```diff
@@ -44,6 +44,28 @@ In addition to the attributes listed above, the following attribute is also expo
 
 ## Additional Examples
 
+### Updating User Group Membership Management in Terraform
+
+### Overview
+There is a potential race condition within Terraform when managing user accounts and their respective group memberships. A user might be deleted before Terraform disassociates them from a user group. This can lead to an error during terraform apply because the user ID no longer exists when the group resource is being updated.
```
pranav-new-relic, line 50: "race condition" thank you for giving this condition an apt term :)

pranav-new-relic, line 50:
```suggestion
There is a potential race condition within Terraform when managing user accounts and their respective group memberships. A user might be deleted before Terraform disassociates them from a user group. This can lead to an error during `terraform apply` because the user ID no longer exists when the group resource is being updated.
```

pranav-new-relic (approved):

pranav-new-relic:
beautiful

pranav-new-relic (approved):

vagrawal-newrelic (commented):
approved

Issue #2774, opened 2024-11-13
Resource `newrelic_alert_muting_rule`: Support for Muting Rule "End Behaviour"


jackie-linz:
### Feature Description
Provide ability to change muting rule end behaviour
![image](https://github.com/user-attachments/assets/6942a4ff-afaf-4303-8ab2-a75c9e558b3a)

Aashirwadjain:
Hi @jackie-linz , Thank you for reporting the issue. This feature has been successfully implemented in the latest release of the Terraform provider, version v3.54.0. Hence, Closing this issue.

PR #2783, opened 2024-12-17
feat(alert_muting_rule): Add `action_on_muting_rule_window_ended` attribute in `newrelic_alert_muting_rule` Terraform Resource

Aashirwadjain:
# Description
- The addition of the `action_on_muting_rule_window_ended` attribute in the `newrelic_alert_muting_rule` Terraform resource aims to enhance the functionality of muting rules within New Relic's alerting system using terraform.
- The attribute `action_on_muting_rule_window_ended` is added to MutingRule Schema Resource and handler methods.
- Link to GitHub Issue: https://github.com/newrelic/terraform-provider-newrelic/issues/2774

## Type of change

## Checklist:

### DIFF
--- a/newrelic/resource_newrelic_alert_muting_rule.go
+++ b/newrelic/resource_newrelic_alert_muting_rule.go
@@ -187,6 +187,7 @@ func resourceNewRelicAlertMutingRule() *schema.Resource {
 			"action_on_muting_rule_window_ended": {
 				Type:        schema.TypeString,
 				Optional:    true,
+				Computed:    true,
 				Description: "The action when the muting rule window is ended or disabled.",
 				ValidateFunc: validation.StringInSlice(
 					[]string{
--- a/newrelic/resource_newrelic_alert_muting_rule.go
+++ b/newrelic/resource_newrelic_alert_muting_rule.go
@@ -3,6 +3,7 @@ package newrelic
 import (
 	"context"
 	"fmt"
+	"github.com/newrelic/newrelic-client-go/v2/pkg/alerts"
 	"log"
 	"regexp"
 	"time"
@@ -183,11 +184,17 @@ func resourceNewRelicAlertMutingRule() *schema.Resource {
 				Elem:        scheduleSchema(),
 				Description: "The time window when the MutingRule should actively mute incidents.",
 			},
-			"end_behaviour": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Description:  "The action when the muting rule window is ended or disabled.",
-				ValidateFunc: validation.StringInSlice([]string{"CLOSE_ISSUES_ON_INACTIVE", "DO_NOTHING"}, false),
+			"action_on_muting_rule_window_ended": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The action when the muting rule window is ended or disabled.",
+				ValidateFunc: validation.StringInSlice(
+					[]string{
+						string(alerts.AlertsActionOnMutingRuleWindowEndedTypes.CLOSE_ISSUES_ON_INACTIVE),
+						string(alerts.AlertsActionOnMutingRuleWindowEndedTypes.DO_NOTHING),
+					},
+					false,
+				),
 			},
 		},
 	}
--- a/newrelic/resource_newrelic_alert_muting_rule.go
+++ b/newrelic/resource_newrelic_alert_muting_rule.go
@@ -183,6 +183,12 @@ func resourceNewRelicAlertMutingRule() *schema.Resource {
 				Elem:        scheduleSchema(),
 				Description: "The time window when the MutingRule should actively mute incidents.",
 			},
+			"end_behaviour": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "The action when the muting rule window is ended or disabled.",
+				ValidateFunc: validation.StringInSlice([]string{"CLOSE_ISSUES_ON_INACTIVE", "DO_NOTHING"}, false),
+			},
 		},
 	}
 }
--- a/newrelic/structures_newrelic_alert_muting_rule.go
+++ b/newrelic/structures_newrelic_alert_muting_rule.go
@@ -18,8 +18,8 @@ func expandMutingRuleCreateInput(d *schema.ResourceData) (alerts.MutingRuleCreat
 		Description: d.Get("description").(string),
 	}
 
-	if endBehaviour, ok := d.GetOk("end_behaviour"); ok {
-		createInput.ActionOnMutingRuleWindowEnded = expandMutingRuleEndBehaviour(endBehaviour.(string))
+	if actionOnMutingRuleWindowEnded, ok := d.GetOk("action_on_muting_rule_window_ended"); ok {
+		createInput.ActionOnMutingRuleWindowEnded = alerts.AlertsActionOnMutingRuleWindowEnded(actionOnMutingRuleWindowEnded.(string))
 	}
 
 	if e, ok := d.GetOk("condition"); ok {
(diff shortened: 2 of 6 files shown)


vagrawal-newrelic (approved):
approved

newrelic/resource_newrelic_alert_muting_rule.go
```diff
@@ -183,6 +183,12 @@ func resourceNewRelicAlertMutingRule() *schema.Resource {
 				Elem:        scheduleSchema(),
 				Description: "The time window when the MutingRule should actively mute incidents.",
 			},
+			"end_behaviour": {
```
pranav-new-relic, line 186: @Aashirwadjain 

This is a nice, concise attribute name, but by convention, we need make sure that the name of an attribute in a Terraform resource fully corresponds to what's present in the API, so the customer doesn't find it difficult to understand what `end_behaviour` refers to, while what they are looking for actually is `actionOnMutingRuleWindowEnded`. 

Please rename the attribute accordingly. Ideally, it could be as is in the API though long, `action_on_muting_rule_window_ended`. We have had to add such lengthy attributes at times for API parity in the past too, so this should be okay.

In addition, in cases like these, since you'll now have to rename all references of this attribute, in future (whenever applicable for small-scale changes like these), I would suggest defining attribute names as constants (if there are not too many), similar to this

https://github.com/newrelic/terraform-provider-newrelic/blob/25894e484177876891d4407de299bd4692f20e6e/newrelic/resource_helpers_synthetics.go#L113-L115


and using references to this constant instead of directly writing the string with the attribute name, so it would help in such cases where we'd need to rename attributes. You can add it directly to this resource file, the above is just an example.

newrelic/resource_newrelic_alert_muting_rule.go
```diff
@@ -183,6 +183,12 @@ func resourceNewRelicAlertMutingRule() *schema.Resource {
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "The action when the muting rule window is ended or disabled.",
+				ValidateFunc: validation.StringInSlice([]string{"CLOSE_ISSUES_ON_INACTIVE", "DO_NOTHING"}, false),
+			},
```
pranav-new-relic, line 191:
```suggestion
			"end_behaviour": {
				Type:        schema.TypeString,
				Optional:    true,
				Description: "The action when the muting rule window is ended or disabled.",
				ValidateFunc: validation.StringInSlice(
					[]string{
						string(alerts.AlertsActionOnMutingRuleWindowEndedTypes.CLOSE_ISSUES_ON_INACTIVE),
						string(alerts.AlertsActionOnMutingRuleWindowEndedTypes.DO_NOTHING),
					},
					false,
				),
			},
```

newrelic/structures_newrelic_alert_muting_rule.go
```diff
@@ -33,6 +37,17 @@ func expandMutingRuleCreateInput(d *schema.ResourceData) (alerts.MutingRuleCreat
...
+	switch endBehaviour {
+	case string(alerts.AlertsActionOnMutingRuleWindowEndedTypes.CLOSE_ISSUES_ON_INACTIVE):
+		return alerts.AlertsActionOnMutingRuleWindowEndedTypes.CLOSE_ISSUES_ON_INACTIVE
+	case string(alerts.AlertsActionOnMutingRuleWindowEndedTypes.DO_NOTHING):
+		return alerts.AlertsActionOnMutingRuleWindowEndedTypes.DO_NOTHING
+	default:
+		return alerts.AlertsActionOnMutingRuleWindowEndedTypes.CLOSE_ISSUES_ON_INACTIVE
+	}
+}
+
```
pranav-new-relic, line 50: @Aashirwadjain with the change I've mentioned above to the ValidateFunc, this function would no longer be needed, as the `ValidateFunc` would now make sure a "string" is sent, that too, only with the two valid values as imported from the right type from the `alerts` package in the Go Client.

This function can be discarded, and areas calling this function can be converted as shown below.

**from**
```
	if endBehaviour, ok := d.GetOk("end_behaviour"); ok {
		createInput.ActionOnMutingRuleWindowEnded = expandMutingRuleEndBehaviour(endBehaviour.(string))
	}
```

**to**
```
	if endBehaviour, ok := d.GetOk("end_behaviour"); ok {
		createInput.ActionOnMutingRuleWindowEnded = endBehaviour
	}
```
(remember to change the attribute name as specified in the first comment)

pranav-new-relic:
@Aashirwadjain please add documentation changes here, without which customers would not be aware of the addition of this new attribute.

https://github.com/newrelic/terraform-provider-newrelic/blob/main/website/docs/r/alert_muting_rule.html.markdown?plain=1#L48-L57

pranav-new-relic:
@Aashirwadjain in addition, please add relevant changes to integration tests, to make sure this new attribute is also part of the configuration that is created, read, and updated by the integration test for muting rules. 

Please see the following file, add wherever appropriate. By this, I mean - you don't have to add the attribute to all tests, just make sure the attribute is added wherever needed to make sure atleast one test fully covers this attribute too, along with other attributes of the resource (during create, read, update).

https://github.com/newrelic/terraform-provider-newrelic/blob/25894e484177876891d4407de299bd4692f20e6e/newrelic/resource_newrelic_alert_muting_rule_test.go#L3

pranav-new-relic:
@Aashirwadjain one occurrence that you will also find while testing your changes :) 


https://github.com/user-attachments/assets/f31dd1fd-5d8c-4a90-988b-861e810f362d

- compile your local changes and try using a configuration of the resource `newrelic_alert_muting_rule`, but without the new attribute you've added. Perform a `terraform plan` and `terraform apply` to create a brand new muting rule.
- immediately after the above, run a `terraform plan`. Since no external changes were made in the UI, or in the Terraform configuration, the expectation of this `terraform plan` would be a "no changes" output, but you see a that a drift is being shown, of a change that the customer did not make, with regard to this new attribute.

Think of why this is happening, and what needs to be added/deleted/fixed in the code to prevent this occurrence. Fixing this is important, otherwise this would affect current customers using this Terraform resource without the new attribute (if these changes are merged without the fix, they would see a change upon `terraform plan` they did not make, or have no idea about).

newrelic/resource_newrelic_alert_muting_rule.go
Aashirwadjain: Thanks for reference @pranav-new-relic ,`end_behaviour` is modified as `action_on_muting_rule_window_ended`

newrelic/structures_newrelic_alert_muting_rule.go
Aashirwadjain: Removed `expandMutingRuleEndBehaviour` method and implemented typecasting for `alerts.AlertsActionOnMutingRuleWindowEnded`

Aashirwadjain:
Thanks for suggestions @pranav-new-relic, Following changes are made accordingly: 
- Documentation is updated and added description of `action_on_muting_rule_window_ended` attribute.
- `action_on_muting_rule_window_ended` attribute is added to integration tests.
- `flattenMutingRule` method is updated and added validations.

pranav-new-relic:
image

newrelic/structures_newrelic_alert_muting_rule.go
```diff
@@ -278,10 +278,7 @@ func flattenMutingRule(mutingRule *alerts.MutingRule, d *schema.ResourceData) er
 	configuredCondition := x.([]interface{})
 
 	_ = d.Set("enabled", mutingRule.Enabled)
-
-	if _, ok := d.GetOk("action_on_muting_rule_window_ended"); ok {
-		_ = d.Set("action_on_muting_rule_window_ended", mutingRule.ActionOnMutingRuleWindowEnded)
-	}
+	_ = d.Set("action_on_muting_rule_window_ended", mutingRule.ActionOnMutingRuleWindowEnded)
```
pranav-new-relic, line 281: @vagrawal-newrelic @Aashirwadjain this is nice, but the attribute doesn't need to be `Computed` for this to happen since the read function is anyway being called in create; you can just have this condition. 

The only thing to double check is that since this is coming from an enum in client go, if there is no value returned from the API for this attribute, is an empty string effectively being sent to Terraform from the Go Client via this attribute and saved as "" to the state; if this is the case, we should be okay; and if not, a condition might need to be added accordingly. I don't think this should be the case though.

Please test these changes (for all cases we ideated previously) and let me know if you discover something unusual

website/docs/r/alert_muting_rule.html.markdown
```diff
@@ -54,7 +55,7 @@ The following arguments are supported:
   * `name` - The name of the MutingRule.
   * `description` - The description of the MutingRule.
   * `schedule` - (Optional) Specify a schedule for enabling the MutingRule. See [Schedule](#schedule) below for details
-
+  * `action_on_muting_rule_window_ended` - (Optional) The action when the muting rule window is ended or disabled. Valid values are `CLOSE_ISSUES_ON_INACTIVE`, `DO_NOTHING`
```
pranav-new-relic, line 58:
```suggestion
  * `action_on_muting_rule_window_ended` - (Optional) The action when the muting rule window is ended or disabled. Valid values are `CLOSE_ISSUES_ON_INACTIVE`, `DO_NOTHING`.
```

just a `.` :)

pranav-new-relic (approved):

pranav-new-relic:
we should be good to merge once the requested changes are done, but ⚠️ 
- after the freeze
- after Rohan's PR #2785 is merged
- after the Go Client with these changes https://github.com/newrelic/newrelic-client-go/pull/1259 is released, and `go.mod` in this PR is updated with the corresponding release version of the Go Client containing those changes

website/docs/r/alert_muting_rule.html.markdown
Aashirwadjain: Done

PR #2799, opened 2025-01-30
feat(application_settings) : Revamp application apm settings to sync with the latest features/functionalities in apm settings in ui

vinay-newrelic:
**_Right now, this TF resource only supports a subset of the application settings available on the APM UI .
This PR concludes the changes required for `newrelic_application_settings` resource to add the missing set of configuration settings that are available on the UI_**

#### Test Configuration
```hcl
resource "newrelic_application_settings" "app" {
  guid                   = "Mxxxxxxxxxxxxxxxxxxxx"
  name                   = "XXXXXXXXXXXXXX"
  app_apdex_threshold    = "0.8"
  use_server_side_config = true
  transaction_tracer {
    explain_query_plans {
      query_plan_threshold_type  = "VALUE"
      query_plan_threshold_value = "0.8"
    }
    sql {
      record_sql = "OBFUSCATED"
    }
    stack_trace_threshold_value = "0.3"
    transaction_threshold_type  = "VALUE"
    transaction_threshold_value = "0.9"
  }
  error_collector {
    expected_error_classes = ["StandardError"]
    expected_error_codes   = ["403"]
    ignored_error_classes  = ["HTTPError"]
    ignored_error_codes    = ["407"]
  }

  tracer_type            = "DISTRIBUTED_TRACING"
  enable_thread_profiler = true
  enable_slow_sql        = true
}
```

### DIFF
--- a/newrelic/resource_newrelic_application_settings.go
+++ b/newrelic/resource_newrelic_application_settings.go
@@ -106,7 +106,7 @@ func validateTransactionThresholds(d *schema.ResourceDiff, errorsList *[]string)
 func validateServerSideConfig(d *schema.ResourceDiff, errorsList *[]string) {
 	ServerSideConfig := d.Get("use_server_side_config").(bool)
 
-	attrList := []string{"transaction_tracer", "error_collector", "tracer_type", "enable_thread_profiler"}
+	attrList := []string{"transaction_tracer", "error_collector", "tracer_type", "enable_thread_profiler", "enable_slow_sql"}
 	var finalList []string
 	for _, value := range attrList {
 		_, valueBool := d.GetOk(value)
--- a/newrelic/resource_newrelic_application_settings.go
+++ b/newrelic/resource_newrelic_application_settings.go
@@ -2,13 +2,9 @@ package newrelic
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"github.com/newrelic/newrelic-client-go/v2/pkg/agentapplications"
 	"log"
-	"regexp"
-	"strings"
-
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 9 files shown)


newrelic/resource_newrelic_application_settings.go
```diff
@@ -20,99 +25,317 @@ func resourceNewRelicApplicationSettings() *schema.Resource {
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return resourceNewRelicApplicationSettingsRead(ctx, d, meta)
 }
 
 func resourceNewRelicApplicationSettingsDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	// You can not delete application settings
+	// You can only delete applications and not application settings which might be provisioned by this resource
```
pranav-new-relic, line 338: you can probably return a warning here instead of nil, use the account management resource as an example

newrelic/resource_newrelic_application_settings.go
```diff
@@ -20,99 +25,317 @@ func resourceNewRelicApplicationSettings() *schema.Resource {
...
+			Type:     schema.TypeBool,
+			Computed: true,
+			Default:  nil,
+		},
+	}
+}
+
+func apmApplicationSettingsSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"name": {
```
pranav-new-relic, line 149: looks like we now need to deal with both name and guid; based on reports of traffic we receive from the APM team, we can take call on either excluding name and making guid mandatory (or) making guid not mandatory, and handling the case when either name or guid is supplied (in the latter, we need to use an entitySearch query to fetch the entity)

newrelic/resource_newrelic_application_settings.go
```diff
@@ -20,99 +25,317 @@ func resourceNewRelicApplicationSettings() *schema.Resource {
...
+						Optional:    true,
+						Description: "A list of error codes that should be ignored.",
+					},
+				},
+			},
+		},
+		"tracer_type": {
+			Type:         schema.TypeString,
+			Optional:     true,
+			ValidateFunc: validation.StringInSlice([]string{"CROSS_APPLICATION_TRACER", "DISTRIBUTED_TRACING", "NONE", "OPT_OUT"}, true),
```
pranav-new-relic, line 263: for consideration (not immediate action): we might need to consider if OPT_OUT is valid here, considering it's not displayed in the UI

pranav-new-relic, line 263: we might need to make "NONE" default here or accordingly make a code change later in functions so as to allow "NONE" to be sent, if this attribute is commented out in a subsequent apply

newrelic/resource_newrelic_application_settings.go
```diff
@@ -20,99 +25,317 @@ func resourceNewRelicApplicationSettings() *schema.Resource {
...
 	}
+	return nil
+}
+
+// Custom validator that ensures fields are validated correctly
+func validateApplicationSettingsInput(ctx context.Context, d *schema.ResourceDiff, v interface{}) error {
+	var errorsList []string
+
+	if err := validateRealUserMonitoring(d); err != nil {
+		return err
```
pranav-new-relic, line 115: as discussed, we might need to print errors shown upon using blocks without use_server_side_config = true all at once and not sequentially

newrelic/resource_newrelic_application_settings.go
```diff
@@ -20,99 +25,317 @@ func resourceNewRelicApplicationSettings() *schema.Resource {
...
+					"transaction_threshold_type": {
+						Type:         schema.TypeString,
+						Optional:     true,
+						ValidateFunc: validation.StringInSlice([]string{"APDEX_F", "VALUE"}, false),
+						Description:  "The type of threshold for transaction tracing, either 'APDEX_F' or 'VALUE'.",
+					},
+					"transaction_threshold_value": {
+						Type:        schema.TypeFloat,
+						Optional:    true,
+						Description: "The threshold value for transaction tracing when 'transaction_threshold_type' is 'VALUE'.",
```
pranav-new-relic, line 179: as discussed in the call, one of the problems we're noticing is that if `transaction_threshold_type = VALUE` is sent in the first apply and someone tries commenting this out, the change cannot be propagated to NerdGraph since it expects an enum in the given valid set of values. This would result in a drift after the second terraform apply, suggesting that the change from VALUE -> null never happened.

To avoid this problem, it is essential that we make `transaction_threshold_type`, `transaction_threshold_value` and `stack_trace_threshold_value` mandatory, **if** `transaction_tracer` is specified. In this way, the problem with drift we're seeing can also be avoided in a "logical" fashion, since the smallest of apps, as confirmed from the UI, do not exist without the three of these attributes in Transaction Tracing, if Transaction Tracing is on.

image

newrelic/resource_newrelic_application_settings.go
```diff
@@ -20,99 +25,317 @@ func resourceNewRelicApplicationSettings() *schema.Resource {
...
+									Description:  "The type of threshold for explain plans, either 'APDEX_F' or 'VALUE'.",
+								},
+								"query_plan_threshold_value": {
+									Type:         schema.TypeFloat,
+									Optional:     true,
+									ValidateFunc: validation.FloatAtLeast(0),
+									Description:  "The threshold value for explain plans when 'query_plan_threshold_type' is 'VALUE'.",
+								},
+							},
+						},
```
pranav-new-relic, line 222: same logic as in https://github.com/newrelic/terraform-provider-newrelic/pull/2799/files#r1977356249

query_plan_threshold_type, query_plan_threshold_value : explain_query_plans :: transaction_threshold_type, transaction_threshold_value, stack_trace_threshold_value mandatory : transaction_tracer

newrelic/structures_newrelic_application_settings.go
```diff
@@ -1,55 +1,308 @@
...
+	if _, ok := d.GetOk("app_apdex_threshold"); ok || !isImported {
+		if err = d.Set("app_apdex_threshold", ApmSettings.ApmConfig.ApdexTarget); err != nil {
+			return fmt.Errorf("[DEBUG] Error setting app_apdex_threshold value : %#v", err)
+		}
+	}
+	if _, ok := d.GetOk("use_server_side_config"); ok || !isImported {
+		if err = d.Set("use_server_side_config", ApmSettings.ApmConfig.UseServerSideConfig); err != nil {
+			return fmt.Errorf("[DEBUG] Error setting use_server_side_config value : %#v", err)
+		}
+	}
```
pranav-new-relic, line 185: @vinay-newrelic we might need to remove the debug in these statements (and similar occurrences elsewhere within errors)

newrelic/resource_newrelic_application_settings.go
```diff
@@ -20,99 +25,414 @@ func resourceNewRelicApplicationSettings() *schema.Resource {
...
+		if valueBool && !ServerSideConfig {
+			finalList = append(finalList, value)
 		}
 	}
 
-	if app == nil {
-		return diag.Errorf("the name '%s' does not match any New Relic applications", userApp.Name)
+	if len(finalList) > 0 {
+		*errorsList = append(*errorsList, fmt.Sprintf("use_server_side_config must be set to true when %v is configured", finalList))
+	}
```
pranav-new-relic, line 120: @vinay-newrelic the %v here can probably be a comma separated list using a join function (or using string)

website/docs/r/application_settings.html.markdown
```diff
@@ -14,29 +14,62 @@ a reporting agent.
...
+    transaction_threshold_value = "0.5"
+  }
+  error_collector {
+    expected_error_classes = []
+    expected_error_codes = []
+    ignored_error_classes = []
+    ignored_error_codes = []
+  }
+  tracer_type = "NONE"
+  enable_thread_profiler = true
```
pranav-new-relic, line 46: @vinay-newrelic we might need to add slow_sql as discussed in the meeting

website/docs/r/application_settings.html.markdown
```diff
@@ -14,29 +14,62 @@ a reporting agent.
...
+  * `sql` - (Optional) Configuration block for SQL logging.  Including this block enables SQL logging. The following arguments are supported:
+    * `record_sql` - (Required) The level of SQL recording. Valid values ar `OBFUSCATED`,`OFF`,`RAW` (Mandatory attribute when `sql` block is provided).
+* `error_collector` - (Optional) Configuration block for error collection. Including this block enables the error collector. The following arguments are supported:
+  * `expected_error_classes` - (Optional) A list of expected error classes.
+  * `expected_error_codes` - (Optional) A list of expected error codes(any status code between 100-900).
+  * `ignored_error_classes` - (Optional) A list of ignored error classes.
+  * `ignored_error_codes` - (Optional) A list of ignored error codes(any status code between 100-900).
+* `tracer_type` - (Optional) Configures the type of tracer used. Valid values are `CROSS_APPLICATION_TRACER`, `DISTRIBUTED_TRACING`, `NONE`, `OPT_OUT`.
+* `enable_thread_profiler` - (Optional) Enable or disable the collection of thread profiling data.
```
pranav-new-relic, line 72: @vinay-newrelic we might need to add slow_sql as discussed in the meeting

newrelic/resource_newrelic_application_settings.go
```diff
@@ -20,99 +25,414 @@ func resourceNewRelicApplicationSettings() *schema.Resource {
...
+									Required:     true,
+									ValidateFunc: validation.StringInSlice([]string{"OBFUSCATED", "OFF", "RAW"}, true),
+									Description:  "The level of SQL recording, either 'OBFUSCATED', 'OFF', or 'RAW'.",
+								},
+							},
+						},
+					},
+					"stack_trace_threshold_value": {
+						Type:        schema.TypeFloat,
+						Required:    true,
```
pranav-new-relic, line 230: @vinay-newrelic as discussed in the meeting, let us not have this attribute required here; let's make it optional and instead use the validate functions to throw an error if this attribute is not specified with a transaction_tracer{} block; this is because the InternalValidate error (thrown by Terraform's schema validator) is being thrown first before any of our custom validation, which isn't a good experience when there are multiple other custom errors to be thrown, alongside the constraint on this attribute.

newrelic/resource_newrelic_application_settings.go
```diff
@@ -20,99 +25,414 @@ func resourceNewRelicApplicationSettings() *schema.Resource {
...
+
+	if transactionTypeExists && transactionThresholdType == "APDEX_F" && transactionValueExists {
+		*errorsList = append(*errorsList, "`transaction_threshold_value` should not be set when `transaction_threshold_type` is 'APDEX_F'")
+	}
+}
+
+func validateServerSideConfig(d *schema.ResourceDiff, errorsList *[]string) {
+	ServerSideConfig := d.Get("use_server_side_config").(bool)
+
+	attrList := []string{"transaction_tracer", "error_collector", "tracer_type", "enable_thread_profiler"}
```
pranav-new-relic, line 109: @vinay-newrelic similar to the `enable_slow_sql` condition we've just added in this list; you might need to double check if `enable_slow_sql` needs to be added in similar lists elsewhere too :)

nr-developer-toolkit (approved):

pranav-new-relic (approved):
finally, all good! This was quite an effort to move - fantastic work @vinay-newrelic 👏

PR #2820, opened 2025-02-26
chore: experiment using a different test framework for frequently failing tests

pranav-new-relic:
TBD, DO NOT MERGE

PR #2821, opened 2025-03-02
fix(build): revamp integration test framework via selective product capability tag based run processes

pranav-new-relic:
### Background

Owing to minor gaps found in the functioning of NerdGraph API endpoints in terms of adapting to Terraform resources when simultaneously performing read calls followed by create/update calls - in brief, because of tiny gaps in functionality between resources in the provider and NerdGraph API endpoints which are _extremely insignificant in real-time use cases of Terraform resource consumption by customers_, but are causing an impact in our integration test frameworks with "intermittent"/"occasional" failures, most runs of integration test runs on our PRs seem to end up with failures. While the logical solution to this problem is to identify a method to "increase" the gap between the `Config` and `Check` stages in integration tests, there isn't native support for this in the Terraform Acceptance Test Framework; in addition, adding a sleep method to relevant functions (and similar hacks) seem to occasionally solve the problem, but fail more often than not. This brings about a need to reform our integration test framework, since it is certainly not a good experience for contributors to see integration test checks failing on PRs all the time.

#### Alternate Test Frameworks? 😕 

Owing to the requirement we have and the necessity to be able to customize everything around integration test runs (unlike the regular Terraform Acceptance Test Framework), via #2820, solutions via alternate frameworks such as Terratest, Kitchen-Terraform were explored. However, these come with their own set of problems: most significantly, the downside of extensive customisability coming as the necessity to customize Terraform runs of the configuration too, with one having to specify folder based workspaces for each integration test and setting up test runners for each workspace, which is an obvious gigantic effort for addition and maintenance to move, considering the fact that while it may help serve the purpose we're looking for, it is not a quick solution and all the effort needed is very specific to each resource.

#### Fitting the Current Test Framework into Optimal Run Procedures to Allow Better Resource Allocation and Selective Test Runs 🤔👍  

With all of the above said, with the experience with PRs in mind, the idea has ever since shifted (caused by the unavailability of enough alternatives) from revamping the test framework with newer alternatives, to continuing to use the same test framework, but with better optimization and build procedures in place to selectively run relevant integration tests, allowing for better retries, parallelism (i.e. resource allocation on test runners), and saving time on integration test runs (which, in turn, can be used to more test retries, reduce parallelism, etc).

The idea is that the provider comprises resources across a plethora of product capabilities; while there exist 80+ resources/data sources in total, most of these exist across distinct product lines and can be demarcated on this basis to make sure integration tests associated with only those product mappings found in edited files are run - this would make sure we're running integration tests not just corresponding to the specific files/resources edited, but from the entire product mapping; while also making sure we don't run the entire test suite for changes pertaining to one or two product capabilities.

While this sounds pretty unconventional, as mentioned above, this is a useful approach for us to free up resources to enable better retries and parallelism, allowing for better acting on intermittent failures; while in general too, this would allow someone contributing to code in Alerts and Workflow Integrations "not" be blocked by intermittent test failures in Synthetics, while they will be able to confirm nothing's broken in Alerts and Workflow Integrations is broken, as tests in these product fields would be run. It is important to note that this new approach to integration tests would not affect the notification mechanism the New Relic team has, which would run all integration tests in the suite and report issues, which would especially be useful in the case of an incident; this GitHub workflow would continue to exist as-is, and report failures to Slack, so New Relic can still be notified; the only modification is that such observations do not need to be dependent on PR checks.

##### Working

The entire approach revolves around `-tags` in `gotestsum`, our test runner. While the current implementation has the `test-integration` task run all tests with `-tags=integration`, additional logic needed to be added to determine the "tags" of which corresponding tests would need to be run. To deal with this, 
- an additional module, `integration_test_utiliities` has been added with a bash script and other utilities to identify files changed on the current branch, map them to product capabilities and identify relevant tags, which are then relayed to the `test-integration` make task to run tests corresponding only to these tags
- however, since the determination of tags corresponding to files changed is also something needing quite a lot of logic, varying functionality when run locally and when run via GitHub workflows, and dealing with scenarios in which new files are added/deleted, all of this has also been added to `integration_test_utilities` and wired to `make compile`, to facilitate any changes to product mappings to be handled upon compilation, so as to allow `make test-integration` to pick the right tags and run the right tests. There's a huge bunch of other details, but it'll be too lengthy to talk about it here :)

### Follow-Up Action
- monitor the `test-integration-all` task a.k.a the cronjob to make sure this remains unaffected
- use PRs such as #2836 and #2846 to be 100% sure if the desired functionality of `integration_test_utilities` is seen
- overall, keep tabs on the functioning of integration tests and all GitHub actions after the merge of this PR to make sure nothing is broken by this merge; this would also be useful to determine if we can have greater or lesser retries, parallelism on the integration test runs
- if we confirm everything's fine, add a `README.md` to `integration_test_utilities` to explain usage/local setup etc to contributors (outside New Relic, across other teams in New Relic, etc)


pranav-new-relic:
Nice, keep in mind `2bb6bf6` @pranav-new-relic

PR #2837, opened 2025-03-18
fix(release): fix goreleaser script to update release-notes

vinay-newrelic:
fixed goreleaser script to update release-notes

### DIFF
--- a/build/release.mk
+++ b/build/release.mk
@@ -14,7 +14,7 @@ release-clean:
 
 release-publish: clean tools release-notes
 	@echo "=== $(PROJECT_NAME) === [ release-publish  ]: Publishing release via $(REL_CMD)"
-	$(REL_CMD) --release-notes=$(SRCDIR)/tmp/$(RELEASE_NOTES_FILE)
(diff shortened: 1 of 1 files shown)


pranav-new-relic:
@vinay-newrelic in order to avoid empty releases, let's include this as part of the release to be done with changes to the Azure integrations resource (if that is done earlier, or I'll include it in #2821 which anyway has a bunch of changes to builds etc)

vinay-newrelic:
> @vinay-newrelic in order to avoid empty releases, let's include this as part of the release to be done with changes to the Azure integrations resource (if that is done earlier, or I'll include it in #2821 which anyway has a bunch of changes to builds etc)

@pranav-new-relic I am planning to have it as a separate merge commit so that it becomes easy to revert back or track changes in future and will plan to merge the gcp new update account mutation change([2836](https://github.com/newrelic/terraform-provider-newrelic/pull/2836)) prior to this and release the version after this build pr is merged so that we can avoid empty release and even track it down as a separate merge commit

vagrawal-newrelic (approved):
LGTM

PR #2839, opened 2025-03-20
feat(application_settings): introduce backward compatibility for name

pranav-new-relic:
As we have upgraded application_settings resource to move away from application id to guid in the latest release. We observed a potential bug while handling the backward compatibility of the resource due to application id being assumed as guid and hence invalid guid error is being faced by customers. This fix has already been anticipated owing to technical constraints while preparing for backward compatibility. However we have prioritised to fix this over technical constraints and drafted a potential solution to overcome this . We strictly recommend user/customers to move away from name as a mandatory field to identify application entities and same has been mentioned in the registry documentation.


vagrawal-newrelic (approved):
LGTM

PR #2844, opened 2025-03-31
feat(alerts): adds support for signal_seasonality for NRQL baseline conditions

akane0915:
# Description

- Adds support for `signal_seasonality` for baseline NRQL conditions
- See ticket: https://new-relic.atlassian.net/browse/NR-368788

Related to [Go Client PR](https://github.com/newrelic/newrelic-client-go/pull/1291)

## Type of change

## Checklist:

## How to test this change?

- Try creating a `newrelic_nrql_alert_condition` of type `baseline`.
- Ensure you can set `signal_seasonality` successfully.
- Try creating a `newrelic_nrql_alert_condition` of type `static` without `signal_seasonality`.

E.g.
```
resource "newrelic_nrql_alert_condition" "baseline_condition" {
  account_id                   = 400304
  policy_id                    = 1514915
  type                         = "baseline"
  name                         = "Baseline Condition from Terraform"
  enabled                      = false
  baseline_direction           = "upper_only"
  nrql {
    query = "SELECT average(duration) FROM Transaction"
  }
  critical {
    operator              = "above"
    threshold             = 1
    threshold_duration    = 300
    threshold_occurrences = "all"
  }
  aggregation_window = 60
  aggregation_delay  = 120
  fill_option        = "static"
  fill_value         = 1
  signal_seasonality = "daily"
}
```

### DIFF
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -394,10 +394,19 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Description:  "Seasonality under which a condition's signal(s) are evaluated. Valid values are: 'NEW_RELIC_CALCULATION', 'HOURLY', 'DAILY', 'WEEKLY', or 'NONE'. To have New Relic calculate seasonality automatically, set to 'NEW_RELIC_CALCULATION' (default). To turn off seasonality completely, set to 'NONE'.",
-				ValidateFunc: validation.StringInSlice([]string{"NEW_RELIC_CALCULATION", "HOURLY", "DAILY", "WEEKLY", "NONE"}, true),
+				ValidateFunc: validation.StringInSlice(
+					[]string{
+						string(alerts.NrqlSignalSeasonalities.NewRelicCalculation),
+						string(alerts.NrqlSignalSeasonalities.Hourly),
+						string(alerts.NrqlSignalSeasonalities.Daily),
+						string(alerts.NrqlSignalSeasonalities.Weekly),
+						string(alerts.NrqlSignalSeasonalities.None),
+					},
+					true,
+				),
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
 					// If a value is not provided and the condition uses the default value, don't show a diff. Also case insensitive.
-					return (strings.EqualFold(old, "NEW_RELIC_CALCULATION") && new == "") || strings.EqualFold(old, new)
+					return (strings.EqualFold(old, string(alerts.NrqlSignalSeasonalities.NewRelicCalculation)) && new == "") || strings.EqualFold(old, new)
 				},
 			},
 		},
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -396,7 +396,8 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 				Description:  "Seasonality under which a condition's signal(s) are evaluated. Valid values are: 'NEW_RELIC_CALCULATION', 'HOURLY', 'DAILY', 'WEEKLY', or 'NONE'. To have New Relic calculate seasonality automatically, set to 'NEW_RELIC_CALCULATION' (default). To turn off seasonality completely, set to 'NONE'.",
 				ValidateFunc: validation.StringInSlice([]string{"NEW_RELIC_CALCULATION", "HOURLY", "DAILY", "WEEKLY", "NONE"}, true),
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					return strings.EqualFold(old, new) // Case fold this attribute when diffing
+					// If a value is not provided and the condition uses the default value, don't show a diff. Also case insensitive.
+					return (strings.EqualFold(old, "NEW_RELIC_CALCULATION") && new == "") || strings.EqualFold(old, new)
 				},
 			},
 		},
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -391,9 +391,9 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 				},
 			},
 			"signal_seasonality": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Description:  "Seasonality under which a condition's signal(s) are evaluated. Valid values are: 'NEW_RELIC_CALCULATION', 'HOURLY', 'DAILY', 'WEEKLY', or 'NONE'. To have New Relic calculate seasonality automatically, set to 'NEW_RELIC_CALCULATION' (default). To turn off seasonality completely, set to 'NONE'.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Seasonality under which a condition's signal(s) are evaluated. Valid values are: 'NEW_RELIC_CALCULATION', 'HOURLY', 'DAILY', 'WEEKLY', or 'NONE'. To have New Relic calculate seasonality automatically, set to 'NEW_RELIC_CALCULATION' (default). To turn off seasonality completely, set to 'NONE'.",
 				ValidateFunc: validation.StringInSlice(
 					[]string{
 						string(alerts.NrqlSignalSeasonalities.NewRelicCalculation),
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -389,6 +389,15 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 					return strings.EqualFold(old, new) // Case fold this attribute when diffing
 				},
 			},
+			"signal_seasonality": {
+				Type: 				schema.TypeString,
+				Optional: 		true,
+				Description: 	"Seasonality under which a condition's signal(s) are evaluated. Valid values are: 'HOURLY', 'DAILY', 'WEEKLY', 'NONE', or null. To have New Relic calculate seasonality automatically, set to null (default). To turn off seasonality completely, set to 'NONE'.",
+				ValidateFunc: validation.StringInSlice([]string{"HOURLY", "DAILY", "WEEKLY", "NONE"}, true),
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return strings.EqualFold(old, new) // Case fold this attribute when diffing
+				},
+			},
 		},
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(30 * time.Second),
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -127,6 +127,7 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			StateContext: resourceImportStateWithMetadata(2, "type"),
 		},
+		CustomizeDiff: validateNrqlConditionAttributes,
 		Schema: map[string]*schema.Schema{
 			"policy_id": {
 				Type:        schema.TypeInt,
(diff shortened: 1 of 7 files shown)


newrelic/resource_newrelic_nrql_alert_condition_test.go
```diff
@@ -668,6 +668,39 @@ func TestAccNewRelicNrqlAlertCondition_TitleTemplate(t *testing.T) {
...
+	resourceName := "newrelic_nrql_alert_condition.foo"
+	rName := acctest.RandString(5)
+	signalSeasonality := "daily"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheckEnvVars(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckNewRelicNrqlAlertConditionDestroy,
+		Steps: []resource.TestStep{
+			// Test: Create baselin condition with signal seasonality
```
founddrama, line 681:
```suggestion
			// Test: Create baseline condition with signal seasonality
```

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -99,6 +99,7 @@ The following arguments are supported:
 - `aggregation_timer` - (Optional) How long we wait after each data point arrives to make sure we've processed the whole batch. Use `aggregation_timer` with the `event_timer` method. The timer value can range from 0 seconds to 1200 seconds (20 minutes); the default is 60 seconds. `aggregation_timer` cannot be set with `nrql.evaluation_offset`.
 - `evaluation_delay` - (Optional) How long we wait until the signal starts evaluating. The maximum delay is 7200 seconds (120 minutes).
 - `slide_by` - (Optional) Gathers data in overlapping time windows to smooth the chart line, making it easier to spot trends. The `slide_by` value is specified in seconds and must be smaller than and a factor of the `aggregation_window`.
+- `signal_seasonality` - (Optional) Seasonality under which a condition's signal(s) are evaluated. Only available for baseline conditions. Valid values are: 'HOURLY', 'DAILY', 'WEEKLY', 'NONE', or null. To have New Relic calculate seasonality automatically, set to null. To turn off seasonality completely, set to 'NONE'.
```
founddrama, line 102:
```suggestion
- `signal_seasonality` - (Optional) Seasonality under which a condition's signal(s) are evaluated. Only available for baseline conditions. Valid values are: `HOURLY`, `DAILY`, `WEEKLY`, `NONE`, or `null`. To have New Relic calculate seasonality automatically, set to `null`. To turn off seasonality completely, set to `NONE`.
```

To be consistent with other items, please backtick the explicit values (e.g., see `fill_option` or `aggregation_method`)

newrelic/resource_newrelic_nrql_alert_condition.go
```diff
@@ -396,7 +396,8 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
 				Description:  "Seasonality under which a condition's signal(s) are evaluated. Valid values are: 'NEW_RELIC_CALCULATION', 'HOURLY', 'DAILY', 'WEEKLY', or 'NONE'. To have New Relic calculate seasonality automatically, set to 'NEW_RELIC_CALCULATION' (default). To turn off seasonality completely, set to 'NONE'.",
 				ValidateFunc: validation.StringInSlice([]string{"NEW_RELIC_CALCULATION", "HOURLY", "DAILY", "WEEKLY", "NONE"}, true),
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					return strings.EqualFold(old, new) // Case fold this attribute when diffing
+					// If a value is not provided and the condition uses the default value, don't show a diff. Also case insensitive.
+					return (strings.EqualFold(old, "NEW_RELIC_CALCULATION") && new == "") || strings.EqualFold(old, new)
```
akane0915, line 400: Use case for this is:
- User creates a baseline condition with no signal seasonality specified.
- We default to "NEW_RELIC_CALCULATION" in the backend
- User updates a different field, say `enabled`
- Without this diff suppress function, there would be a diff of `"NEW_RELIC_CALCULATION" ~> null` which wouldn't actually doing anything to the condition but it might confuser the user.

pranav-new-relic:
All is well with the integration tests and lint checks :) (as of the latest commit, https://github.com/newrelic/terraform-provider-newrelic/pull/2844/commits/723e61bd4c883184816011344036499a54840155 which references https://github.com/newrelic/newrelic-client-go/pull/1291/commits/8671fd0bd838a8f3a3b64b297b824d69e82d96b8 in the New Relic Go Client)


https://github.com/user-attachments/assets/392512c9-6004-46d7-b7df-21186dda1fc6

newrelic/resource_newrelic_nrql_alert_condition.go
```diff
@@ -389,6 +390,16 @@ func resourceNewRelicNrqlAlertCondition() *schema.Resource {
...
 			},
+			"signal_seasonality": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Seasonality under which a condition's signal(s) are evaluated. Valid values are: 'NEW_RELIC_CALCULATION', 'HOURLY', 'DAILY', 'WEEKLY', or 'NONE'. To have New Relic calculate seasonality automatically, set to 'NEW_RELIC_CALCULATION' (default). To turn off seasonality completely, set to 'NONE'.",
+				ValidateFunc: validation.StringInSlice([]string{"NEW_RELIC_CALCULATION", "HOURLY", "DAILY", "WEEKLY", "NONE"}, true),
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					// If a value is not provided and the condition uses the default value, don't show a diff. Also case insensitive.
+					return (strings.EqualFold(old, "NEW_RELIC_CALCULATION") && new == "") || strings.EqualFold(old, new)
+				},
```
pranav-new-relic, line 401:
```suggestion
				ValidateFunc: validation.StringInSlice(
					[]string{
						string(alerts.NrqlSignalSeasonalities.NewRelicCalculation),
						string(alerts.NrqlSignalSeasonalities.Hourly),
						string(alerts.NrqlSignalSeasonalities.Daily),
						string(alerts.NrqlSignalSeasonalities.Weekly),
						string(alerts.NrqlSignalSeasonalities.None),
					},
					true,
				),
				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
					// If a value is not provided and the condition uses the default value, don't show a diff. Also case insensitive.
					return (strings.EqualFold(old, string(alerts.NrqlSignalSeasonalities.NewRelicCalculation)) && new == "") || strings.EqualFold(old, new)
				},
```

pranav-new-relic: just wanted to replace hard coded references to each of these strings with references in the client as a good practice so as to facilitate easier modifications to the values corresponding to each of these types (if any) in the future :)

newrelic/resource_newrelic_nrql_alert_condition.go
pranav-new-relic: 👍

newrelic/structures_newrelic_nrql_alert_condition.go
```diff
@@ -72,6 +74,13 @@ func expandNrqlAlertConditionCreateInput(d *schema.ResourceData) (*alerts.NrqlCo
 		}
 	}
 
+	if conditionType == "baseline" {
+		if attr, ok := d.GetOk("signal_seasonality"); ok {
+			seasonality := alerts.NrqlSignalSeasonality(strings.ToUpper(attr.(string)))
+			input.SignalSeasonality = &seasonality
+		}
+	}
+
```
pranav-new-relic, line 83:
```suggestion
	if conditionType == "baseline" {
		if attr, ok := d.GetOk("signal_seasonality"); ok {
			seasonality := alerts.NrqlSignalSeasonality(strings.ToUpper(attr.(string)))
			input.SignalSeasonality = &seasonality
		} else {
			seasonality := alerts.NrqlSignalSeasonalities.NewRelicCalculation
			input.SignalSeasonality = &seasonality
		}
	}
```

I'll explain why I suggest this modification in the following comment (just under this).

newrelic/structures_newrelic_nrql_alert_condition.go
```diff
@@ -132,6 +141,13 @@ func expandNrqlAlertConditionUpdateInput(d *schema.ResourceData) (*alerts.NrqlCo
...
```
pranav-new-relic, line 150:
```suggestion
	if conditionType == "baseline" {
		if attr, ok := d.GetOk("signal_seasonality"); ok {
			seasonality := alerts.NrqlSignalSeasonality(strings.ToUpper(attr.(string)))
			input.SignalSeasonality = &seasonality
		} else {
			seasonality := alerts.NrqlSignalSeasonalities.NewRelicCalculation
			input.SignalSeasonality = &seasonality
		}
	}
```

I'd like to explain why I suggest this change here, and also a similar change in the function handling create (the above comment). While we have handled all discrepancies we've noticed by 
- restoring `omitempty` with the discussed datatypes in the client, 
- introducing a new `NEW_RELIC_CALCULATION` type instead of `null` (which led to the problem with an `omitempty` not existing), and
- handling diff suppression in the provider here to suppress diffs if the API returns `NEW_RELIC_CALCULATION` against `signal_seasonality` but the value in the configuration is `null`,

one of the downsides I'm seeing with the current code in the provider and the client is the fact that because `omitempty` is back, it becomes impossible for customers who could have already set a valid `signal_seasonality` (e.g. `DAILY`) to remove it entirely (in API terms, move it back to `NEW_RELIC_CALCULATION`); while the API does this under the hoodbased on what you shared earlier when there is no `signalSeasonality` in the request, it doesn't seem so since a diff is shown after trying to remove an already configured `signal_seasonality = DAILY` and performing an apply, which is leaving a diff upon the next plan suggesting that `DAILY` was never gone.

We can visit more details on why this happens later, but I believe a simpler solution to this problem (as you've implemented everything else like the diff suppression along the lines of this problem :)) is to add an `else` condition to wherever we're adding `signalSeasonality` in the create/update handlers, to say that if there is a non-null `signal_seasonality` in the state, that is picked and sent to the API, else, `NEW_RELIC_CALCULATION` is sent to the API; by which means "all" baseline conditions created/updated will have a `signalSeasonality` for sure, which will help keep the experience consistent and prevent such drift - this is basically a way of making `NEW_RELIC_CALCULATION` the default `signal_seasonality` on baseline NRQL conditions, unless a different value is specified.

To better understand what I mean, you can try compiling these changes (without my suggestions), try applying an NRQL alert condition with `signal_seasonality = DAILY`, then removing it and trying to apply again; the expected behaviour of the subsequent plan after this would be to see no drift, but we'd see a drift as I've explained above. You can then add these suggestions, and the drift would be seen.

I'd like you to review if these changes sound apt to you and add them accordingly :)

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -180,6 +181,7 @@ resource "newrelic_nrql_alert_condition" "foo" {
 
   # baseline type only
   baseline_direction = "upper_only"
+  signal_seasonality = "none"
```
pranav-new-relic, line 184: how about adding a valid `signal_seasonality` here like `HOURLY` or `DAILY`? @akane0915

akane0915: `NONE` is a valid `signal_seasonality` but I can update it anyways.

newrelic/structures_newrelic_nrql_alert_condition.go
akane0915: This makes sense. I actually did notice this in my testing but I figured that instead of removing the `signalSeasonality` entirely on update, a user could just set it to `NEW_RELIC_CALCULATION` which is effectively the same. I didn't think to do this, and I think this is a good idea!

newrelic/structures_newrelic_nrql_alert_condition.go
```diff
@@ -145,6 +149,11 @@ func expandNrqlAlertConditionUpdateInput(d *schema.ResourceData) (*alerts.NrqlCo
 		if attr, ok := d.GetOk("signal_seasonality"); ok {
 			seasonality := alerts.NrqlSignalSeasonality(strings.ToUpper(attr.(string)))
 			input.SignalSeasonality = &seasonality
+		} else {
+			// Null is equivalent to the default value of `NEW_RELIC_CALCULATION` in the API
+			// so this effectively allows the user to null out the signal seasonality on update
```
pranav-new-relic, line 154: @akane0915 nice, I like the comment here :)

pranav-new-relic (approved):

nr-developer-toolkit (approved):

pranav-new-relic:
Integration tests work fine with the latest changes too. We should be good to proceed as all requested changes have been handled.

PR #2871, opened 2025-05-02
feat(alerts): add silent alerts field to condition terms

gffuentes:
# Description

Adds support for Silent Alerts field (disable_health_status_reporting) for all conditions

Go client side: https://github.com/newrelic/newrelic-client-go/pull/1296

Fixes # https://new-relic.atlassian.net/browse/NR-367381

## Type of change

## Checklist:

## How to test this change?

The feature is has relevant FF on for 100% in staging.

Test by trying to create a baseline condition with the new field:

```
resource "newrelic_nrql_alert_condition" "baseline_condition" {
  account_id                   = 400304
  policy_id                    = 1514915
  type                         = "baseline"
  name                         = "Baseline Condition from Terraform"
  enabled                      = false
  baseline_direction           = "upper_only"
  nrql {
    query = "SELECT average(duration) FROM Transaction"
  }
  critical {
    operator              = "above"
    threshold             = 1
    threshold_duration    = 300
    threshold_occurrences = "all"
    disable_health_status_reporting = true
  }
  aggregation_window = 60
  aggregation_delay  = 120
  fill_option        = "static"
  fill_value         = 1
}
```

### DIFF
--- a/newrelic/resource_newrelic_nrql_alert_condition.go
+++ b/newrelic/resource_newrelic_nrql_alert_condition.go
@@ -98,6 +98,12 @@ func termSchema() *schema.Resource {
 					},
 				},
 			},
+			"disable_health_status_reporting": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Violations will not change system health status for this term.",
+			},
 		},
 	}
 }
--- a/website/docs/r/nrql_alert_condition.html.markdown
+++ b/website/docs/r/nrql_alert_condition.html.markdown
@@ -130,6 +130,7 @@ The `term` block supports the following arguments:
 - `duration` - (Optional) **DEPRECATED:** Use `threshold_duration` instead. The duration of time, in _minutes_, that the threshold must violate for in order to create an incident. Must be within 1-120 (inclusive).
 - `time_function` - (Optional) **DEPRECATED:** Use `threshold_occurrences` instead. The criteria for how many data points must be in violation for the specified threshold duration. Valid values are: `all` or `any`.
 - `prediction` - (Optional) **BETA PREVIEW: the `prediction` field is in limited release and only enabled for preview on a per-account basis.** Use `prediction` to open alerts when your static threshold is predicted to be reached in the future. The `prediction` field is only available for _static_ NRQL alert conditions. See [Prediction](#prediction) below for details.
+- `disable_health_status_reporting` - (Optional) `true` or `false`. Defaults to `false` when field not included in TF config. Violations will not change system health status for this term.
 
 ~> **NOTE:** When a `critical` or `warning` block is added to this resource, using either `duration` or `threshold_duration` (one of the two) is mandatory. Both of these should not be specified.
 
--- a/go.mod
+++ b/go.mod
@@ -2,14 +2,12 @@ module github.com/newrelic/terraform-provider-newrelic/v2
 
 go 1.23.6
 
-replace github.com/newrelic/newrelic-client-go/v2 => github.com/newrelic/newrelic-client-go/v2 v2.58.1-0.20250506230928-b6297b48cc0b
-
 require (
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.26.1
 	github.com/mitchellh/go-homedir v1.1.0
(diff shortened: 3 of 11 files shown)


akane0915 (approved):

newrelic/resource_newrelic_nrql_alert_condition.go
```diff
@@ -98,6 +98,12 @@ func termSchema() *schema.Resource {
 					},
 				},
 			},
+			"disable_health_status_reporting": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
```
pranav-new-relic, line 104: FYI you wouldn't need to change this :) but having `Default: false` with `Optional: true` is more of a semantic thing as `Default: false` should suffice (Terraform would then know that this would possess `false` even if it isn't specified, and if specified, takes the associated value from the configuration)

website/docs/r/nrql_alert_condition.html.markdown
```diff
@@ -130,6 +130,7 @@ The `term` block supports the following arguments:
 - `duration` - (Optional) **DEPRECATED:** Use `threshold_duration` instead. The duration of time, in _minutes_, that the threshold must violate for in order to create an incident. Must be within 1-120 (inclusive).
 - `time_function` - (Optional) **DEPRECATED:** Use `threshold_occurrences` instead. The criteria for how many data points must be in violation for the specified threshold duration. Valid values are: `all` or `any`.
 - `prediction` - (Optional) **BETA PREVIEW: the `prediction` field is in limited release and only enabled for preview on a per-account basis.** Use `prediction` to open alerts when your static threshold is predicted to be reached in the future. The `prediction` field is only available for _static_ NRQL alert conditions. See [Prediction](#prediction) below for details.
+- `disable_health_status_reporting` - (Optional) `true` or `false`. Defaults to `false` when field not included in TF config. Violations will not change system health status for this term.
```
pranav-new-relic, line 133: @gffuentes / @akane0915 while I was going through this line of code added, I just looked up the docs above this and some code - here are some thoughts I wanted to share; this isn't for this PR but general suggestions/feedback etc :) 

I feel like the first few lines under the `## Terms` block is slightly misleading for customers. While we are making clear that `term` is now deprecated and `critical` and/or `warning` needs to be instead, the lines right under this again seem to encourage the usage of `term` as we say
- NRQL alert conditions support up to two terms. At least one `term` must have `priority` set to `critical` and the second optional `term` must have `priority` set to `warning`.
   - we're kinda suggesting the use of `term` with `priority` = `critical` or `warning` as opposed to using `critical` and/or `warning` blocks
- The `term` block supports the following arguments:
   - I think we should, again, make clear here that the following is the schema supported by `critical` and/or `warning` to make sure customers don't end up using `term`. Upon looking up code, I think I understand why this could have been written - the fact that the schema supported by `term` and `critical`/`warning` are the same `termSchema()` could have led to this, but I think we need to make this a bit more clear in the docs to prevent usage of `term`.

pranav-new-relic:
@gffuentes I've just added some comments above - they're more for reference and don't need any changes. With the requested feature flag addition and CI checks run here, we should be all good to go, as tests on Staging (all, associated with the NRQL Alert Condition resource) seem to be running fine 👍

website/docs/r/nrql_alert_condition.html.markdown
gffuentes: haha yeah i see what you mean

I'll think about this a little bit and see if i can improve the wording

pranav-new-relic:
@gffuentes LGTM 👍 we'll only need to wait until the Terraform Provider release issue is unblocked, and we should then be good to go

pranav-new-relic (approved):

pranav-new-relic:
we're seeing a weird lint failure stating that there's a dead link

image

while the "dead link" is actually active https://github.com/newrelic-experimental/nr-terraform-json-dashboard-examples/blob/main/dash_composed.tf

We wouldn't want to hold up the release though, just because of this weird behaviour of the linter; @pranav-new-relic to investigate into this soon before more contributions are received.

✅

pranav-new-relic (approved):

PR #2907, opened 2025-07-10
feat(oci): Add support for OCI linkAccount

msivant:
# Description

Add support for OCI linkAccount and updateAccount

JIRA LINK: https://new-relic.atlassian.net/browse/NR-437888
CLIENT PR: https://github.com/newrelic/newrelic-client-go/pull/1318

## Type of change

## Checklist:

## How to test this change?

- Step 1
- Step 2
- etc

### DIFF
--- a/integration_test_utilities/integration_test_mappings.yaml
+++ b/integration_test_utilities/integration_test_mappings.yaml
@@ -205,6 +205,9 @@ resource_newrelic_cloud_gcp_link_account.go:
 resource_newrelic_cloud_gcp_link_account_test.go:
   test: true
   product_mapping: CLOUD
+resource_newrelic_cloud_oci_link_account.go:
+  test: false
+  product_mapping: CLOUD
 resource_newrelic_data_partition.go:
   test: false
   product_mapping: LOGGING_INTEGRATIONS
--- a/newrelic/provider_newrelic.go
+++ b/newrelic/provider_newrelic.go
@@ -156,6 +156,7 @@ func Provider() *schema.Provider {
 			"newrelic_cloud_azure_integrations":                 resourceNewRelicCloudAzureIntegrations(),
 			"newrelic_cloud_gcp_integrations":                   resourceNewrelicCloudGcpIntegrations(),
 			"newrelic_cloud_gcp_link_account":                   resourceNewRelicCloudGcpLinkAccount(),
+			"newrelic_cloud_oci_link_account":                   resourceNewRelicCloudOciAccountLinkAccount(),
 			"newrelic_data_partition_rule":                      resourceNewRelicDataPartition(),
 			"newrelic_entity_tags":                              resourceNewRelicEntityTags(),
 			"newrelic_events_to_metrics_rule":                   resourceNewRelicEventsToMetricsRule(),
--- a/newrelic/resource_newrelic_cloud_oci_link_account.go
+++ b/newrelic/resource_newrelic_cloud_oci_link_account.go
@@ -1,0 +1,140 @@
+package newrelic
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/newrelic/newrelic-client-go/v2/pkg/cloud"
+)
+
+func resourceNewRelicCloudOciAccountLinkAccount() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceNewRelicCloudOciAccountLinkCreate,
+		ReadContext:   resourceNewRelicCloudOciAccountLinkRead,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
+			"account_id": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Computed:    true,
+				Description: "The New Relic account ID where you want to link the OCI account.",
+				ForceNew:    true,
+			},
+			"tenant_id": {
+				Type:        schema.TypeString,
+				Description: "The OCI tenant identifier.",
+				Required:    true,
+				ForceNew:    true,
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Description: "The name of the linked account.",
+				Required:    true,
+			},
+		},
+	}
+}
+
+func resourceNewRelicCloudOciAccountLinkCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	providerConfig := meta.(*ProviderConfig)
+	client := providerConfig.NewClient
+
+	accountID := selectAccountID(providerConfig, d)
+
+	linkAccountInput := expandOciCloudLinkAccountInput(d)
+
+	var diags diag.Diagnostics
+
+	retryErr := retry.RetryContext(ctx, d.Timeout(schema.TimeoutCreate), func() *retry.RetryError {
+		cloudLinkAccountPayload, err := client.Cloud.CloudLinkAccountWithContext(ctx, accountID, linkAccountInput)
+		if err != nil {
+			return retry.NonRetryableError(err)
+		}
+
+		if len(cloudLinkAccountPayload.Errors) > 0 {
+			for _, err := range cloudLinkAccountPayload.Errors {
+				if strings.Contains(err.Message, "The tenantId you entered does not correspond to your OCI account") {
+					return retry.RetryableError(fmt.Errorf("%s : %s", err.Type, err.Message))
+				}
+				diags = append(diags, diag.Diagnostic{
+					Severity: diag.Error,
+					Summary:  err.Type + " " + err.Message,
+				})
+			}
+		}
+
+		if len(cloudLinkAccountPayload.LinkedAccounts) > 0 {
+			d.SetId(strconv.Itoa(cloudLinkAccountPayload.LinkedAccounts[0].ID))
+		}
+
+		return nil
+	})
+
+	if retryErr != nil {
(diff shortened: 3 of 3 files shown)


examples/modules/cloud-integrations/oci/crossregion.tf
```diff
@@ -0,0 +1,7 @@
+# Create the vault replica in the dynamic destination region
+resource "oci_kms_vault_replication" "vault_replica" {
```
amol-asthana: We dont need this now, right ?
As Shubham confirmed.

examples/modules/cloud-integrations/oci/main.tf
```diff
@@ -0,0 +1,157 @@
...
+data "oci_functions_applications" "existing_functions_apps" {
+  compartment_id = var.compartment_ocid
+}
+
+locals {
+  freeform_tags = {
+    newrelic-terraform = "true"
+  }
+  # Names for the network infra
+  vcn_name        = "newrelic-metrics-vcn"
```
amol-asthana: This should take it from variable? As customer might use existing one.

examples/modules/cloud-integrations/oci/policy.tf
```diff
@@ -0,0 +1,40 @@
...
+  ][0]
+  is_home_region = var.region == local.home_region
+}
+
+
+#Resource for the dynamic group
+resource "oci_identity_dynamic_group" "nr_serviceconnector_group" {
+  count          = local.is_home_region ? 1 : 0
+  compartment_id = var.tenancy_ocid
+  description    = "[DO NOT REMOVE] Dynamic group for service connector"
```
amol-asthana: We should use similar `description` prefix.. everywhere.

examples/modules/cloud-integrations/oci/policy.tf
```diff
@@ -0,0 +1,40 @@
...
+  compartment_id = var.tenancy_ocid
+  description    = "[DO NOT REMOVE] Policy to have any connector hub read from monitoring source and write to a target function"
+  name           = var.newrelic_metrics_policy
+  statements     = [
+    "Allow service keymanagementservice to manage vaults in tenancy",
+    "Allow service keymanagementservice to manage secret-bundles in tenancy",
+    "Allow dynamic-group ${var.dynamic_group_name} to read metrics in tenancy",
+    "Allow dynamic-group ${var.dynamic_group_name} to use fn-function in tenancy",
+    "Allow dynamic-group ${var.dynamic_group_name} to use fn-invocation in tenancy",
+    "Allow dynamic-group ${var.dynamic_group_name} to manage stream-family in tenancy",
```
amol-asthana: why we need these `statements`?

"Allow dynamic-group ${var.dynamic_group_name} to manage stream-family in tenancy",
  "Allow dynamic-group ${var.dynamic_group_name} to manage repos in tenancy",

msivant: I have included these policies as they were part of the ORM approach. After consulting with Shubham, we've determined that it's no longer need and we can remove them.

examples/modules/cloud-integrations/oci/policy-setup/main.tf
```diff
@@ -0,0 +1,146 @@
+resource "oci_identity_compartment" "newrelic_compartment" {
```
ahegdeNR, line 1: @msivant you will have to output compartment OCID and vault OCID so they can be used by metrics and logs module. I don't think they can be re-used if they're not in module outputs.

examples/modules/cloud-integrations/oci/policy-setup/locals.tf
```diff
@@ -0,0 +1,19 @@
...
+    if region.is_home_region
+  ][0]
+  is_home_region = var.region == local.home_region
+
+  freeform_tags = {
+    newrelic-terraform = "true"
+  }
+  newrelic_metrics_access_policy   = contains(split(",", var.policy_stack), "METRICS")
+  newrelic_logs_access_policy      = contains(split(",", var.policy_stack), "LOGS")
+  newRelic_core_integration_policy = contains(split(",", var.policy_stack), "COMMON")
```
ahegdeNR: we don't need to provide option for common for terraform. This was required for ORM in case of update scenario. But for terraform customers use it as module and they won't need this additional check for COMMON

examples/modules/cloud-integrations/oci/policy-setup/main.tf
```diff
@@ -0,0 +1,146 @@
...
+  freeform_tags       = local.freeform_tags
+  timeouts {
+    create = "30m"
+    update = "30m"
+    delete = "30m"
+  }
+}
+
+resource "oci_vault_secret" "ingest_api_key" {
+  count          = local.newRelic_core_integration_policy ? 1 : 0
```
ahegdeNR: all these checks can be removed wherever newRelic_core_integration_policy is used

ahegdeNR (changes requested):

website/docs/guides/cloud_integrations_guide.html.markdown
```diff
@@ -163,3 +164,124 @@ Variables:
...
+  create_vcn         = true
+  function_subnet_id = "" # ignored when create_vcn = true
+
+  # Vault secret OCIDs (dummy)
+  ingest_api_secret_ocid = "ocid1.vaultsecret.oc1..dddddddigingestsecret"
+  user_api_secret_ocid   = "ocid1.vaultsecret.oc1..eeeeeeeeusersecret123"
+
+  # Connector hub payload + configuration
+  payload_link = "https://objectstorage.us-ashburn-1.oraclecloud.com/n/newrelic/b/payloads/o/metrics-payload.json"
+  connector_hubs_data = [
```
ahegdeNR: My guess is this won't work in this format, it needs to be stringified in code. Can you test it? Something like 

```
"[{\"display_name\":\"nr-service-connector-1\",\"description\":\"Service connector for logs from compartment A to New Relic\",\"log_sources\":[{\"compartment_id\":\"ocid1.tenancy.oc1..***\",\"log_group_id\":\"ocid1.loggroup.oc1.iad.***\"}]}]"
```

ahegdeNR (changes requested):

website/docs/guides/cloud_integrations_guide.html.markdown
```diff
@@ -163,3 +164,124 @@ Variables:
...
+  # Networking
+  create_vcn         = true
+  function_subnet_id = "" # ignored when create_vcn = true
+
+  # Vault secret OCIDs (dummy)
+  ingest_api_secret_ocid = "ocid1.vaultsecret.oc1..dddddddigingestsecret"
+  user_api_secret_ocid   = "ocid1.vaultsecret.oc1..eeeeeeeeusersecret123"
+
+  # Connector hub payload + configuration
+  payload_link = "https://objectstorage.us-ashburn-1.oraclecloud.com/n/newrelic/b/payloads/o/metrics-payload.json"
```
ahegdeNR: payload link is not required for terraform

examples/modules/cloud-integrations/oci/policy-setup/locals.tf
msivant: Done!

examples/modules/cloud-integrations/oci/policy-setup/main.tf
msivant: Done!

examples/modules/cloud-integrations/oci/policy-setup/main.tf
msivant: Done!

website/docs/guides/cloud_integrations_guide.html.markdown
msivant: I have removed it at vars and the doc was not updated. Good catch 🙌

website/docs/guides/cloud_integrations_guide.html.markdown
msivant: Done!

website/docs/guides/cloud_integrations_guide.html.markdown
```diff
@@ -163,3 +164,135 @@ Variables:
...
+      "region": "us-ashburn-1"
+    }
+  ]
+  ```
+* `ingest_api_secret_ocid` / `user_api_secret_ocid` – Vault secret OCIDs for ingest and user API keys (avoid embedding plain‑text keys).
+* `newrelic_endpoint` – Logical endpoint selector; the module maps this value to the actual metric ingest URL (use the EU variant for EU accounts).
+* `region` – OCI region key (short code) where resources for this module are created (for example: `iad`, `phx`, `fra`). Provide ONLY the region key, not the full region identifier (so use `iad` instead of `us-ashburn-1`).
+
+> Required ordering: Always apply the `policy-setup` module before the `metrics-integration` module. Only run them together in a single configuration if Terraform can resolve the dependency graph (for example, by referencing outputs from `policy-setup`). Applying metrics integration without the prerequisite policies will result in authorization failures when creating Service Connector Hub resources or invoking functions.
+
```
ahegdeNR: @msivant This comment is not required. The point of modules is that the customers should be able to run multiple modules at once (one after the other in single script). Terraform always follows the order. You can test this out once and remove this note altogether. You anyways have mentioned that policy module is mandatory. That should be sufficient

examples/modules/cloud-integrations/oci/policy-setup/main.tf
```diff
@@ -0,0 +1,149 @@
+resource "oci_identity_compartment" "newrelic_compartment" {
+  compartment_id = var.tenancy_ocid
```
ahegdeNR, line 2: @msivant since we're planning to add vcn as part of this module (not confirmed yet) you can add local.is_home_region check for compartment, vault, and dynamic group creation as well. That way when customers deploy this module multiple times for different region they are not duplicated. Only new vcn gets created per region

website/docs/guides/cloud_integrations_guide.html.markdown
msivant: @ahegdeNR the order is important as the vault keys, policies and linkedAccount are needed to be created in prior

PR #2912, opened 2025-07-14
feat(dashboard): support multiple account id's when creating dashboards

vinay-newrelic:
https://new-relic.atlassian.net/browse/NR-436500

problem context:
After the multiple account query was launched, the backend system in New Relic platform got be able to respond the JSON file supporting the multiple account data. Once, however, newrelic-cli makes a Terraform formatted file, its output doesn't contain the information on the part of multiple account (losing or trimmed this kind of information). It resulted in different output of dashboards or sometimes the series of provisioning doesn't work as expected.

As tentative solution for this, the customer needs to manually create the dashboard. This is very laborious to do the flow and can potentially include some kind of human error.


pranav-new-relic:
@vinay-newrelic some thoughts:
1. I think we need to modify the existing docs page of the `newrelic_one_dashboard` resource (wherever there is `account_id` ) to make clear in some subsection that this now supports multiple account IDs - maybe just a one line detail or a brief example, and we can redirect them to this other page you've written. 
   - I guess some of the `account_id`s in the examples can also be updated to reflect this, in the existing docs page of the `newrelic_one_dashboard` resource
2. Have you tested the scenario with drift with multiple account IDs in `jsonencode()`  - upon first performing a `terraform apply` with account IDs `[A1, A2]` and changing them to `[A1, A3]` in the UI, is the drift shown effectively (as expected)? Some more scenarios I've been thinking about -
   - Along the same lines, a more important concern would be if a `terraform apply` is done with account IDs `[A1, A2]` and these are reduced to `[A2]` in the UI, as I think the logic handler would then see this as `accountId`  and not `accountIDs` (in the API) which might affect the way we're treating this
   - Similarly, is this working effectively if a `terraform apply` is performed with `account_id = A1` and then in the UI, if `[A1, A2]` are selected?
   - What does the logic do if `account_id = jsonencode([A1])` is used instead of `account_id = A1`? Can this lead to any unexpected behaviour?
   - If you've already tested multiple permutations and combinations along these lines, we should be totally fine to proceed :)
3. The changes in `testing.tf` need to be reverted
4.  Tiny nitpick: In the schema against `account_id`, can you just update the `Description:` field to reflect that single or multiple account IDs can be specified? The reason I say this is that a lot of customers rely on IDE extensions for Terraform development which explain attributes to the customer based on the `Description:` field.


Especially if (2) is catered to (and we're sure everything's working perfectly as intended), we should be good to proceed 👍

newrelic/resource_newrelic_one_dashboard.go
```diff
@@ -522,9 +522,9 @@ func dashboardWidgetNRQLQuerySchemaElem() *schema.Resource {
 	return &schema.Resource{
 		Schema: map[string]*schema.Schema{
 			"account_id": {
-				Type:        schema.TypeString,
-				Optional:    true,
-				Computed:    true,
+				Type:     schema.TypeString,
+				Optional: true,
+				//Computed:    true,
```
pranav-new-relic: 🤔 wondering if this is needed

PR #2916, opened 2025-08-06
fix(dashboard): add billboard settings support in newrelic_one_dashboard resource

Aashirwadjain:
## Description
This PR aims to add billboard settings support in newrelic_one_dashboard resource.

## Type of change

## Checklist:

## Tests
- Create Resource 
  image
- Update Resource
   image
- Delete Resource
   image


pranav-new-relic (commented):
Nice work @Aashirwadjain - I've tested this multifold myself and don't seem to land in any unexpected drift; so this is great. One final bunch of changes we should still make, though:
- The `link{}`, `visual{}` and `grid_options{}` blocks in `billboard_settings{}` can be empty but can still result in a successful `terraform apply` - we're not enforcing the condition of each of these blocks containing atleast one of the optional attributes. We could have enforced this directly via `AtleastOneOf`, but there are other attributes which have similar names that could lead to a conflict, so we would need to add logic in the `validateDashboard*` function to prevent customers from keeping these blocks empty and applying configuration. Same goes with the `billboard_settings{}` block too - if it is just empty, it should throw an error suggesting that it should comprise one of `link{}`, `visual{}`, `grid_options{}`.
- Merge conflicts need to be resolved (also, point to the latest Go Client version)
- An extra tiny nitpick (not necessarily now): looks like the schema of `billboardSettings.visual.alignment` includes "auto" | "stacked" | "inline", but we're allowing for the inclusion of "stacked" and "inline" - I think this makes sense (since the other value would be similar to no value, I guess) but it might be helpful to double check with the Dashboards UI team to know if this is to be considered a valid value.

pranav-new-relic:
Let me know once done, and we'll merge and release

pranav-new-relic:
@Aashirwadjain let's skip (1) as discussed in the call to account for the timeline to release this today - let's add this as a fast-follow task.

vagrawal-newrelic (approved):
LGTM

PR #2930, opened 2025-09-08
fix(newrelic_one_dashboard): add plan-time validation for `account_ids` in `variable{}` block

Aashirwadjain:
## Description

This PR aims to add a plan-time validation for the `account_ids` attribute inside the `nrql_query{}` for `variable{}` block. It ensures that `terraform plan` will fail if the attribute is either omitted or set to an empty list (`[]`).

## Type of change

## Checklist:

### DIFF
--- a/newrelic/resource_newrelic_one_dashboard.go
+++ b/newrelic/resource_newrelic_one_dashboard.go
@@ -145,6 +145,7 @@ func dashboardVariableSchemaElem() *schema.Resource {
 						"account_ids": {
 							Type:        schema.TypeList,
 							Optional:    true,
+							Computed:    true,
 							Description: "New Relic account ID(s) to issue the query against.",
 							Elem: &schema.Schema{
 								Type: schema.TypeInt,
--- a/newrelic/resource_newrelic_one_dashboard.go
+++ b/newrelic/resource_newrelic_one_dashboard.go
@@ -145,7 +145,6 @@ func dashboardVariableSchemaElem() *schema.Resource {
 						"account_ids": {
 							Type:        schema.TypeList,
 							Optional:    true,
-							Computed:    true,
 							Description: "New Relic account ID(s) to issue the query against.",
 							Elem: &schema.Schema{
 								Type: schema.TypeInt,
--- a/newrelic/structures_newrelic_one_dashboard.go
+++ b/newrelic/structures_newrelic_one_dashboard.go
@@ -213,8 +213,7 @@ func expandVariableNRQLQuery(in []interface{}, meta interface{}) *dashboards.Das
 
 		// Default to provider account ID if not set or empty
 		if !hasAccountIDs {
-			log.Printf("[DEBUG] expandVariableNRQLQuery: account_ids not set or empty, attempting to default from provider...")
-			
+			log.Printf("[DEBUG] expandVariableNRQLQuery: account_ids not set or empty, attempting to default from provider.")
 			if metaMap, ok := meta.(map[string]interface{}); ok {
 				if acct, exists := metaMap["account_id"]; exists {
 					accountIDs = []int{acct.(int)}
--- a/newrelic/structures_newrelic_one_dashboard.go
+++ b/newrelic/structures_newrelic_one_dashboard.go
@@ -139,7 +139,7 @@ func expandDashboardVariablesInput(variables []interface{}, meta interface{}) ([
 		}
 
 		if q, ok := v["nrql_query"]; ok && len(q.([]interface{})) > 0 {
-			variable.NRQLQuery = expandVariableNRQLQuery(q.([]interface{}), meta)
+			variable.NRQLQuery = expandVariableNRQLQuery(q.([]interface{}))
 		}
 
 		if r, ok := v["replacement_strategy"]; ok {
@@ -190,7 +190,7 @@ func expandVariableItems(in []interface{}) []dashboards.DashboardVariableEnumIte
 	return out
 }
 
-func expandVariableNRQLQuery(in []interface{}, meta interface{}) *dashboards.DashboardVariableNRQLQueryInput {
+func expandVariableNRQLQuery(in []interface{}) *dashboards.DashboardVariableNRQLQueryInput {
 
 	var out dashboards.DashboardVariableNRQLQueryInput
 
@@ -203,42 +203,6 @@ func expandVariableNRQLQuery(in []interface{}, meta interface{}) *dashboards.Das
 
 	return &out
 
-	//var out dashboards.DashboardVariableNRQLQueryInput
-	//
-	//for _, v := range in {
-	//	cfg := v.(map[string]interface{})
-	//
-	//	var accountIDs []int
-	//	rawAccountIDs, hasAccountIDs := cfg["account_ids"]
-	//
-	//	// Check if account_ids is explicitly set AND non-empty
-	//	if hasAccountIDs && rawAccountIDs != nil {
-	//		accountIDsSlice := rawAccountIDs.([]interface{})
-	//		if len(accountIDsSlice) > 0 {
-	//			// Non-empty slice, use as-is
-	//			accountIDs = expandVariableAccountIDs(accountIDsSlice)
-	//		} else {
-	//			// Empty slice - treat as omitted for defaulting purposes
-	//			hasAccountIDs = false
-	//		}
-	//	}
-	//
-	//	// Default to provider account ID if not set or empty
-	//	if !hasAccountIDs {
-	//		log.Printf("[DEBUG] expandVariableNRQLQuery: account_ids not set or empty, attempting to default from provider.")
-	//		if metaMap, ok := meta.(map[string]interface{}); ok {
-	//			if acct, exists := metaMap["account_id"]; exists {
-	//				accountIDs = []int{acct.(int)}
-	//			}
-	//		}
-	//	}
-	//
-	//	out = dashboards.DashboardVariableNRQLQueryInput{
-	//		AccountIDs: accountIDs,
-	//		Query:      nrdb.NRQL(cfg["query"].(string)),
-	//	}
-	//}
-	//return &out
 }
 
 func expandVariableAccountIDs(in []interface{}) []int {
--- a/newrelic/structures_newrelic_one_dashboard.go
+++ b/newrelic/structures_newrelic_one_dashboard.go
@@ -37,7 +37,7 @@ func expandDashboardInput(d *schema.ResourceData, meta interface{}, dashboardNam
 		return nil, err
 	}
 
-	dash.Variables, err = expandDashboardVariablesInput(d.Get("variable").([]interface{}), meta)
+	dash.Variables, err = expandDashboardVariablesInput(d.Get("variable").([]interface{}))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


newrelic/structures_newrelic_one_dashboard.go
```diff
@@ -190,16 +190,42 @@ func expandVariableItems(in []interface{}) []dashboards.DashboardVariableEnumIte
...
+				hasAccountIDs = false
+			}
+		}
+
+		// Default to provider account ID if not set or empty
+		if !hasAccountIDs {
+			log.Printf("[DEBUG] expandVariableNRQLQuery: account_ids not set or empty, attempting to default from provider.")
+			if metaMap, ok := meta.(map[string]interface{}); ok {
+				if acct, exists := metaMap["account_id"]; exists {
+					accountIDs = []int{acct.(int)}
```
pranav-new-relic: @Aashirwadjain a tiny nitpick here: a cleaner way to do this would be by typecasting `meta` into the type `*ProviderConfig` which is the case with a lot of resources, but I just realised this resource has many usages of this kind where `meta` is being typecasted into `map[string]interface{}`. We can go with the former instead

Have a few other things - we'll discuss this in office 👍

pranav-new-relic:
we'll also need documentation changes as stated in the ticket, since the definition of `nrql_query` points to the same section in variable blocks and widget blocks, though they're not supposed to be the same

newrelic/structures_newrelic_one_dashboard.go
Aashirwadjain: Hi @pranav-new-relic, the current implementation is based on the existing design. The `*ProviderConfig` implementation that you might have observed are suitable for data source or resource files. 

The key difference is that data sources and resources receive different types for the meta parameter:
- Data sources: Receive meta as `*ProviderConfig` directly.
- Resources: Receive meta that gets processed into `map[string]interface{}` before being passed to structure functions.

Hence, the data source can directly cast to `*ProviderConfig` which is not the case with structures.

Aashirwadjain:
Updated the documentation, we now have different sections for both widget and variable query blocks.

newrelic/structures_newrelic_one_dashboard.go
pranav-new-relic: Not necessarily, this does not appear to be the right reason - though I do get your point. The real reason is

https://github.com/newrelic/terraform-provider-newrelic/blob/aea0d38b10441340baa053150f4bb79c0f9b49ab/newrelic/resource_newrelic_one_dashboard.go#L830-L836

which is then sent as `meta` into the expand functions in the structures file (and that's not how "all" resources do it). There's no design paradigm that says it has to be done this way (and this shouldn't be needed too, why exclude all other "meta" details), but now that we have this already in the resource - I don't think it would be a good use of time to edit all of it and continue this conversation on such a tiny nitpick 😄, so I see what you're saying 👍

website/docs/r/one_dashboard.html.markdown
```diff
@@ -249,6 +249,21 @@ EOT
...
+|------------|----------------|--------|---------|
+| **Widget** `nrql_query` | `account_id` | String/Number or JSON-encoded array | `12345` or `jsonencode([12345, 67890])` |
+| **Variable** `nrql_query` | `account_ids` | List of integers | `[12345, 67890]` |
+
+**Historical Context:** This difference exists because:
+- **Variables** were designed with multi-account support from the beginning (v3.9.0+).
+- **Widgets** originally supported only single accounts, with multi-account support added later (v3.65.0+) using JSON encoding for backward compatibility.
+
+Both blocks default to the provider's configured account ID when the account field is omitted or empty.
+
```
pranav-new-relic: this is nice, let's have this at the bottom above "Additional Examples"

pranav-new-relic: Also, since this is a useful section, after having shifted this above Additional Examples, let's add a sentence in the 
- Nested `nrql_query` blocks (for Widgets)
- Nested `nrql_query` blocks (for Variables)
sections at the bottom of each along the lines of "this is not to be confused with `nrql_query` used with variables - for more details on the differences between the two, please see the "..." section" in the widgets block and vice versa for variables

website/docs/r/one_dashboard.html.markdown
```diff
@@ -249,6 +249,21 @@ EOT
 ```
 See additional [examples](#additional-examples).
 
+## Important: NRQL Query Schema Differences
```
pranav-new-relic:
```suggestion
### Important: `nrql_query` Blocks: Schema Differences
```

pranav-new-relic:
Also, we do not seem to have this case covered by a new test or an existing test modified. Either of these could have helped identify this edge case much in prior. Please ensure this change is covered by a test case as well. This is something we would want to do by default when a functional change is made to an attribute in any resource in the Terraform Provider.

website/docs/r/one_dashboard.html.markdown
Aashirwadjain: Updated.

website/docs/r/one_dashboard.html.markdown
Aashirwadjain: Updated.

newrelic/structures_newrelic_one_dashboard.go
Aashirwadjain: Yes, that's what I was trying to explain above, thanks for confirming. 👍

Aashirwadjain:
Thanks for the review @pranav-new-relic. My initial push was focused primarily on validating the core implementation approach before building out the full test suite. Now that the approach is confirmed, I have added the test case covering the update scenario you described, which confirms the fix for the drift.

newrelic/resource_newrelic_one_dashboard.go
```diff
@@ -145,6 +145,7 @@ func dashboardVariableSchemaElem() *schema.Resource {
 						"account_ids": {
 							Type:        schema.TypeList,
 							Optional:    true,
+							Computed:    true,
```
pranav-new-relic: @Aashirwadjain setting this to Computed would imply that the attribute shall be controlled internally, meaning no drift would be shown when, say, the Account ID against the NRQL Query of the Variable in the Dashboard is changed. This, again, is undesired - as we would not want the diff to _not_ be shown in the case of real changes being made. Again, a simple test would help reveal this behaviour - create a dashboard with an empty `account_ids`, go to the UI of the dashboard and change the Account ID against the NRQL query in the variable, and run a `terraform plan` here again; the expectation would be to see a drift but you'll see none.

This calls for more changes needed elsewhere in the code, landing us in a tricky situation. We will need to ensure that defaulting works in a way that no drift is shown when no "real" change is made (default account ID -> empty list or vice versa), while also ensuring that the changes for this do not end up _not_ displaying real changes (e.g. when the account ID upstream is changed). Does this need diffsuppression? Or more changes in the flatten functions in read? Or elswhere? Take your time and think about it, let's push this only when this edge case is fixed to not affect customer experience 👍 

I'm guessing for sure there are other attributes/instances which seem to encounter this defect and will need correction (and not just this), but the solution you'll figure out can help the case with the rest.

pranav-new-relic: Also, the UI test I've talked about above, if done with the example in the current docs page https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/one_dashboard would help reveal that when you go to the UI of the dashboard and edit the variable, it would display that the NRQL query is invalid (though the API did accept the request when creating this from Terraform). However, to not let customers into this experience, let's change the NRQL query in the example here to a more meaningful one that doesn't throw this error

website/docs/r/one_dashboard.html.markdown
```diff
@@ -521,13 +526,81 @@ The following arguments are supported:
...
     * `ignore_time_range` - (Optional) An argument with a boolean value that is supported only by variables of `type` _nrql_ - when true, the time range specified in the query will override the time picker on dashboards and other pages.
     * `excluded` - (Optional) An argument with a boolean value. With this turned on, the query condition defined with the variable will not be included in the query. Defaults to `false`.
+
+### Nested `nrql_query` blocks (for Variables)
+
+Nested `nrql_query` blocks in **variable** allow you to make NRQL queries to populate dashboard variables with dynamic values from one or more accounts.
+
+The following arguments are supported:
+
+  * `account_ids` - (Optional) List of account IDs such as `[12345, 67890]`. If omitted or set to `[]` (empty list), defaults to the provider's configured account ID.
```
Aashirwadjain: Hi @pranav-new-relic, Should we label this attribute as (Required) in the documentation?

Here's the situation:
- Technically, in the schema, the attribute is still marked as `Optional: true`.
- However, the new validation function makes it functionally required – a terraform plan will fail if the user omits it or provides an empty list.

Feel free to provide any suggestions on the best way to word it.

pranav-new-relic: Good point :) let's have it labeled "Required" though (technically, we're just custom-validation what would've otherwise been validated by the provider if labeled Required, so we can do so)

newrelic/structures_newrelic_one_dashboard.go
```diff
@@ -1862,6 +1905,54 @@ func validateDashboardVariableOptions(d *schema.ResourceDiff) error {
...
+
+			// Check if account_ids is missing or empty
+			if !accountIDsOk {
+				return fmt.Errorf("variable[%d]: account_ids is required for NRQL variables. Default provider account ID is %d - consider setting account_ids = [%d]",
+					i, defaultAccountID, defaultAccountID)
+			}
+
+			if accountIDsSlice, ok := accountIDs.([]interface{}); ok && len(accountIDsSlice) == 0 {
+				return fmt.Errorf("variable[%d]: account_ids cannot be empty for NRQL variables. Default provider account ID is %d - consider setting account_ids = [%d]",
+					i, defaultAccountID, defaultAccountID)
```
pranav-new-relic, line 1910: Since we're returning the error directly when a case is found, this would be a problem when the customer has more than one variable blocks violating the condition; and only the first block violating the condition would lead to the error being thrown; so we'll need to collect the errors and return them. Hint: "..."

A friendly reminder again - please test in multiple P&Cs to unearth scenarios like these, or more :) 

Also try adding variable blocks with no contents, variable blocks with `nrql_query` but no contents, etc etc; combinations of a valid `nrql_query` block in a variable and and another invalid block - these would not only make sure the error is returned as expected, but will also expose any loopholes we might have in the error handling / interface expansion logic here which we can then fix.

newrelic/structures_newrelic_one_dashboard.go
```diff
@@ -201,6 +202,43 @@ func expandVariableNRQLQuery(in []interface{}) *dashboards.DashboardVariableNRQL
...
+	//			}
+	//		}
+	//	}
+	//
+	//	out = dashboards.DashboardVariableNRQLQueryInput{
+	//		AccountIDs: accountIDs,
+	//		Query:      nrdb.NRQL(cfg["query"].(string)),
+	//	}
+	//}
+	//return &out
```
pranav-new-relic: we can get rid of all this and the extra meta arguments in expand functions (like you just mentioned on DM) - can be rewritten and thought of afresh when needed

Aashirwadjain:
@pranav-new-relic  The `data_format{}` block references are fixed in documentation as mentioned in #2937.

newrelic/structures_newrelic_one_dashboard.go
Aashirwadjain: Updated.

website/docs/r/one_dashboard.html.markdown
Aashirwadjain: Sure, I updated it to Required.

newrelic/structures_newrelic_one_dashboard.go
Aashirwadjain: Good point, thanks for the suggestion. For now, I'd prefer to stick with the current approach. It keeps the error output clean and also prevents overwhelming users with a potentially long list of errors. This is also consistent with how we handle similar validation elsewhere in the codebase.

pranav-new-relic: 👍

PR #2816, opened 2025-02-20
feat(alerts): Add prediction support to NRQL alert conditions

founddrama:
# Description

- **Adds support for `prediction` fields on NRQL alert conditions**
- `prediction` field is only available for static NRQL conditions (not baseline/anomaly conditions)
- **Current state of the feature:** released internally and open to limited preview (public preview coming shortly)
- Depends on https://github.com/newrelic/newrelic-client-go/pull/1271

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist:

- [x] My commit message follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] My code is formatted to [Go standards](https://go.dev/blog/gofmt)
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes. Go [here](https://github.com/newrelic/terraform-provider-newrelic/blob/main/CONTRIBUTING.md#testing) for instructions on running tests locally.

## How to test this change?

For an account that is current added to the feature flag:

- Create a static NRQL alert condition that uses the `prediction` field
- Should also be able to update that condition
- Should not interfere with the CRUD operations of NRQL alert conditions that do NOT have the `prediction` field
- (FWIW: we've been testing this internally since December)

PR #2950, opened 2025-10-10
feat(alerts): add support for target entity in NRQL conditions

wulczer:
The target entity feature for NRQL conditions is still in beta, but I'd like to start doing internal testing on it.

Correspoding `newrelic-client-go` PR: newrelic/newrelic-client-go#1337

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/newrelic/go-agent/v3 v3.30.0
 	github.com/newrelic/go-insights v1.0.3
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/newrelic/terraform-provider-newrelic/v3
 
 go 1.23.6
 
+replace github.com/newrelic/newrelic-client-go/v2 => github.com/newrelic/newrelic-client-go/v2 v2.70.3-0.20251010092902-bb23128e2200
```
founddrama: 🤔 I believe you want to update the version in the `require` block, not add a `replace`

Example: #2816

PR #2959, opened 2025-10-17
feat(alerts): add support for target entity in NRQL conditions

wulczer:
This is the second attempt at this, now that the backend is ready for these changes (in particular, the `targetEntity` field is now visible regardless of opt-in status, so conditions can be fetched even if the account doesn't have access to the mutation field.

The previous attempt was in #2950.


pranav-new-relic (approved):
LGTM

Issue #2956, opened 2025-10-16
Unable to create an account


aprohorov-callsign:
Hey, 

I'm using the provider for account management for more than 1 year, but suddenly it's stop working.

Here mine simple example:
```
terraform {
  required_version = "~> 1.0"
  required_providers {
    newrelic = {
      source = "newrelic/newrelic"
    }
  }
}

variable "newrelic_api_key" {
  description = "NewRelic API key."
  type        = string
  sensitive   = true
}

variable "newrelic_account" {
  description = "NewRelic account."
  type        = string
}

provider "newrelic" {
  account_id = var.newrelic_account
  region     = "EU"
  api_key    = var.newrelic_api_key
}

resource "newrelic_account_management" "this" {
  name   = "test account"
  region = "eu01"
}
```

But when I'm trying to apply it, it just stuck on creation.

Terraform version: 1.13.3

PR #2974, opened 2025-11-20
fix(accountmanagement): revise working approach to manage accounts, improve read efficiency

pranav-new-relic:
# Description

Improves the working approach of the Terraform resource for New Relic account management with efficient read operations using targeted API calls.

Of late (e.g. #2956), we've been hearing few reports from customers who have been failing to create accounts via the `newrelic_account_management` resource, especially those in the EU region - and the root cause appears to be a very slow indexing seen with the current `managedAccounts` endpoint, which is experiencing quite a delay when trying to index an account that was just created via the `terraform apply` operation on this resource. This appears to be an expected consequence, as the endpoint is meant to query "all" managed accounts within the customer's organization - so, as an experimental workaround (that is ought to be made permanent, after a few tests within this PR), the objective is to change this approach to use the new `customeradministration` endpoint to be able to fetch the account by its ID (though this seems to be experiencing a significant delay as well, <= 10 seconds), and identify if this helps remediate the case customers are seeing with this resource. 

## Type of change

- [x] Enhancement (non-breaking change which adds functionality)

### Resource: `newrelic_account_management`
**File**: `newrelic/resource_newrelic_account_management.go`

Full CRUD operations for account lifecycle:
- **Create/Update**: Uses `accountmanagement` package
- **Read**: Uses `customeradministration.GetAccounts()` with precise organization ID and account ID filters, implements retry logic for eventual consistency
- **Delete**: Cancels account with warning message

Schema: `name` (string, required), `region` (string, required: `us01`|`eu01`)

### Read Efficiency Improvement

Previous approach would require multiple API calls or broad queries. New implementation:
- Resource reads: Single `GetAccounts()` call with dual filters (org + account ID)
- Data source reads: Scoped `ListAccountsWithContext()` with client-side filtering
- 10s delay post-create/update for backend indexing consistency

### API Packages
- `newrelic-client-go/v2/pkg/accountmanagement` - lifecycle operations
- `newrelic-client-go/v2/pkg/customeradministration` - filtered reads

## Checklist:

### Resource test:
```hcl
resource "newrelic_account_management" "test" {
  name   = "test-account"
  region = "us01"
}
```

### DIFF
--- a/newrelic/resource_newrelic_account_management.go
+++ b/newrelic/resource_newrelic_account_management.go
@@ -26,6 +26,11 @@ func resourceNewRelicAccountManagement() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(5 * time.Minute),
+		},
 		Schema: map[string]*schema.Schema{
 			NewRelicAccountManagementSchemaName: {
 				Type:        schema.TypeString,
--- a/newrelic/resource_newrelic_account_management_test.go
+++ b/newrelic/resource_newrelic_account_management_test.go
@@ -54,26 +54,22 @@ func TestAccNewRelicAccountManagement_Basic(t *testing.T) {
 }
 
 func TestAccNewRelicAccountManagement_Import(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pranav-new-relic:
@copilot based on all of the files changed in this PR as you may find, can you please generate a comprehensive, yet concise description for this PR that I can add to the description field here? You may want to do so by generating a temporary README.md file in the patch that you would open, from which I shall copy the contents and attach to the description of the PR here. Please ensure that we cover key details - the description does not need to be too elaborate, but should be able to briefly explain the changes made here, the packages affected and a gist of things which have moved across such as datatypes, the API files and more.

PR #2989, opened 2025-12-17
feat(alerts): crud actions for Compound Conditions

gffuentes:
# Description

Adds terraform support for the new compound conditions feature. Allows users to CRUD compound conditions from terraform.

Fixes # ([issue](https://new-relic.atlassian.net/browse/NR-499013))

## Type of change

## Checklist:

- [x] New and existing unit tests pass locally with my changes. Go [here](https://github.com/newrelic/terraform-provider-newrelic/blob/main/CONTRIBUTING.md#testing) for instructions on running tests locally. **NOTE:** Feature will be released as limited preview

## How to test this change?

example .tf config:

```
resource "newrelic_alert_compound_condition" "complex" {
  account_id         = 400304
  policy_id          = "1465274"
  name               = "[terra] Complex Infrastructure Alert"
  enabled            = false
  trigger_expression = "(A AND B) OR C"
  component_conditions {
    id    = 9155302
    alias = "A"
  }
  component_conditions {
    id    = 9143033
    alias = "B"
  }
  component_conditions {
    id    = 5266866
    alias = "C"
  }
}
```

### DIFF
--- a/newrelic/resource_newrelic_alert_compound_condition_test.go
+++ b/newrelic/resource_newrelic_alert_compound_condition_test.go
@@ -1,5 +1,4 @@
 //go:build integration || ALERTS
-// +build integration ALERTS
 
 package newrelic
 
--- a/newrelic/structures_newrelic_alert_compound_condition_test.go
+++ b/newrelic/structures_newrelic_alert_compound_condition_test.go
@@ -1,5 +1,4 @@
 //go:build unit
-// +build unit
 
 package newrelic
 


newrelic/resource_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,227 @@
...
+		ReadContext:   resourceNewRelicAlertCompoundConditionRead,
+		UpdateContext: resourceNewRelicAlertCompoundConditionUpdate,
+		DeleteContext: resourceNewRelicAlertCompoundConditionDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
+			"account_id": {
+				Type:        schema.TypeInt,
+				Optional:    true,
```
akane0915, line 28: Why is this optional? Should be required, no?

newrelic/resource_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,227 @@
...
+				Computed:    true,
```
akane0915, line 29: Why is this computed? According to gemini, 
```
In the context of a Terraform provider written in Go, the Computed field is a boolean setting within a resource's schema. It tells Terraform that the value of a specific attribute is not determined by the user's configuration, but is instead generated or returned by the underlying API or cloud provider.
```
This is user provided, so doesn't seem like it should be marked as computed. Thoughts?

newrelic/resource_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,227 @@
...
+				Description: "The ID of the policy where this condition should be used.",
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The title of the compound alert condition.",
+			},
+			"enabled": {
+				Type:        schema.TypeBool,
+				Optional:    true,
```
akane0915: Why is this optional? Should be required, no?

newrelic/resource_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,227 @@
...
+			"trigger_expression": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Expression that defines how component condition evaluations are combined. Valid operators are 'AND', 'OR', 'NOT'. For more complex expressions, use parentheses. Simple example: 'A AND B'. Complex example: 'A AND (B OR C) AND NOT D'.",
+			},
+			"component_conditions": {
+				Type:        schema.TypeSet,
+				Required:    true,
+				MinItems:    2,
+				Description: "The list of NRQL conditions to be combined. Each component condition must be enabled. Must have at least 2 components.",
```
akane0915: Legacy conditions are valid component conditions, so we should not specify `NRQL`. I would just say `The list of existing conditions to be combined using the `trigger_expression. Each component condition must be enabled. Must have at least 2 component conditions.`

newrelic/resource_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,227 @@
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "The ID of the existing NRQL alert condition to use as a component.",
```
akane0915: Don't specify `NRQL`. Legacy conditions are valid.

newrelic/resource_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,227 @@
...
+			"runbook_url": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Runbook URL to display in notifications.",
+			},
+			"threshold_duration": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Computed:    true,
+				Description: "The duration, in seconds, that the trigger expression must be true before the compound alert condition will activate. Between 30-1440 seconds.",
```
akane0915: Max is actually 24 hours, so `86,400 seconds`

newrelic/resource_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,227 @@
...
+			},
+			"entity_guid": {
```
akane0915: Why is `entity_guid` part of this schema?

newrelic/resource_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,227 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The unique entity identifier of the alert compound condition in New Relic.",
+			},
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Second),
+		},
+	}
```
akane0915, line 99: Why 30 seconds? Is this what happens for the nrql condition resource?

newrelic/resource_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,227 @@
...
+		return diag.FromErr(err)
+	}
+
+	if len(conditions) == 0 {
+		log.Printf("[WARN] Compound alert condition %s not found, removing from state", conditionID)
+		d.SetId("")
+		return nil
+	}
+
+	condition := conditions[0]
```
akane0915, line 166: So how does this work? Does this only return one condition? Are we able to support the same functionality as the Nerdgraph endpoint which allows sorting, filtering, and returns all condition if no filter provided? Asking this naively since I'm not sure we need to support that.

akane0915: Gemini:

How Terraform handles "Multiple" items
If a user has 10 compound alert conditions in their configuration, Terraform will call this function 10 separate times (often in parallel).

Instance A: Terraform calls Read, passes ID "123", function fetches "123".

Instance B: Terraform calls Read, passes ID "456", function fetches "456".

What if you wanted to read "All" of them?
If you wanted to fetch all conditions and display them (without managing them as individual resources), you wouldn't use a Resource. Instead, you would use a Data Source.

In a Data Source Read function, you would:

Remove the ID filter.

Loop through the entire conditions slice.

Save the list into a schema.TypeList or schema.TypeSet in the state.

Summary
This function is a "Resource Read." Its job is to verify that the specific resource Terraform thinks it owns still exists and is up to date. It acts as a one-to-one bridge between a single block in a .tf file and a single object in the New Relic database.

newrelic/structures_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return &input, nil
+}
+
+// expandAlertCompoundConditionUpdateInput builds the update input from Terraform schema
+func expandAlertCompoundConditionUpdateInput(d *schema.ResourceData) (*alerts.CompoundConditionUpdateInput, error) {
+	policyID := strconv.Itoa(d.Get("policy_id").(int))
+	name := d.Get("name").(string)
+	enabled := d.Get("enabled").(bool)
+	triggerExpression := d.Get("trigger_expression").(string)
```
akane0915, line 50: I'm trying to understand how `update` works. Does this mean that if you create a compound alert, with say, `name: My first compound alert`, and then you go to update the condition to change a different field, say the `facet_matching_behavior`, do you need to provide the name again? Otherwise, will the name get overriden with `nil`? Ideally, I think the original name wouldn't get overriden unless another name is provided.

newrelic/resource_newrelic_alert_compound_condition_test.go
```diff
@@ -0,0 +1,396 @@
...
+		alias = "A"
+	}
+
+	component_conditions {
+		id    = split(":", newrelic_nrql_alert_condition.condition_b.id)[1]
+		alias = "B"
+	}
+}
+`, name, testAccountID)
+}
```
akane0915, line 395: Nit: Could test removing an optional field on update.

website/docs/r/alert_compound_condition.html.markdown
```diff
@@ -0,0 +1,293 @@
...
+layout: "newrelic"
+page_title: "New Relic: newrelic_alert_compound_condition (BETA PREVIEW)"
+sidebar_current: "docs-newrelic-resource-compound-alert-condition"
+description: |-
+  Create and manage compound alert conditions in New Relic. This feature is in Beta Preview.
+---
+
+# Resource: newrelic_compound_alert_condition (BETA PREVIEW)
+
+Use this resource to create and manage compound alert conditions in New Relic. Compound conditions allow you to combine multiple alert conditions using logical expressions (AND, OR) to create more sophisticated alerting logic.
```
akane0915: Nit: Change to `(AND, OR, NOT)`

website/docs/r/alert_compound_condition.html.markdown
```diff
@@ -0,0 +1,293 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `policy_id` - (Required) The ID of the policy where this alert compound condition should be used.
+- `name` - (Required) The title of the compound alert condition.
+- `trigger_expression` - (Required) Expression that defines how component condition evaluations are combined. Valid operators are 'AND', 'OR', 'NOT'. For more complex expressions, use parentheses. Use the aliases from `component_conditions` to build expressions like `"A AND B"`, `"A OR B"`, `"(A AND B) OR C"`, or `"A AND NOT B"`.
+- `component_conditions` - (Required) The list of NRQL conditions to be combined. Each component condition must be enabled. Must include at least 2 components. See [Component Conditions](#component-conditions) below for details.
+- `enabled` - (Optional) Whether or not the compound alert condition is enabled. Defaults to `true`.
```
akane0915: Enabled is not optional

website/docs/r/alert_compound_condition.html.markdown
```diff
@@ -0,0 +1,293 @@
...
+- `account_id` - (Optional) The New Relic account ID for managing your compound alert conditions. Defaults to the account ID set in your environment variable `NEW_RELIC_ACCOUNT_ID`.
```
akane0915, line 236: Account ID is not optional

website/docs/r/alert_compound_condition.html.markdown
```diff
@@ -0,0 +1,293 @@
...
+- `facet_matching_behavior` - (Optional) How the compound condition will take into account the component conditions' facets during evaluation. Valid values are:
+  - `FACETS_IGNORED` - (Default) Facets are not taken into consideration when determining when the compound alert condition activates
+  - `FACETS_MATCH` - The compound alert condition will activate only when shared facets have matching values
+- `runbook_url` - (Optional) Runbook URL to display in notifications.
+- `threshold_duration` - (Optional) The duration, in seconds, that the trigger expression must be true before the compound alert condition will activate. Between 30-1440 seconds.
```
akane0915: Upper limit is `86400 sec`

website/docs/r/alert_compound_condition.html.markdown
```diff
@@ -0,0 +1,293 @@
...
+
+### Component Conditions
+
+The `component_conditions` block supports the following arguments:
+
+- `id` - (Required) The ID of the existing NRQL alert condition to use as a component.
```
akane0915: Remove `NRQL`

website/docs/r/alert_compound_condition.html.markdown
```diff
@@ -0,0 +1,293 @@
...
+- `alias` - (Required) The identifier that will be used in the compound alert condition's `trigger_expression` (e.g., 'A', 'B', 'C'). Must start with a letter and contain only letters, numbers, and underscores. Must be unique within the alert compound condition.
```
akane0915: `Must start with a letter and contain only letters, numbers, and underscores.` This isn't accurate. Valid values are a, b, c, d, e.

website/docs/r/alert_compound_condition.html.markdown
```diff
@@ -0,0 +1,293 @@
...
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+- `id` - The ID of the compound alert condition.
+- `entity_guid` - The unique entity identifier of the alert compound condition in New Relic.
```
akane0915: I don't think we expose the entity_guid in any of the endpoints.

newrelic/structures_newrelic_alert_compound_condition.go
```diff
@@ -0,0 +1,158 @@
...
+	_ = d.Set("runbook_url", condition.RunbookURL)
+	_ = d.Set("threshold_duration", condition.ThresholdDuration)
+
+	// Flatten component conditions - ONLY id and alias (per user requirement)
+	componentConditions := flattenComponentConditions(condition.ComponentConditions)
+	if err := d.Set("component_conditions", componentConditions); err != nil {
+		return fmt.Errorf("error setting component_conditions: %v", err)
+	}
+
+	// Note: entity_guid might be available in the API response if needed
```
akane0915: Is this accurate?

newrelic/structures_newrelic_alert_compound_condition.go
akane0915: Did you have a response to this @gffuentes ?

website/docs/r/alert_compound_condition.html.markdown
```diff
@@ -0,0 +1,292 @@
...
+  - `FACETS_MATCH` - The compound alert condition will activate only when shared facets have matching values
+- `runbook_url` - (Optional) Runbook URL to display in notifications.
+- `threshold_duration` - (Optional) The duration, in seconds, that the trigger expression must be true before the compound alert condition will activate. Between 30-86400 seconds.
+
+### Component Conditions
+
+The `component_conditions` block supports the following arguments:
+
+- `id` - (Required) The ID of the existing alert condition to use as a component.
+- `alias` - (Required) The identifier that will be used in the compound alert condition's `trigger_expression` (e.g., 'b', 'b', 'c', 'd', 'e').
```
akane0915: Typo here. Should be `a`

newrelic/resource_newrelic_alert_compound_condition.go
gffuentes: yeah im just copying from that from resource_newrelic_nrql_alert_condition

newrelic/resource_newrelic_alert_compound_condition.go
gffuentes: for nrql_alert_condition we also dont have a way to view multiple. I didnt know that about data sources, we do have a couple of data sources in this repo but im not sure if they are being used to view multiples

pranav-new-relic: @akane0915 @gffuentes I just discovered this chat, so wanted to join in :) 

Ideally, from a "schema" POV, a resource tends to correspond only to one item (e.g. an Alert Condition), or a mapping of one item to another (e.g. an Alert Policy Channel - a legacy example :)). This is because - logically, when you're writing the configuration of a resource, you're writing the values of attributes for "that one item" to be managed, which would result in the creation of one resource - since we wouldn't be able to specify multiple, a resource is not expected to help "read" multiple items too. Additionally, a resource can possess one ID, so it'd be more of a database unique/primary key constraint compelling the management of one item via the resource. 

If you guys do want the ability to read through multiple items, a data source is better suited just like Gemini said, but it isn't completely right too. We'd have a similar schema defined with a data source, and so, the one item that is supposed to emerge as the pick (based on the query we use to search) is intended to be used with the attributes in the schema to fetch all values; if there are multiple such items intended to be read, we cannot have the schema effectively allow access values of all attributes (Item1.Attr1, Item2.Attr1, etc do not work). In such a case, you might need to rewrite the schema of the data source to support a string as the final response, which would include a JSON-ized version of all the items found in the search results, that can then be decoded at the customer's end using the `jsondecode()` function, something we can make them aware of in the docs of the provider - but in this case too, the onus is on the customer to sift through multiple results - though it's not something that should not be done.

Just wanted to let you guys know, so you can take a call, if you'd like to add a data source :)

PR #3026, opened 2026-02-10
fix(newrelic-browser-application): add retry mechanism for browser resource

Aashirwadjain:
# Description

This PR aims to add retry mechanism for `newrelic_browser_application` resource.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/newrelic/resource_newrelic_agent_application_browser.go
+++ b/newrelic/resource_newrelic_agent_application_browser.go
@@ -131,7 +131,7 @@ func resourceNewRelicBrowserApplicationRead(ctx context.Context, d *schema.Resou
 			return resource.NonRetryableError(err)
 		}
 
-		if resp == nil {
+		if resp == nil || *resp == nil {
 			return resource.RetryableError(fmt.Errorf("entity with GUID %s not found", guid))
 		}
 


newrelic/resource_newrelic_agent_application_browser.go
```diff
@@ -109,7 +114,26 @@ func resourceNewRelicBrowserApplicationCreate(ctx context.Context, d *schema.Res
...
+		}
+
+		return nil
+	})
+
+	if retryErr != nil {
+		return diag.FromErr(retryErr)
+	}
+
+	return nil
```
pranav-new-relic: why have the retry in the create function and not in read?

Aashirwadjain: Fixed.

Aashirwadjain:
Yes @pranav-new-relic, I was able to reproduce it and fix it.

### Before changes:
image

### After changes:
image

Aashirwadjain:
### Create:
image

### Import with wrong GUID:
image


### Import with correct GUID:
image

pranav-new-relic:
> Upon testing the negative case (e.g. during read), the error shown is entity type is not BrowserApplicationEntity and not the real RetryableError, meaning something's not right with the logic, and it is likely that the negative case was never tested with these changes. Please ensure the negative case is tested, the logic is fixed to show the actual RetryableError in this case as the current error would be misleading to someone trying to use a wrong entity GUID (and they end up seeing an error that sounds like a bug, and not suggestive of the entity not being found)

Aashirwadjain:
### Read (negative case):
image

### Read (positive case):
image

PR #3036, opened 2026-02-23
fix(data_source): use ExactName for notification destination lookup

shubh-0708:
Description
Changed the newrelic_notification_destination data source to use the ExactName filter instead of Name when searching by name. This ensures exact name matching rather than "contains" matching.

  Previously, searching for a destination with name "foo" could return destinations named "foobar" or "myfoo". Now it will only return destinations with exactly the name "foo".

  Depends on: newrelic/newrelic-client-go#1374

  Type of change

  - Bug fix (non-breaking change which fixes an issue)

  Checklist:

  - My commit message follows https://www.conventionalcommits.org/en/v1.0.0/
  - My code is formatted to https://go.dev/blog/gofmt
  - I have performed a self-review of my own code
  - I have made corresponding changes to the documentation
  - I have added tests that prove my fix is effective or that my feature works
  - New and existing unit tests pass locally with my changes. Go https://github.com/newrelic/terraform-provider-newrelic/blob/main/CONTRIBUTING.md#testing for instructions on running tests locally.

  How to test this change?

  1. Create multiple notification destinations with similar names (e.g., "test", "test-email", "mytest")
  2. Use the newrelic_notification_destination data source with name = "test"
  3. Before fix: Could return "test-email" or "mytest" due to contains matching
  4. After fix: Returns only the destination with exactly the name "test"

  Example Terraform configuration:
  data "newrelic_notification_destination" "example" {
    name = "test"
  }

  output "destination_id" {
    value = data.newrelic_notification_destination.example.id
  }

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/newrelic/go-agent/v3 v3.30.0
 	github.com/newrelic/go-insights v1.0.3
-	github.com/newrelic/newrelic-client-go/v2 v2.78.0
+	github.com/newrelic/newrelic-client-go/v2 v2.79.1
 	github.com/stretchr/testify v1.9.0
 	golang.org/x/exp v0.0.0-20240325151524-a685a6edb6d8
 )
--- a/go.sum
+++ b/go.sum
@@ -270,8 +270,8 @@ github.com/newrelic/go-agent/v3 v3.30.0 h1:ZXHCT/Cot4iIPwcegCZURuRQOsfmGA6wilW+S
 github.com/newrelic/go-agent/v3 v3.30.0/go.mod h1:9utrgxlSryNqRrTvII2XBL+0lpofXbqXApvVWPpbzUg=
 github.com/newrelic/go-insights v1.0.3 h1:zSNp1CEZnXktzSIEsbHJk8v6ZihdPFP2WsO/fzau3OQ=
 github.com/newrelic/go-insights v1.0.3/go.mod h1:A20BoT8TNkqPGX2nS/Z2fYmKl3Cqa3iKZd4whzedCY4=
-github.com/newrelic/newrelic-client-go/v2 v2.78.0 h1:RxQ0pkoFCW4U+a2olLh8z+kOG+q3gxnXMdU6z7vvM68=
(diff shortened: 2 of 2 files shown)


newrelic/data_source_newrelic_notifications_destination.go
```diff
@@ -21,7 +21,7 @@ func dataSourceNewRelicNotificationDestination() *schema.Resource {
 			"id": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				ExactlyOneOf: []string{"id", "name"},
+				ExactlyOneOf: []string{"id", "name", "exact_name"},
```
tansar-newrelic, line 24: does it throw exception if you are trying to add more than one field. show me how it behaves

newrelic/data_source_newrelic_notifications_destination.go
```diff
@@ -112,12 +122,15 @@ func dataSourceNewRelicNotificationDestinationRead(ctx context.Context, d *schem
 
 	if len(destinationResponse.Entities) == 0 {
 		d.SetId("")
-		if idValue != "" && nameValue == "" {
+		if idValue != "" {
```
tansar-newrelic, line 125: if we are removing this nameValue condition, is it based on the assumption that only one of these three can be provided in the request ?
reference to this - https://github.com/newrelic/terraform-provider-newrelic/pull/3036/changes#r2882120847

newrelic/data_source_newrelic_notifications_destination.go
shubh-0708: │ Error: Invalid combination of arguments
│ 
│   with data.newrelic_notification_destination.extact_name,
│   on main.tf line 23, in data "newrelic_notification_destination" "extact_name":
│   23:   name = "amit"
│ 
│ "name": only one of `exact_name,id,name` can be specified, but `exact_name,name` were specified

newrelic/data_source_newrelic_notifications_destination.go
shubh-0708: Yes, that's correct. The schema defines ExactlyOneOf: []string{"id", "name", "exact_name"} which ensures Terraform validates that exactly one of these three attributes is provided. So when we reach this error handling code, only one of idValue, nameValue, or exactNameValue will be non-empty. The simplified conditions are safe because of this schema-level enforcement.

pranav-new-relic (approved):
LGTM

PR #3059, opened 2026-04-09
fix(region): add Japan (JP) region support

akifullahkhan-png:
## Summary
Follow-up to #3065 (which added core JP region support). This PR adds two remaining items that weren't included in that PR:

1. JP CloudWatch metrics endpoint for the AWS cloud-integrations example
2. A JP case-insensitivity test case in the provider region integration test

## Changes
- `examples/modules/cloud-integrations/aws/main.tf` — add `JP = "https://aws-api.jp.nr-data.net/cloudwatch-metrics/v1"` to the `newrelic_urls` map
- `newrelic/provider_newrelic_integration_test.go` — add a test case verifying that lowercase `"jp"` is rejected with the expected error (case-insensitivity validation)

## Context
The bulk of JP region support (provider validation, alert channel, account_management, docs, client-go bump) was merged in #3065. This PR is scoped down to just the AWS integration URL and the additional test case.

## Test plan
- [ ] Unit tests pass (`go test -tags unit ./newrelic/...`)
- [ ] `TestAccNewRelicProvider_Region` integration test passes with the new JP case


examples/modules/cloud-integrations/aws/main.tf
```diff
@@ -125,6 +125,7 @@ locals {
   newrelic_urls = {
     US      = "https://aws-api.newrelic.com/cloudwatch-metrics/v1"
     EU      = "https://aws-api.eu01.nr-data.net/cloudwatch-metrics/v1"
+    JP      = "https://aws-api.jp01.nr-data.net/cloudwatch-metrics/v1"
```
akifullahkhan-png, line 128: Need to confirm this Endpoint.

akifullahkhan-png: Confirmed the endpoint with ingest team- https://newrelic.slack.com/archives/C01508Q2879/p1775542419946949

pranav-new-relic (approved):

PR #3062, opened 2026-04-14
fix(log-parsing-rule): prevent state corruption on authentication errors

pranav-new-relic:
## Summary
This PR fixes a critical bug in `newrelic_log_parsing_rule` where invalid API credentials cause silent state file corruption. Previously, authentication errors during `terraform plan` or `apply` would remove the resource from state instead of failing with an error, requiring manual `terraform import` to recover. The fix updates error handling to properly distinguish between "resource not found" (404) and other errors like authentication failures, aligning with the pattern used across 24+ other resources in this provider.

## Changes
- Updated `resourceNewRelicLogParsingRuleRead` to use type assertion for `NotFound` errors
- Modified `getLogParsingRuleByID` to return properly typed `NotFound` errors
- Added import for `nrErrors` from newrelic-client-go error package
- Authentication and permission errors now properly propagate to Terraform instead of silently clearing state

Resolves NR-548992

### DIFF
--- a/newrelic/resource_newrelic_log_parsing_rule.go
+++ b/newrelic/resource_newrelic_log_parsing_rule.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/newrelic/newrelic-client-go/v2/newrelic"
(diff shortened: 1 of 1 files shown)


pranav-new-relic:
Tested manually - fix works as expected. Auth errors now properly fail instead of corrupting state. Ready to merge.

nr-developer-toolkit (approved):
LGTM

PR #3106, opened 2026-06-10
test(federated-logs): skip fedlogs tests + bump client to v2.89.0

pranav-new-relic:
## What

Mirrors [newrelic-client-go#1425](https://github.com/newrelic/newrelic-client-go/pull/1425) on the Terraform provider side: takes the federated-logs unit and acceptance tests off the default CI path while the shared test account doesn't hold the `federated_logs` entitlement, and pulls in the matching client release.

### Skip federated-logs tests

- **`newrelic/structures_newrelic_federated_logs_test.go`** — switched the build tag from `unit` to a placeholder `unit_fedlogs_skip` so the file is excluded from `make test-unit` (`-tags=unit`).
- **`newrelic/resource_newrelic_federated_logs_setup_test.go`** and **`newrelic/resource_newrelic_federated_logs_partition_test.go`** — dropped the `integration` half of the build tag (`integration || LOGGING_INTEGRATIONS` → `LOGGING_INTEGRATIONS`) so these acceptance tests no longer run under `make test-integration-all` (`-tags=integration`). They can still be exercised explicitly with `-tags=LOGGING_INTEGRATIONS` when the entitlement is in place.

## Why

The federated-logs API gateway returns `ACCESS_DENIED` for the shared test account. From inspection of `logging/federated-logs-api-gateway` and `entity-platform/ep-next-gen-api`, every federated-logs op is gated by **two** checks:

1. **Capability** — `federated_log.{create,read,update,delete}.{setup,partition}`
2. **Entitlement** — account must have `federated_logs` on `productLine=generic` per AIS

The feature flag `Logging/federated_logs` is already on for the test account, so the failure is at the entitlement and/or capability layer. PASS treats `federated_logs` as packaged (tied to a SKU), so granting it requires either a subscription update or a manual cohort add via `#help-product_and_subs`. Tracking that separately; this PR just unblocks CI in the meantime.

## Test plan

- [x] `go build ./...` — clean
- [x] `go test -count=1 -tags=unit ./newrelic/...` — green (file is excluded)
- [x] `go mod tidy` — no further changes
- [ ] CI green on this PR

## Follow-up

Once the test account is granted the `federated_logs` entitlement, revert each build-tag change (`unit_fedlogs_skip` → `unit`; reinstate `integration ||` on the two acceptance test files).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/newrelic/go-agent/v3 v3.30.0
 	github.com/newrelic/go-insights v1.0.3
-	github.com/newrelic/newrelic-client-go/v2 v2.88.0
+	github.com/newrelic/newrelic-client-go/v2 v2.89.0
 	github.com/stretchr/testify v1.9.0
 	golang.org/x/exp v0.0.0-20240325151524-a685a6edb6d8
 	gopkg.in/yaml.v3 v3.0.1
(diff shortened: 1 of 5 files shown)


nr-developer-toolkit (approved):
LGTM

To be reverted after there is adequate confidence on the test account/user having the right capabilities

PR #3111, opened 2026-06-11
fix(dashboard): make variable nrql_query account_ids optional without drift

jsgithub1:
### Summary

Allows `account_ids` to be omitted (or set to an empty list) in a dashboard variable's `nrql_query` block, defaulting to the provider's configured account ID. Critically, after `terraform apply` a subsequent `terraform plan` reports **no drift** — and in particular no drift proposing to discard the defaulted account ID — while genuine changes still surface as drift.

### Background / what changed since the initial review

The first iterations of this PR tried to default the value in `CustomizeDiff` with the attribute marked `Optional + Computed`. That approach cannot work, for a concrete SDK reason:

- `ResourceDiff.SetNew` (and `SetNewComputed`) only accept **top-level** schema keys — `checkKey` is called with `nested=false`. A nested list key like `variable.0.nrql_query.0.account_ids` is always rejected with `SetNew: invalid key`, which fails the plan at apply time.
- Marking the list `Computed` additionally causes an *omitted* value to be carried forward from prior state, which masks real drift (e.g. an account changed in the UI while the config stays empty).

This is the dead-end raised in earlier review. It has been replaced with **read-side normalization**, which needs neither `Computed` nor `CustomizeDiff`.

### Approach

`account_ids` is plain `Optional`:

- **Write (`expand`)**: an empty/omitted `account_ids` is defaulted to the provider account ID before being sent upstream.
- **Read (`flatten`)**: when the configuration is empty **and** the upstream value is exactly `[providerAccountID]`, the value is normalized back to an empty list, so state matches the empty config and no drift is reported. Any other upstream value — a different account, multiple accounts, or a value changed in the UI — is written through unchanged, so real changes still drift. Explicitly configured values are always written through as-is.

The provider account ID is threaded from `meta` through the flatten call chain.

### Behavior

| Config transition | Plan result |
|---|---|
| omitted → omitted / empty → empty / omitted ↔ empty | no drift |
| explicit `[X]` → explicit `[X]` (unchanged) | no drift |
| omitted/empty ↔ explicit (form change, incl. `[pDef]`) | drift (convergent — a real config change) |
| explicit `[A]` → explicit `[B]` | drift |
| upstream value ≠ provider default while config is empty (e.g. UI change) | drift |

### Testing

- **Acceptance** (`TestAccNewRelicOneDashboard_VariableAccountIDsDriftDetection`): passes against the live API for all provider-default-account scenarios (omitted/empty no-drift, explicit-value no-drift, and the convergent form-change drift cases). Cross-account *value-change* scenarios (S11/S12) require a second accessible account and are gated behind `NEW_RELIC_SUBACCOUNT_ID`.
- **Unit** (`TestFlattenVariableNRQLQuery_AccountIDsNormalization`): covers the read-side normalization directly, including the "written through unchanged" cases that back the value-change scenarios.

### Notes

- The branch was rebased onto `main`, so earlier inline review comments may show as outdated.

### DIFF
--- a/newrelic/resource_newrelic_one_dashboard.go
+++ b/newrelic/resource_newrelic_one_dashboard.go
@@ -150,7 +150,7 @@ func dashboardVariableSchemaElem() *schema.Resource {
 						"account_ids": {
 							Type:        schema.TypeList,
 							Optional:    true,
-							Description: "New Relic account ID(s) to issue the query against.",
+							Description: "New Relic account ID(s) to issue the query against. Defaults to the account ID specified in the provider configuration.",
 							Elem: &schema.Schema{
 								Type: schema.TypeInt,
 							},
@@ -934,7 +934,7 @@ func resourceNewRelicOneDashboardCreate(ctx context.Context, d *schema.ResourceD
 			return diag.FromErr(err)
 		}
 
-		return diag.FromErr(flattenDashboardUpdateResult(result, d))
+		return diag.FromErr(flattenDashboardUpdateResult(result, d, meta.(*ProviderConfig).AccountID))
 
 	}
 
@@ -959,7 +959,7 @@ func resourceNewRelicOneDashboardRead(ctx context.Context, d *schema.ResourceDat
 		return diag.FromErr(err)
 	}
 
-	return diag.FromErr(flattenDashboardEntity(dashboard, d))
+	return diag.FromErr(flattenDashboardEntity(dashboard, d, providerConfig.AccountID))
 }
 
 func resourceNewRelicOneDashboardUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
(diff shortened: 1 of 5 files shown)


pranav-new-relic (approved):
LGTM, thank you for the contribution @jsgithub1 :) 

We might still want to handle a few edge cases around imports as I see them and similar tiny gaps (which is why we usually do not prefer handling state adjustments in expand/flatten functions, outside the immediate scope of the functions directly managing attribute state), but the extra-ordinary case with this resource does compel us to do this and I'd totally agree with you - I would do the same after all deliberation. I'll look into closing the remnants as necessary - that's on me.

Thanks a lot for the work here, again - James :)

pranav-new-relic:
Mandatory local check, cuz integration tests are disabled on our forks

image

image

All local testing by @pranav-new-relic confirms these changes introduce no unintended repercussions, and should fix the issue; we should be good to merge.

pranav-new-relic:
@jsgithub1 This is now out with v3.95.2 🎉

PR #3114, opened 2026-06-17
fix(workload): add backticks automatically within workload terraform 

pjaiswal-sys:
# Description

When creating a workload using the New Relic Terraform Provider, the entity query is not formatted correctly. The issue seems to be related to Terraform failing to add the needed backticks when creating workloads with search queries. They mentioned that they can manually add the backticks, and this will resolve the issue, however if they're necessary it would be ideal to have the provider auto-correct them or produce an error on the invalid format. 

Customer is expecting to see a formatting error, or have the Terraform Provider fix the formatting for them, in relation to using entity_search_query and tags.

## Type of change

## Checklist:


rkumarburidi (approved):

rkumarburidi (dismissed):

pranav-new-relic:
[FYI](https://newrelic.slack.com/archives/CP9Q30JDC/p1782184951148769?thread_ts=1782129240.751669&cid=CP9Q30JDC):

newrelic/resource_newrelic_workload.go
```diff
@@ -53,6 +53,16 @@ func resourceNewRelicWorkload() *schema.Resource {
...
+				Set: func(v interface{}) int {
+					// Custom hash function that normalizes queries before hashing
+					// This ensures queries that only differ in backtick formatting are treated as identical
+					m := v.(map[string]interface{})
+					if query, ok := m["query"].(string); ok {
+						normalizedQuery := formatEntitySearchQueryTags(query)
+						return schema.HashString(normalizedQuery)
+					}
+					return 0
+				},
```
pranav-new-relic, line 65: @pjaiswal-sys this is an interesting solution to the bug we discovered (rarely used in New Relic's Terraform Provider), thanks for adding this 👍 

I've tested this myself too and I don't seem to be able to reproduce the bug so I think this is fine, except I see two edge cases with this:
- while no customer would never intentionally do so, if we have someone specify the list `entity_search_query` with a backticked version and a non backticked version of the same query, Terraform would now see both of these as identical (as this function changes how the hash is computed) so this could be a minor unintended side effect/edge case - it'll lead to drift accordingly (though no one would really do that unless they land at this accidentally)
- when someone imports an existing workload into Terraform with a backticked `entity_search_query`, the backticked version will continue to exist in their config (that is generated via the import) which is again not a problem, but could be a bit inconsistent if other resources they define have no backticks (and we're doing it for them)

While these are not major concerns like I said, I think it may be beneficial to add a note to the documentation suggesting customers to use backticks with `tags.*` clauses in the `entity_search_query` list with some examples, so we can have this behaviour introduced by the customer themselves for newer resources, while everything we've added acts as a obvious fallback mechanism for new resources (if they skip adding backticks themselves) / converts existing resources.

pranav-new-relic:
~@pjaiswal-sys also, it looks like the test cases with the backticked versions have been discarded, and I think you may have done that since the bug we talked about is reproducible when we don't add backticks (and the function you've added operates on that clause) - that is completely logical, but just so we don't run into accidental failures with future edits to the resource (the backticked scenario is covered against future modifications if any), it might make sense to have a new test added/existing test modified to have a backticked clause~

I've just added it as part of https://github.com/newrelic/terraform-provider-newrelic/pull/3114/commits/dd1da84c8d96bc7d119f55dea7cc937651b0ff90

newrelic/resource_newrelic_workload.go
pjaiswal-sys: Backticks are required in entity_search_query tags due to dependencies that fail when backticks are absent.

  New Workload Creation:
  - Backticks are automatically added to all entity_search_query tags
  - Works correctly regardless of whether backticks are provided in the input
  - Dependency map functions properly

  Existing Workloads:
  - May contain inconsistencies (some tags with backticks, some without)
  - Inconsistencies will be resolved when the workload is updated
  - Upon update, all entity_search_query tags will be normalized with backticks

  Testing

  Tested scenario with partial backtick coverage:
  - Some entity_search_query tags had backticks
  - Some entity_search_query tags did not have backticks
  - Result: All tags were successfully updated to include backticks

pranav-new-relic: @pjaiswal-sys thanks for adding this - I think this is true (and I did find these observations when I tested this too), but just to clarify what I meant by the comment above:

While all of these are expected to happen and our code covers this, there are a few _edge cases_ we need to take into account which would be introduced by this code change (which I've listed above); though these edge cases are very minor in nature, just to counter this, it might be useful to drop a note in the docs of this resource to ask customers to preferably use backticks with `tags.*` clauses on their end to counter these minimal edge cases too. 

We can skip it for now and add it later (like the case with the cross account entity search I'd discussed with you) but I just wanted to let you know about this so you have this in this PR or as a future consideration 👍 either choice is fine :)

pranav-new-relic:
@pjaiswal-sys 
- can you please fix merge conflicts in this PR and/or sync this PR with the latest on main (as we've now merged the other PR written by Manideep)? 
- once done, can you please sign the CLA here, in [this comment](https://github.com/newrelic/terraform-provider-newrelic/pull/3114#issuecomment-4729273523)
- once you're done with both of the above, please take a look at my latest comment ([here](https://github.com/newrelic/terraform-provider-newrelic/pull/3114#discussion_r3491911868)) and if you'd like to add it as part of this PR, please do so; or if we might want to do it later, we should be good to proceed with merging this PR 👍

newrelic/resource_newrelic_workload.go
pjaiswal-sys: @pranav-new-relic we take a note of it, will update the documentation on new PR or in future works, 
For now we can deploy these changes as it is

pranav-new-relic (approved):

Issue #3117, opened 2026-06-29
newrelic_monitor_downtime: provider plugin crashes on read/import — "len(tags) < 4" retry never satisfied (v3.93.1)


villelahdenvuo:
### Summary

Reading a `newrelic_monitor_downtime` (config-driven `import`, or any `plan`/refresh once it's in state) reliably crashes the provider plugin:

```
Error: Plugin did not respond
  …plugin.(*GRPCProvider).ReadResource call.
```
with the plugin process exiting `error="exit status 1"`.

### Environment
- Provider: `newrelic/newrelic` **3.93.1**
- Terraform: 1.15.6 (darwin_arm64)
- Region: **EU** (`api.eu.newrelic.com`)

### Root cause

In `resourceNewRelicMonitorDowntimeRead` (`newrelic/resource_newrelic_monitor_downtime.go`), the read retry does:

```go
entity = (*resp).(*entities.GenericEntity)   // ① unchecked type assertion
tags = entity.GetTags()
if len(tags) < 4 {                            // ② requires ≥ 4 tags
    return resource.RetryableError(fmt.Errorf("enough tags not found. retrying"))
}
```

A `MONITOR_DOWNTIME` entity only carries **3 system tags** (`account`, `accountId`, `trustedAccountId`) and no user tags, so `len(tags) < 4` is **never satisfied** → the `RetryContext` runs the full 30s, times out, and the plugin exits. Trace excerpt:

```
[WARN] WaitForState timeout after 30s
[WARN] Unable to find application entity: enough tags not found. retrying
[ERROR] provider: plugin process exited: error="exit status 1"
[ERROR] (*GRPCProvider).ReadResource: rpc error: code = Unavailable desc = error reading from server: EOF
Error: Plugin did not respond
```

(The NerdGraph entity query itself returns HTTP 200 with the full entity — the failure is purely the tag-count gate.)

Separately, the type assertion at ① has no `ok`/nil guard, which is the nil-interface panic reported in #2559 (`interface conversion: entities.EntityInterface is nil, not *entities.GenericEntity`) — same code path, so worth fixing together.

### Reproduction
```hcl
resource "newrelic_monitor_downtime" "repro" {
  name       = "Example downtime"
  mode       = "DAILY"
  start_time = "2021-10-03T17:30:00"
  end_time   = "2021-10-04T06:30:00"
  time_zone  = "Europe/Helsinki"
}

import {
  to = newrelic_monitor_downtime.repro
[...]

Issue #2559, opened 2024-01-30
Terraform Provider Plugin is Crashing


moazzamali-tcp:
### There was a outage today on newrelic and after that getting error while performing terraform changes:

Stack trace from the terraform-provider-newrelic_v3.29.0 plugin:

panic: interface conversion: entities.EntityInterface is nil, not *entities.GenericEntity

goroutine 115 [running]:
github.com/newrelic/terraform-provider-newrelic/v2/newrelic.resourceNewRelicMonitorDowntimeRead.func1()        
        /home/runner/work/terraform-provider-newrelic/terraform-provider-newrelic/newrelic/resource_newrelic_monitor_downtime.go:245 +0x34a
github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry.RetryContext.func1()
        /home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/retry/wait.go:27 +0x56
github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry.(*StateChangeConf).WaitForStateContext.func1()       
        /home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/retry/state.go:110 +0x1ff
created by github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry.(*StateChangeConf).WaitForStateContext    
        /home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/retry/state.go:83 +0x1d8

Error: The terraform-provider-newrelic_v3.29.0 plugin crashed!

This is always indicative of a bug within the plugin. It would be immensely
helpful if you could report the crash with the plugin's maintainers so that it
can be fixed. The output above should help diagnose the issue.

pranav-new-relic:
Hi @moazzamali-tcp, thank you for reporting this! Unfortunately, it could have been that the relevant outage caused a disruption to API calls performed to fetch details of entities from NerdGraph; which in the case of this resource, is  an entity with the GUID of that of the Monitor Downtime - this may have subsequently led to the provider panicking. The error you've attached is usually never expected.

I've tested the monitor downtime resource out sometime ago and now, and find that it is working fine (doesn't seem to return this error anymore) and I'm assuming it is so as the error above should have lasted only during the outage. Can you please try now and let me know if you still see the issue? Thanks!

[...]

PR #3118, opened 2026-06-29
fix(monitor_downtime): don't crash the plugin on read (log.Fatalf → diag error, guard entity assertion)

villelahdenvuo:
## Summary

`resourceNewRelicMonitorDowntimeRead` can crash the provider process. On read it retries fetching the entity for up to 30s and, on timeout, calls `log.Fatalf`:

```go
entity = (*resp).(*entities.GenericEntity)   // unguarded assertion
tags = entity.GetTags()
if len(tags) < 4 {
    return resource.RetryableError(fmt.Errorf("enough tags not found. retrying"))
}
// ...
if retryErr != nil {
    log.Fatalf("Unable to find application entity: %s", retryErr)   // os.Exit(1)
}
```

`log.Fatalf` calls `os.Exit(1)`, killing the plugin — Terraform reports it as an opaque **`Error: Plugin did not respond`** (`plugin process exited: exit status 1`).

This is reliably triggered when the entity never reaches the `>= 4` tags the retry waits for. A `MONITOR_DOWNTIME` entity that only carries the `account` / `accountId` / `trustedAccountId` system tags (e.g. importing/refreshing one created in the UI) never satisfies `len(tags) < 4`, so the retry always times out and the plugin is `os.Exit`-ed. Reproduced on **v3.93.1**, EU region. See #3117 for full trace/repro.

The unguarded `(*resp).(*entities.GenericEntity)` assertion is also the nil-interface panic from #2559 (`interface conversion: entities.EntityInterface is nil, not *entities.GenericEntity`), so this guards it too.

## Change

- Replace `log.Fatalf(...)` with `return diag.FromErr(...)` — a provider must never `os.Exit`; surface a normal error so Terraform reports a real diagnostic instead of a crash.
- Use a comma-ok + nil check on the type assertion so a transient nil/other-typed response retries instead of panicking.

No change to the happy path (entity with the expected tags still reads exactly as before).

## Testing

- `go build ./newrelic/` and `gofmt` pass.
- The success path is covered by the existing `TestAccNewRelicMonitorDowntime_*` acceptance tests (unchanged behavior).
- The crash path is awkward to exercise without mocking `client.Entities` — happy to add a test in whatever style you prefer.

## Note / possible follow-up

This PR is scoped to **stop the plugin crash**. Fully *importing* a downtime whose configuration isn't exposed via the tags the read expects (the `len(tags) < 4` case) would need a separate change to how the read reconstructs config; with this fix that scenario returns a clear error instead of crashing.

Refs #3117, #2559

### DIFF
--- a/newrelic/resource_newrelic_monitor_downtime.go
+++ b/newrelic/resource_newrelic_monitor_downtime.go
@@ -241,7 +241,13 @@ func resourceNewRelicMonitorDowntimeRead(ctx context.Context, d *schema.Resource
 		if err != nil {
 			return resource.RetryableError(err)
 		}
-		entity = (*resp).(*entities.GenericEntity)
+		// The entity may be nil or a different concrete type if it isn't indexed yet; guard the
+		// type assertion so a transient nil response retries instead of panicking the plugin.
+		genericEntity, ok := (*resp).(*entities.GenericEntity)
+		if !ok || genericEntity == nil {
+			return resource.RetryableError(fmt.Errorf("monitor downtime %s not found yet, retrying", d.Id()))
+		}
+		entity = genericEntity
 		tags = entity.GetTags()
 		if len(tags) < 4 {
 			return resource.RetryableError(fmt.Errorf("enough tags not found. retrying"))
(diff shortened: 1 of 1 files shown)


villelahdenvuo:
@pranav-new-relic Looks like you committed a fix for this to a different (dependabot) branch, so I guess this PR can be closed?

pranav-new-relic:
Hey @villelahdenvuo, since we have a few other reports coming in, I'd like to keep this PR open for a quick reference to pick up, until there's better clarity on the occurrences leading to this issue and the real solution we'd like to go forward with. Thanks for contributing again :) I shall act on this PR accordingly (based on our next steps)

villelahdenvuo:
@pranav-new-relic Any progress on this?

villelahdenvuo:
@pranav-new-relic There seems to be some changes already implemented based on the conflict. The issue #3117 remains open for reference so I'm closing this PR as obsolete to get it off my dashboard. Thanks for looking into this. :-)

PR #3125, opened 2026-07-07
feat: jp region support and region-neutral provider hardening

### LINKED ISSUE #2968: Error using newrelic_notification_destination for lookup "The argument "destination_id" is required, but no definition was found."
We're using a `newrelic_notification_destination` datasource as a lookup to make the notification channel friendly for our developers.

This is done in a module which is/can be used in multiple location. Issue it that is randomly throws an error that `destination_id` is empty in the datasource (the element itself can be found). See code below

### Terraform Version
`terraform 1.13.3`

### Affected Resource(s)
- `newrelic_notification_channel`
- `newrelic_notification_destination`

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration
```hcl
variable "notification_destinations" {
  type        = list(string)
  description = "The name of the Notification Destinations"
}

data "newrelic_notification_destination" "alert_destination" {
  for_each = toset(var.notification_destinations)
  name     = each.key
}

resource "newrelic_notification_channel" "default" {
  for_each = toset(var.notification_destinations)

  name           = data.newrelic_notification_destination.alert_destination[each.key].name
  type           = data.newrelic_notification_destination.alert_destination[each.key].type
  destination_id = data.newrelic_notification_destination.alert_destination[each.key].id
}
```

### Actual Behavior

A error might occor randomly, depending on the implementation:

```txt
Error: Missing required argument
with module.aaa["1"].module.bbb[0].module.ccc.module.ddd[0].module.eee.newrelic_notification_channel.default["my-existing-destination-name"],
on .terraform/xxx.tf line 11, in resource "newrelic_notification_channel" "default":
  destination_id = data.newrelic_notification_destination.alert_destination[each.key].id
The argument "destination_id" is required, but no definition was found.
```

Since it doesn't throw a "element not found" key error, the datasource seems to fetch the actual data, but destination appears to be empty.

Couldn't find any detailed (useful) information in the logs (when running with `TF_LOG=debug`).

### Expected Behavior
Create the notification channel

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- https://github.com/newrelic/terraform-provider-newrelic/issues/2070

pranav-new-relic:
### JP region support

- **`newrelic_insights_event`**: swap the abandoned `go-insights v1.0.3` client for `newrelic-client-go/v2/pkg/events`, so events route to the correct Insights collector (US, EU, or JP) based on the provider's `region` argument. `insights_insert_url` is now a no-op (with a warning), deprecated starting v3.95.0.
- **`newrelic_alert_channel` and `newrelic_alert_policy_channel`**: add a `CustomizeDiff` plan-time gate that returns an actionable error when `region = "JP"`, since the underlying REST v2 Alert Channels API is not provisioned on JP.
- **`newrelic_account_management`**: stop forwarding the deprecated `regionCode` field to the mutation; `region` is now a deprecated no-op argument starting v3.95.0.

### Region-neutral fixes

- **`newrelic_synthetics_secure_credential`**: fix a Read race where `entitySearch` indexing lag caused state to be silently cleared, producing a misleading "Credential already exists" error on the next apply. Read now retries through the Read timeout before clearing state.
- **`newrelic_notification_destination` data source**: mark `id` as `Computed` so `name`/`exact_name` lookups populate the `id` attribute (fixes #2968; ports #3130).

### Docs, tests, tooling

- Deprecation and JP-availability notes on the touched resources, matching the tone of existing banners on `nrql_drop_rule` / `monitor_downtime`.
- Acceptance-test updates for `account_management` (drop the removed validator) and `notification_destination` (assert `id` is set).
- New `helpers_newrelic_region.go` with a reusable `blockJPRegionDiff` factory for future JP-region gates, and a `testing/.gitignore` so local tfstate/logs stay out of commits.


newrelic/data_source_newrelic_account.go
```diff
@@ -23,9 +23,11 @@ func dataSourceNewRelicAccount() *schema.Resource {
...
 			},
 			NewRelicAccountManagementSchemaRegion: {
-				Type:        schema.TypeString,
-				Computed:    true,
-				Description: "The region code of the account (e.g., us01, eu01).",
+				Type:     schema.TypeString,
+				Computed: true,
+				Description: "The region code of the account, as reported by New Relic (e.g., " +
+					"`us01`, `eu01`, `jp01`). Read-only; used only to surface the region an account " +
+					"already lives in.",
```
pranav-new-relic: @pranav-new-relic I think we can keep the description subtle here without needing to "explain" the region in such elaborate detail, and remove references to the regions which are expected to be supported by the data source. Please ensure the same fixture in the docs of the account management data source (we might just want to delist `region`as an attribute that is expected to be exported by the data source, which we shall eventually retire from the data source's code too at some point).

newrelic/helpers_newrelic_region.go
```diff
@@ -0,0 +1,55 @@
...
+	return fmt.Errorf(
+		"%s is not supported in the JP region: the legacy REST v2 Alert "+
+			"Channels API has been deprecated globally "+
+			"since 2024 and is not being provisioned on the JP region. Migrate "+
+			"JP configurations to `newrelic_notification_destination` and "+
+			"`newrelic_notification_channel` - the modern NerdGraph-based "+
+			"replacements that are available in every region. See "+
+			"https://docs.newrelic.com/docs/alerts-applied-intelligence/notifications/intro-notifications/ "+
+			"for the migration path",
+		resourceName,
```
pranav-new-relic, line 43: @pranav-new-relic this is a nice error message to have - we can probably make this a little more compact and towards the end, allow some room to "also" add a link to the getting started guide we have in the Terraform Provider from the point it explains alert policies, workflows, notifications and more, ideally https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/guides/getting_started#add-a-notification-channel

website/docs/r/alert_channel.html.markdown
```diff
@@ -12,6 +12,8 @@ Use this resource to create and manage New Relic alert channels.
...
 
+-> **Note - Not available in the JP region.** This resource is backed by the legacy REST v2 Alert Channels API, which is not provisioned on the New Relic Japan region. If your provider is configured with `region = "JP"`, a `terraform plan` will fail with an actionable error message. For JP configurations, use [`newrelic_notification_destination`](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/notification_destination) and [`newrelic_notification_channel`](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/notification_channel) instead - both are NerdGraph-based and available in every region.
+
```
pranav-new-relic: @pranav-new-relic I don't really like the structuring of this comment - it's a bit too concise (than needed), not that it should be elaborate, but this misses key details.

Firstly, it might be helpful to start the note with "A Note on the Availability of this Resource in Japan (JP)", something similar or a more concise version that still carries the same meaning. 

In the next sentence, while everything is right in its current form, we don't seem to be emphasizing on the fact that the lack of provisioning of RESTv2 in JP is because the API has been deprecated, "which is also why this resource has been deprecated for a very long time now, as explained in the note above as well", plus there is also the need to specify that it is highly essential to move to the NerdGraph-based newer alternatives of these resources (i.e. notifications, destinations and workflows) to avoid interruption of usage when RESTv2 is formally sunsetted.

Finally, towards the end, along with references to the resource specific documentation, let us also add a link to the getting started guide which explains how alert conditions and alert policies can be wired to notification destinations, notification channels and workflows in Terraform as I've explained in a previous comment above.

Please apply the exact same feedback to the note added in the documentation of the `newrelic_alert_policy_channel` resource as well (in fact, in that resource, we only seem to be stating that the lack of its functioning in JP is because `newrelic_alert_channel` does not work in JP due to the aforementioned reasons, but we might want to make clear that neither of these two resources work in JP because of the REST v2 sunsetting - you get my point).

newrelic/data_source_newrelic_account.go
pranav-new-relic: Addressed in `41e6be73`. Shortened the schema `Description` on the `region` computed attribute in `data_source_newrelic_account.go` to a single sentence ("The region code of the account, as reported by New Relic.") and dropped the whole `## Attribute Reference` block for `region` from `website/docs/d/account.html.markdown` - so the attribute is no longer advertised as a stable export in the docs, aligning with the plan to eventually retire it from the data source altogether.

newrelic/helpers_newrelic_region.go
pranav-new-relic: Addressed in `41e6be73`. Tightened the message and swapped the intro-notifications docs link for the getting-started walkthrough. New error body reads: `%s cannot be used in the JP region: it is backed by the legacy REST v2 Alert Channels API, which was globally deprecated in 2024 and is not provisioned on JP. Migrate to the NerdGraph-based \`newrelic_notification_destination\` and \`newrelic_notification_channel\` resources (and \`newrelic_workflow\` in place of \`newrelic_alert_policy_channel\`); see the getting-started guide at https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/guides/getting_started#add-a-notification-channel for a walkthrough`.

website/docs/r/alert_channel.html.markdown
pranav-new-relic: Addressed in `41e6be73`, and applied the same treatment to `newrelic_alert_policy_channel`'s doc. Both notes now:

- open with **"A note on the availability of this resource in Japan (JP)."**;
- call out that the REST v2 Alert Channels API was globally deprecated in 2024 - the same reason both resources have been marked deprecated for a while, referencing the warning above;
- explain that JP simply never provisioned REST v2, so the resource cannot be used there and a `terraform plan` fails early with an actionable error;
- state clearly that moving to the NerdGraph replacements is **not** a JP-only concern - it is highly essential in every region to avoid an interruption when REST v2 is formally sunsetted;
- link to both the resource-specific replacements and the [getting-started guide](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/guides/getting_started#add-a-notification-channel).

For `newrelic_alert_policy_channel` specifically, the note now makes clear that **neither** of the two resources works on JP because of the REST v2 sunsetting - not merely because `alert_policy_channel` depends on `alert_channel`. Mentions `newrelic_workflow` alongside the notification-destination/channel resources as the modern mapping replacement.

nr-developer-toolkit (approved):

PR #3130, opened 2026-07-13
fix: populate id attribute on newrelic_notification_destination data source

### LINKED ISSUE #2968: Error using newrelic_notification_destination for lookup "The argument "destination_id" is required, but no definition was found."

mvanhorn:
# Description

The `id` attribute on the `newrelic_notification_destination` data source was declared `Optional` only (as part of the `ExactlyOneOf{id, name, exact_name}` filter group) and was **not** marked `Computed`. In terraform-plugin-sdk v2, when a data source's `id` attribute is `Optional` without `Computed`, the top-level `id` attribute in state is governed by config, not by `d.SetId()`. So on a `name`/`exact_name` lookup (where `id` is left unset in config) the `id` attribute is written back as `null` even though the internal resource ID (`d.Id()`) is populated via `d.SetId(destination.ID)`.

Consumers that feed the looked-up destination `id` into another resource then hit a Terraform core error, e.g. feeding it into `newrelic_notification_channel.destination_id`:

```
Error: Missing required argument
  destination_id = data.newrelic_notification_destination.alert_destination[each.key].id
The argument "destination_id" is required, but no definition was found.
```

The fix adds `Computed: true` to the `id` schema attribute so the SDK propagates the resource ID to the `id` attribute on every lookup path. This matches how every other data source in this provider surfaces its `id` (e.g. `data_source_newrelic_authentication_domain.go`, `data_source_newrelic_group_management.go`, `data_source_newrelic_user_management.go`), and mirrors the existing `account_id` attribute on this same data source (`Optional: true, Computed: true`).

`ExactlyOneOf` is evaluated against config-set values only, so requiring exactly one of `id`/`name`/`exact_name` in config still holds; a computed `id` does not count toward the group. Behavior for `id`-based lookups is unchanged.

Also adds `TestCheckResourceAttrSet(..., "id")` assertions to the `_ByName` and `_ByExactName` acceptance tests. The existing `testAccNewRelicNotificationDestination` check only asserts `r.Primary.ID` (i.e. `d.Id()`), which was always set; the new assertions cover the `id` *attribute* that Terraform interpolation actually reads.

## Why this matters

Without this, `name`/`exact_name` lookups of a notification destination expose a `null` `id`, which silently breaks any downstream reference to `.id`. Because module instances that sometimes filter by `id` work fine, the failure appears intermittent/random, making it hard to diagnose. This is a one-line schema fix that brings the data source in line with the rest of the provider.

Fixes #2968

## Type of change

## Checklist:

## How to test this change?

The acceptance tests are `TestAccNewRelicNotificationDestinationDataSource_ByName` and `_ByExactName`, which now assert the `id` attribute is set. These require live New Relic credentials and were not run locally; CI runs the acceptance suite. Verified locally: `go build ./...` and `go vet ./newrelic/...` pass, and `gofmt` is clean on the changed files.

- Create a `newrelic_notification_destination` resource.
- Read it back via the `newrelic_notification_destination` data source filtering by `name` (and by `exact_name`).
- Confirm `data.newrelic_notification_destination.<x>.id` resolves to the destination ID (non-empty) and can be passed to `newrelic_notification_channel.destination_id` without a "Missing required argument" error.

Closes #2968


pranav-new-relic:
Hey @mvanhorn - thanks a lot for the fix! We rolled this into our quarterly maintenance sweep of the provider, and since the diff was byte-identical to what you had here, we ported it into #3125 and it shipped as part of [**v3.95.0**](https://github.com/newrelic/terraform-provider-newrelic/releases/tag/v3.95.0) - closing this one as it's already on main. Appreciate the clear write-up on the SDK v2 gotcha! :bow:

PR #3103, opened 2026-06-04
feat(cloud): add GCP Dimensional Metrics integration resources with WIF support

nr-vkollipara:
## Summary

Adds two new Terraform resources for the GCP Dimensional Metrics integration using Workload Identity Federation (WIF) — no long-lived service account keys required:

- **`newrelic_cloud_gcp_dm_link_account`** — Links a GCP project to a New Relic account using a WIF-federated service account.
- **`newrelic_cloud_gcp_dm_integrations`** — Enables individual GCP service integrations (all 34 services) on a linked account.

Three ready-to-use example modules are included:

| Module | IAM approach | Use case |
|--------|-------------|----------|
| `gcp-dimensional-metrics-folder-iam` | Folder-level (4 roles once on folder) | Multiple projects under a shared GCP folder |
| `gcp-dimensional-metrics-project-iam` | Project-level (4 roles × N projects) | Small number of projects, no folder access |
| `gcp-dimensional-metrics-multi-service` | Folder-level + two project groups | Different services per group (analytics vs. compute) |

### `newrelic_cloud_gcp_dm_link_account`

```hcl
resource "newrelic_cloud_gcp_dm_link_account" "example" {
  account_id            = var.newrelic_account_id
  name                  = "my-gcp-project"
  project_id            = "my-gcp-project-id"
  service_account_email = google_service_account.newrelic.email
  audience              = "//iam.googleapis.com/${google_iam_workload_identity_pool_provider.newrelic.name}"
}
```

### `newrelic_cloud_gcp_dm_integrations`

```hcl
resource "newrelic_cloud_gcp_dm_integrations" "example" {
  account_id        = var.newrelic_account_id
  linked_account_id = newrelic_cloud_gcp_dm_link_account.example.id

  big_query { metrics_polling_interval = 60 }  # LP: supports 60 s minimum
  pub_sub   { metrics_polling_interval = 60 }  # LP: supports 60 s minimum
  run       { metrics_polling_interval = 300 }
}
```

## Documentation

- **`website/docs/r/cloud_gcp_dm_link_account.html.markdown`** — new resource reference page
- **`website/docs/r/cloud_gcp_dm_integrations.html.markdown`** — new resource reference page (all 34 services documented, LP services and metrics-only services called out)
- **`website/docs/guides/cloud_integrations_guide.html.markdown`** — fixed GCP DM section:
  - Added `API Gateway` and `Istio` to the services table (were missing)
  - Removed incorrect LP annotation from `Kubernetes Engine`
[...]

PR #3133, opened 2026-07-19
feat(cloud): add GCP Dimensional Metrics resources with WIF support in newrelic_cloud_gcp_link_account

cnivyanthsai:
## Summary

An **optimized alternative to #3103** that responds to the review feedback (avoid two brand-new resources; reduce duplication) **without regressing** the legacy (v1) or Dimensional Metrics (v2) paths.

### Changes vs #3103
1. **Linking merged into the existing `newrelic_cloud_gcp_link_account`.** Setting `audience` + `service_account_email` (each `RequiredWith` the other, `ForceNew`) opts into keyless WIF linking (`cloudAuthenticateIntegration` → link under `gcp_v2`). With both absent, the legacy service-account-key flow is byte-for-byte unchanged. Mode is implicit via field presence — no extra toggle. The separate `newrelic_cloud_gcp_dm_link_account` resource is removed.
2. **Read branches by mode.** WIF uses a minimal `linkedAccount` query (no `integrations` field) to avoid *"Abstract type 'Integration' must resolve to an Object type"* on `gcp_v2` accounts. The legacy path is untouched and only falls back to the minimal query on that specific error (covers a WIF account imported before `audience` is in state).
3. **Client-go reconciliation.** The 7 DM-only services use the consolidated `CloudGcpGenericIntegrationInput` from newrelic/newrelic-client-go#1427 (the per-service input types #3103 referenced were removed upstream).
4. **`newrelic_cloud_gcp_dm_integrations` is table-driven.** A single service-descriptor table drives both schema generation and configure/disable expansion, replacing ~34 repeated `if/else` blocks.
5. **`fetch_tags`/`fetch_table_metrics` removed from the DM resource.** GCP Dimensional Metrics is metrics-only; the backend rejects them (`ERR_INVALID_DATA: … no such parameter(s)`). They defaulted to `false` + `omitempty`, so were silently never sent — which is why they went unnoticed until explicitly set. Surfaced by staging testing. Legacy `newrelic_cloud_gcp_integrations` keeps `fetch_tags` (valid there) and is untouched.

Commits: `0e9886bc` (optimization) + `c73905dd` (fetch_tags removal).

### Why integrations stays a separate resource
Folding the 7 DM services into legacy `newrelic_cloud_gcp_integrations` would regress v1: different schema keys (`fire_base_*`/`fire_store` vs `firebase_*`/`firestore`), a flatten-based Read that throws the abstract-type error on `gcp_v2`, and v2-only services exposed in v1's static schema with no plan-time guard. Keeping it separate also keeps the eventual v1 EOL clean.

### Regression check (v1 / legacy)
- `newrelic_cloud_gcp_integrations` is **byte-identical** to upstream.
- `newrelic_cloud_gcp_link_account` diff is **purely additive** — no legacy line removed/modified; the two new fields are `Optional` (no plan-time drift for existing users).

### Validated on staging (account 11150797)
- WIF authenticate→link, all 34 DM services apply, second-apply no-op (Read correct), `fetch_tags` now rejected at `plan`, teardown unlinks cleanly.

## Dependency
Depends on **newrelic/newrelic-client-go#1427** being released; CI won't compile until then. Verified locally with a `replace` directive against the #1427 branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


newrelic/resource_newrelic_cloud_gcp_dm_integrations.go
```diff
@@ -0,0 +1,575 @@
...
+		Account struct {
+			Cloud struct {
+				LinkedAccount *struct {
+					ID          int `json:"id"`
+					NrAccountID int `json:"nrAccountId"`
+				} `json:"linkedAccount"`
+			} `json:"cloud"`
+		} `json:"account"`
+	} `json:"actor"`
+}
```
pranav-new-relic, line 44: @cnivyanthsai let us **not** have these requests going out directly out of the Terraform Provider. The right precedent is for all kinds of API requests to be routed to APIs from the Terraform Provider via the New Relic Go Client, else this would be meted out of certain functionalities linked to API hits configured in the Go Client (directly using a NerdGraph query with `NerdGraph.QueryWithResponseAndContext` is prohibited in the Terraform Provider).

Now, I understand from this comment that this is intended to be a simple query without the `integrations` field, since the `integrations` field is leading to a few issues (though I'm not sure if we can be 100% sure this is an intended issue and shall not affect customers - please check on that front too), but if this is genuinely required, my suggestion would be to add a `cloud_api_.go` file to the `cloud` package in the Go Client PR (with all of your other changes), add this query and functions to call the query (with a meaningful name) there in `cloud_api_.go`, with the function written similar to all of those which exist in `cloud_api.go`. 

This would ensure that this function you've added continues to remain out of Tutone's scope (since it only modifies `types.go` and `<package_name>_api.go` in a package, but not the other files, so it would not get deleted, and you can continue to make use of the types already defined in `types.go` to unpack the results into respective fields, which you can directly reference in this resource. In this way, the implementation in Terraform would be leaner, you have an established template to use in the Go Client also, and we do not break the general paradigm of "not" directly calling an API from the provider. Please make these changes.

newrelic/resource_newrelic_cloud_gcp_dm_integrations.go
```diff
@@ -0,0 +1,575 @@
...
+			continue
+		}
+		svc.configure(&gcpInput, gcpDmServiceValues{
+			LinkedAccountID:        linkedAccountID,
+			MetricsPollingInterval: getInt(svc.key + ".0.metrics_polling_interval"),
+		})
+	}
+
+	return gcpInput, gcpDisable
+}
```
pranav-new-relic, line 575: @cnivyanthsai I went through this function, and after some time, it was clear to me that this is common to the create and update functions to be able to "add" integrations being added to the config (into the create payload) and delete integrations (add them into the delete payload), which makes sense, but I think this is not evident straightaway, because of a bunch of reasons. I think the variables can be named better (especially things like `svc` which can be explained better, not just here, but across the code), `l`, `dis` and more. Can you please make the variable naming more appropriate to the scenario, structure the function better and add one or two lines of explanatory comments inside the function where we have the create/delete scenarios? I just felt this is key to flag here so anyone finding this code for the first time is able to comprehend this better. Thanks!

website/docs/r/cloud_gcp_link_account.html.markdown
pranav-new-relic, line 4: we might need to replace the space between `new` and `relic` with a hyphen `-` here

newrelic/resource_newrelic_cloud_gcp_link_account.go
```diff
@@ -80,6 +123,85 @@ func resourceNewRelicCloudGcpLinkAccountCreate(ctx context.Context, d *schema.Re
...
+// the explicit use_workload_identity_federation flag and the WIF credential fields: when the
+// flag is set, audience and service_account_email are required; when it is not set,
+// those fields must be absent.
+func resourceNewRelicCloudGcpLinkAccountCustomizeDiff(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
+	isDM := d.Get("use_workload_identity_federation").(bool)
+	audience := d.Get("audience").(string)
+	saEmail := d.Get("service_account_email").(string)
+
+	if isDM && (audience == "" || saEmail == "") {
+		return fmt.Errorf("audience and service_account_email are required when use_workload_identity_federation = true")
```
pranav-new-relic, line 143: Tiny cosmetic change

```suggestion
		return fmt.Errorf("`audience` and `service_account_email` are required when `use_workload_identity_federation = true`")
```

just so this looks better with folks who have a GitHub CI based UX. Kindly edit the error below this also accordingly, and similar errors.

newrelic/resource_newrelic_cloud_gcp_dm_integrations.go
pranav-new-relic: Please also do this with `gcpWIFGetLinkedAccountQuery` and the associated types/functions which I see are defined later in the code.

newrelic/resource_newrelic_cloud_gcp_dm_integrations.go
pranav-new-relic: I also notice such variables in places such as 

```
		la := resp.Actor.Account.Cloud.LinkedAccount
		if la == nil {
			d.SetId("")
			return nil
		}
```

kindly make them more aligned with the code / better explanatory.

newrelic/resource_newrelic_cloud_gcp_dm_integrations.go
```diff
@@ -0,0 +1,575 @@
...
+	linkedAccountID, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Existence-check only — do not request integrations.
+	// Fetching the integrations field causes "Abstract type 'Integration' must resolve
+	// to an Object type" errors on environments where GCP Dimensional Metrics integration types are not
+	// fully registered in the GraphQL schema (e.g. staging). The linked_account_id
+	// and account_id are already in state; this Read simply confirms the account exists.
```
pranav-new-relic, line 451: @cnivyanthsai is this a temporary method you would like to have, or is this going to be the state of the resource going forward too (i.e. by "design", do you not want the integrations to be fetched by read, given an API problem)?

I only wanted to flag this for you awareness, since this would mean any removed/added integrations from the UI or other sources (outside of an already created Terraform resource to manage these integrations) would not account to any drift, which is a key expectation of customers who use a resource via Terraform. If this is going to be an expected bug on Production when this is GA'd, you may want to talk to your PM and let customers know about this in an appropriate fashion in the docs. Just wanted to state this here, as I find quite a few references to such functions which are minimalized versions of what we need to fetch.

website/docs/guides/cloud_integrations_guide.html.markdown
```diff
@@ -225,6 +226,261 @@ Variables:
 * service_account_id: The ID of the New Relic GCP [Service Account](https://cloud.google.com/iam/docs/service-accounts) with [Viewer and Service Usage Consumer roles](https://cloud.google.com/iam/docs/understanding-roles). You can find this ID in the New Relic UI by going to `Infrastructure > GCP > Add a GCP project`. For more information [check out the New Relic docs](https://docs.newrelic.com/docs/infrastructure/google-cloud-platform-integrations/get-started/connect-google-cloud-platform-services-new-relic/).
 * project_id: The ID of the project you want to receive data from in GCP.
 
+<a id="gcp-dimensional-metrics"></a>
```
pranav-new-relic, line 229: I don't think we need this

newrelic/resource_newrelic_cloud_gcp_link_account.go
```diff
@@ -102,6 +224,85 @@ func expandGcpCloudLinkAccountInput(d *schema.ResourceData) cloud.CloudLinkCloud
...
+		},
+		"service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/" +
+			serviceAccountEmail + ":generateAccessToken",
+	}
+	b, err := json.Marshal(cred)
+	if err != nil {
+		return "", err
+	}
+	return string(b), nil
+}
```
pranav-new-relic, line 268: @cnivyanthsai can we please move this into a helper file (you can create a new one for this resource, similar to others you see with other resources in the code, or a `structures_*` file can also work) as these elements are all second-class to functions dealing in the actual Terraform operations in this resource, which is all causing this to be very cluttered. Let's shift such auxiliary functions to a different file. I would also suggest doing the same with the array I see that holds integrations `func gcpDmServices()`, the associated types, etc. Except for other inevitable functions, it is ideal to have the resource files comprise only of CRUD operations - everything else is ideal to be moved into a second helpers or structures file, so this would make the code more readable.

website/docs/guides/cloud_integrations_guide.html.markdown
```diff
@@ -225,6 +226,261 @@ Variables:
...
+### GCP Dimensional Metrics
```
pranav-new-relic, line 230: In this entire block of documentation, I don't seem to find references to the three modules this PR also proposes to add, just wanted to flag this for you guys to see if you'd done this on purpose, or if this was accidental, we might want to accordingly explain the usage of the modules (if they're useful to customers), this is the pattern generally seen with integration explainers in this file (please see the other cloud integrations, OCI, etc and feel free to take a call accordingly)

pranav-new-relic (changes requested):

PR #3145, opened 2026-08-13
feat(pathpoint): add pathpoint resource

asravani-nr:
# Description

Fixes # (issue)

## Type of change

## Checklist:

## How to test this change?

The change introduces addition on new resource pathpoint .
To test please add userId/accountId to the FF : 

https://feature-flags.service.nr-ops.net/flag/edit/Profiling/Pathpoint_PP_Enabled


newrelic/structures_newrelic_pathpoint.go
```diff
@@ -0,0 +1,661 @@
...
+				{
+					"source": s.Related.Source,
+					"target": s.Related.Target,
+				},
+			}
+		}
+		result = append(result, m)
+	}
+	return result
+}
```
pranav-new-relic, line 1174: @asravani-nr Based on your observation so far: does the API we're calling upon read return the stages in the exact same order in which they were created on create? If so, I don't think we should have a problem at all - but if not, this could lead to unintended drift, which I just wanted to call out.

Also, what's worth a manual test here is: apply this resource, and then alter the order of stages in the UI (if that is possible) - then run a `terraform plan`;  we should be able to see a diff that suggests swapping the order of elements in the list of stages (this would confirm our logic in the flatten functions is working perfectly as intended, and aligns with the API behaviour upon create and read).

This observation would also apply to KPIs I believe, as that's a list too (just something to make note of during testing, and with regard to the question I've shared above).

newrelic/resource_newrelic_pathpoint_flow.go
```diff
@@ -0,0 +1,558 @@
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    50,
+				Description: "Ordered list of levels within this stage.",
+				Elem:        levelSchema,
+			},
+		},
+	}
+
+	return &schema.Resource{
```
pranav-new-relic, line 17: @asravani-nr I think we might want to start the definition of the resource in this file from this line - everything above this (pertaining to the schema) can be moved to `structures_newrelic_pathpoint.go` or a new `structures_newrelic_pathpoint_schema.go`, just so the code in this file is readable, and is confined to the "primary" logic of the resource's execution. Can you please move the blocks above this accordingly?

newrelic/resource_newrelic_pathpoint_flow.go
```diff
@@ -0,0 +1,558 @@
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Health rollup strategy: ALERT_CONDITIONS or AUTOMATIC_ROLL_UP.",
+				ValidateFunc: validation.StringInSlice([]string{"ALERT_CONDITIONS", "AUTOMATIC_ROLL_UP"}, false),
+			},
+			"refresh_interval": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "How often health statuses refresh: ONE_MINUTE, FIVE_MINUTES, TEN_MINUTES, FIFTEEN_MINUTES, THIRTY_MINUTES.",
+				ValidateFunc: validation.StringInSlice([]string{"ONE_MINUTE", "FIVE_MINUTES", "TEN_MINUTES", "FIFTEEN_MINUTES", "THIRTY_MINUTES"}, false),
```
pranav-new-relic: @asravani-nr an important note to keep in mind when writing enum based values in `ValidateFunc` or elsewhere in Terraform resources is to ensure that if these values are to be derived from the New Relic Go Client (in all ideal cases they would be), we would want to do it that way, as opposed to hard-coded strings here. For instance, I see `PathPointRefreshIntervalTypes` and `PathPointRefreshInterval` in the companion Go Client PR, so we may want to do a `PathPointRefreshInterval.ONE_MINUTE`, `PathPointRefreshInterval.FIVE_MINUTES` and so on here.

The advantage of doing so is; if these enum values undergo a change in the API at a later point in time and you would need to make that change in the Go Client and Terraform, a change made in the Go Client referenced via Terraform would break the existing code, as the older value would no longer be present in the Go Client's enum. In similar other cases, this approach ensures we're referencing the right values from the client with no typos.

Again, this observation pertains to all/any enum based values derived from the Go Client, let us apply this recommendation to all of such values :)

newrelic/resource_newrelic_pathpoint_flow.go
```diff
@@ -0,0 +1,558 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Optional category to group KPIs.",
+			},
+			"account_id": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Computed:    true,
+				Description: "Account ID this KPI belongs to.",
+			},
```
pranav-new-relic: is it possible that the `account_id` here could be != the `account_id` specified in the root level of the schema?

If not, I guess this can be excluded, and defaulted to `d.Get("account_id")` in the create/update functions

newrelic/resource_newrelic_pathpoint_flow.go
```diff
@@ -0,0 +1,558 @@
...
+	providerConfig := meta.(*ProviderConfig)
+	client := providerConfig.NewClient
+
+	guid := pathpoint.EntityGUID(d.Id())
+
+	// After import, account_id may not yet be in state — decode it from the GUID.
+	accountID := selectAccountID(providerConfig, d)
+	if accountID == 0 {
+		accountID = accountIDFromGUID(d.Id())
+	}
```
pranav-new-relic: very thoughtful :) (to conditionally define this for import driven usecases also)

However, has the working of this condition actually been tested? I think we might have a problem here; since this condition would also fire on read functions and we're selecting from `providerConfig`, it could be that the customer did specify an `account_id` in the config that is not the account ID in the `provider{}` block (which includes `api_key`, `account_id`, `region`) but the apply may have worked because the API key works with both accounts - so now, we're reassigning the account to the one in `providerConfig` and setting it in the state later with a `d.Set()` which could mess things up.

My suggestion would be to explore using a `Importer: &schema.ResourceImporter{` at the schema level, which would allow us define an import specific function - will be similar logic there as well, but more modular in nature to debug if there are issues in future. In this case, you can shift the logic of decoding the account ID from the GUID so it is only import specific, and remove all of this from the read function (the rest of the functions are similar and you might want to see how some of these can be made reusable across both read and import functions).

This is a possibility I'd like you to explore, but if you see any issues and we should reconsider this recommendation, please feel free to let me know.

pranav-new-relic: Also, is the idea of decoding the account ID from the GUID safe? I have used it in code often, but we've had to roll it back sometimes, though I do not fully remember why. Does it work for all kinds of entities (I understand this may be obvious, but any discrepancy in future can lead to a problem - so we might need to be 200% sure about this). Does this require a check with EP?

If not, the safest fallback is to take a composite ID inclusive of the account ID and GUID, which will anyway go into the import specific function and not into read (as per the suggestions above)

newrelic/resource_newrelic_pathpoint_flow.go
```diff
@@ -0,0 +1,558 @@
...
+				Optional:    true,
+				Description: "KPIs tracked at the flow level.",
+				Elem:        kpiSchema,
+			},
+			"stages": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    50,
+				Description: "Ordered list of stages that make up this flow.",
+				Elem:        stageSchema,
```
pranav-new-relic: Some suggestions for better validation:
- we might want to add a `MinItems` too to these list based types so users don't initialise empty blocks of `kpis{}` or `stages{}` for no reason (this happens pretty often in the provider, despite our clear instructions in the docs)
- for any customized validation (e.g. to ensure that every block has certain fields and some specific values only if needed, though I do not see aneed for it here rightaway), checkout usages of `CustomizeDiff` in the repo to understand usages and adapt here

pranav-new-relic (changes requested):

pranav-new-relic:
@asravani-nr these are some very well written changes, I really appreciate the thoughtfully written conditions/patterns (Terraform specific), however I have a few suggestions, please take a look and correct the code accordingly.

In addition, this PR is currently pointing to your other Go Client PR (using `go.mod` `go.sum`) so let us not remove this change from the PR, so as to allow CI checks such as compile, lint etc pass in all future runs. 

Also, please ensure all future commits to this branch are all _conventional_ commits.

Also, please sign the CLA in this comment from @CLAassistant https://github.com/newrelic/terraform-provider-newrelic/pull/3145#issuecomment-5276556939

integration_test_utilities/integration_test_mappings.yaml
```diff
@@ -658,3 +664,9 @@ structures_newrelic_workload.go:
 structures_newrelic_workload_test.go:
   test: true
   product_mapping: WORKLOADS
+structures_newrelic_pathpoint.go:
+  test: false
+  product_mapping: PATHPOINTS
+structures_newrelic_pathpoint_test.go:
+  test: true
+  product_mapping: PATHPOINTS
```
pranav-new-relic: looks like the identifier added to the files is `PATHPOINT` but `PATHPOINTS` is added here, which is why the integration tests check is running 0 tests, let us fix this. Can you please:

- revert the changes in this file
- add a new PATHPOINT product mapping here https://github.com/newrelic/terraform-provider-newrelic/blob/cb4746bd80c68856462ff692e1adc91e6c4cfa07/integration_test_utilities/integration_test_utilities_types.go#L13-L50
- after the above, add a `ProductMappingTypes.PATHPOINT` here https://github.com/newrelic/terraform-provider-newrelic/blob/cb4746bd80c68856462ff692e1adc91e6c4cfa07/integration_test_utilities/integration_test_utilities_types.go#L114-L120 with the regex common to all file names associated with this resources, i.e. `pathpoint`
- run a `make compile`, which should regenerate the test mappings YAML file
- commit and push the test mappings YAML file and this other file we've made changes to


This should get the integration tests CI check find the `PATHPOINT` tests and get them running.

As a side note: if this is **not** logically a part of APM, let us proceed with adding `PATHPOINT`, else if this aligns with `APM`, please add this under `APM`.

pranav-new-relic (changes requested):

integration_test_utilities/integration_test_mappings.yaml
asravani-nr: Updated it ,
We belong to APM team but PathPoint is different Capability all together , 
So I believe adding PathPoint would be accurate.

newrelic/resource_newrelic_pathpoint_flow.go
asravani-nr: Updated this minItems for kpi`s , stages, levels,steps,signals.

newrelic/resource_newrelic_pathpoint_flow.go
asravani-nr: KPI`s can be cross-account. 
Hence there can be a possibility of the account_id in root level and kpi might be different.

newrelic/resource_newrelic_pathpoint_flow.go
asravani-nr: Done

newrelic/structures_newrelic_pathpoint.go
asravani-nr: Yes , the sequence of the stages does matter & the api always return in the same order of how the stages are created.

I hav already validated this and looks good to me , I will have a screen recording created on this scenario and attach the same

newrelic/resource_newrelic_pathpoint_flow.go
asravani-nr: Updated based on the comments ,
& I am pretty sure decoding accountId from GUID is safe , as NGEP-FLOW types always have accountID encoded and even we use the same functionality for update, delete mutations in pathpoint-api-service.

pranav-new-relic:
@asravani-nr: I think the comments have been really well addressed, thanks for thoughtfully digging into each of them and having them corrected (e.g. extending the ask for corrections wherever relevant, e.g. adding enum types in the validate functions from the Go Client, structuring functions better, etc). 

While I have not been able to take a _very_ deep look into very specific functions from a review POV (except for which, I think I did take a detailed look at all the changes proposed by this PR in general), I think everything looks perfectly fine, now that the tests are running and we have some inputs too - but I might have a final set of things to ask for here:

1. I see a lot of separate functions dealing in expansion upon update for KPIs, Stages, etc and there seem to be multiple copies of it for different purposes, we might want to simplify them (make them reusable) if feasible; but if the logic does not permit this, primarily, let us add some comments explaining the purpose of these functions (a brief gist) and if possible, with the comment also explaining how it differs from other "similar" replicas of this function (the difference in usecases). The reason I suggest this is: a newbie on your team picking this up for a modification in future should be able to understand the looping across indices in those functions, what they're exactly meant to do etc. and this can help with future reviews within your team as well. Please have them added, I'll take one final look at those functions and similar ones in the structures file to make sure all is okay.
2. While the integration tests appear to be working, there are, in some rare cases, situations wherein complex resources like these (where the operation relies on quite a lot of nested attributes) can produce a drift that may not be fully tracked by integration tests. To do this, it would be helpful if you can run the following local test exercise, iterating these steps with quite a few key combinations of how we expect customers to use this resource (to clarify again, most of this should have been covered by integration tests, but this is just a double check so we make sure all of our CRUD functions deal in every single attribute without fail, and nothing has been missed out on, accidentally).
    - have a `.tf` file comprise the configuration of this resource, 
    - run a `terraform plan` and `terraform apply` (wait for the resource to have been created)
    - post this, run a `terraform plan` and ensure no drift is shown
    - post this, change some of these attributes one after the other in the UI, repeat the `terraform plan` run to ensure the drift is shown as intended (ensure there is no extra, unintended drift shown, and the change made in the UI does not result in no drift being shown at all)
    - post this, change attributes in the config, run a `terraform plan` and `terraform apply` again (update), and repeat the above exercise
3. If/when the documentation of this resource is run through a PM review or with the Docs team (since you're consulting them with the changes), or when you're editing it as well, please make sure a preview of the content is viewed at https://registry.terraform.io/tools/doc-preview - you will need to paste the contents into this page and view it (as and when needed) so we avoid any broken markdown characters (unrecognized by Terraform's markdown previewer, though it mostly works with all general markdown) propagating into the release, and the contents are visually verified too.
5. The CLA still seems to suggest that you haven't signed it - can you please sign it (or, if you have already, please let me know, so I can find why it still reports so - the CLA check is flakey at times).

asravani-nr:
Hi Pranav ,
1-> Updated the Comments
2-> here is the doc I added test scenarios : https://docs.google.com/document/d/1qQw9NyYJdTbhi9dBrAuVZDr4g0OAWxwmHiSqscCLzr8/edit?tab=t.tlwpyi94dx6j
3-> the documentation is in review, once done will update 
4-> already signed the CLA

pranav-new-relic:
Hi @asravani-nr, for(2) I think the test scenarios have been well tested 👍  and should suffice for us to mark this ready for a merge. However, for (1), I don't seem to find the changes in the last two commits https://github.com/newrelic/terraform-provider-newrelic/pull/3145/commits/11057ad502c1eded2c717f45350950d0298b0b75 or https://github.com/newrelic/terraform-provider-newrelic/pull/3145/commits/eab63105bb69cf2217a4bfdf5fc873b6f4741ea0 - can you please double check and let me know when the changes are in this PR? 
Also, it looks like the lint check is failing because of a formatting issue - please check and fix it. I guess the integration tests would also work after the entitlement is added to the customer ID I've shared with you - so with all of this done, I guess we should be good to proceed whenever planned.

asravani-nr:
> Hi @asravani-nr, for(2) I think the test scenarios have been well tested 👍 and should suffice for us to mark this ready for a merge. However, for (1), I don't seem to find the changes in the last two commits [11057ad](https://github.com/newrelic/terraform-provider-newrelic/commit/11057ad502c1eded2c717f45350950d0298b0b75) or [eab6310](https://github.com/newrelic/terraform-provider-newrelic/commit/eab63105bb69cf2217a4bfdf5fc873b6f4741ea0) - can you please double check and let me know when the changes are in this PR? Also, it looks like the lint check is failing because of a formatting issue - please check and fix it. I guess the integration tests would also work after the entitlement is added to the customer ID I've shared with you - so with all of this done, I guess we should be good to proceed whenever planned.

@pranav-new-relic : 
here is the commit where I added the comments for all the functions : https://github.com/newrelic/terraform-provider-newrelic/pull/3145/changes/11057ad502c1eded2c717f45350950d0298b0b75

newrelic/structures_newrelic_pathpoint.go
```diff
@@ -253,7 +253,7 @@ func pathpointFlowSchema() map[string]*schema.Schema {
 			"health_rollup": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				Computed:     true,
+				Default:      "WORST_STATUS_WINS",
```
pranav-new-relic: similar to the values defined in `ValidateFunc`, please reference these default values from the Go Client; `string(pathpoint.***.***.WORST_STATUS_WINS)`

can you please apply this change to all other default values in the PR too

asravani-nr: > similar to the values defined in `ValidateFunc`, please reference these default values from the Go Client; `string(pathpoint.***.***.WORST_STATUS_WINS)`
> 
> can you please apply this change to all other default values in the PR too

Missed this completely ,Thanks for suggesting.
Updated accordingly.

pranav-new-relic (dismissed):

pranav-new-relic:
for OAC/Virtuoso maintainers: **do not merge**

pranav-new-relic (approved):

Issue #1442, opened 2021-09-23
newrelic_alert_policy_channel "Provider produced inconsistent result after apply" error on Terraform 0.12 with NR 2.25.0


syed-shafiullah:
When applying newrelic_alert_policy_channel resource using Terraform 0.12 I get a terraform error about inconsistency:

Error: Provider produced inconsistent result after apply

When applying changes to
When applying changes to newrelic_alert_policy_channel.pagerduty, provider
"registry.terraform.io/-/newrelic" produced an unexpected new value for was
present, but now absent.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.

baNROne:
There is not enough information to be able to investigate the issue. Please resubmit this issue using the project's [bug report issue template](https://github.com/newrelic/terraform-provider-newrelic/blob/main/.github/ISSUE_TEMPLATE/bug_report.md).

PR #3171, opened 2026-08-20
fix(fleet): paginate configuration lookup by name and flag duplicate matches

### LINKED ISSUE #3: Make resource_newrelic_alert_condition enabled status configurable
_This issue was originally opened by @garthkerr as hashicorp/terraform#11123. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
**v0.8.2**

### Affected Resource(s)
- resource_newrelic_alert_condition

The newrelic resource module by @paultyng is outstanding thus far. However, is would be useful to make the `enabled` status a configurable argument:

https://github.com/hashicorp/terraform/blob/3dfe5a47f1a2cb0df32f086b740bad89c028a9a2/builtin/providers/newrelic/resource_newrelic_alert_condition.go#L184

ShaneLillieNR:
# Description

Depends on [newrelic/newrelic-client-go#1442](https://github.com/newrelic/newrelic-client-go/pull/1442), which fixes `FleetControl.GetEntitySearch`'s cursor parameter — it was declared but never actually wired into the GraphQL query, so pagination silently never advanced. That PR changed `cursor` from `string` to `*string`, and has since merged and shipped in [v2.93.2](https://github.com/newrelic/newrelic-client-go/releases/tag/v2.93.2). `go.mod` now points at that real release.

Pulling that fix in surfaced three things here:

1. **Compile break**: `dsFleetReadByName` in `data_source_newrelic_fleet_configuration.go` called `GetEntitySearch("", ...)`; updated to `GetEntitySearch(nil, ...)`.
2. **Latent correctness gap**: with the client-go bug fixed, it became possible (and necessary) to actually page through results. `dsFleetReadByName` previously only ever inspected page one of `AGENT_CONFIGURATION` entities when resolving a fleet configuration by name — a config sitting on page two+ could never be found. It now follows `NextCursor` across every page via a new `findFleetConfigurationsByName` helper, and if more than one configuration shares the requested name, returns a warning diagnostic listing every matching config ID instead of silently picking one.
3. **Dropped diagnostic, found via live testing**: `dataSourceNewRelicFleetConfigurationRead` discarded the `diags` returned by `dsFleetReadByName` on its success path (`return nil` instead of `return diags`), so the "multiple matches" warning from point 2 never actually reached Terraform. This wasn't caught by the mocked unit tests, which only exercise `findFleetConfigurationsByName` in isolation — it surfaced when running a live `terraform apply` against a real account with two same-named fleet configurations (see [test results comment](https://github.com/newrelic/terraform-provider-newrelic/pull/3171#issuecomment-5359848769) below). Fixed by returning `diags` instead of `nil`.

Fixes # (n/a — discovered while validating newrelic/newrelic-client-go#1442 against this provider)

## Type of change

## Checklist:

- [x] My commit message follows conventional commits
- [x] My code is formatted to Go standards
- [x] New and existing unit tests pass locally with my changes

## How to test this change?

- `go test -tags unit ./newrelic/... -run TestUnitFindFleetConfigurationsByName -v`
- Acceptance: `TestAccNewRelicFleetConfigurationDataSource_ByName` (requires `NEW_RELIC_FLEET_TEST_API_KEY`, `NEW_RELIC_FLEET_TEST_ACCOUNT_ID`, `NEW_RELIC_LICENSE_KEY`)
- **Live-tested** against a real account per [Development & Debugging in the New Relic Terraform Provider](https://newrelic.atlassian.net/wiki/spaces/VIRTUOSO/pages/4316004786): created two `newrelic_fleet_configuration` resources sharing a name, looked them up via the `by_name` data source using the dev-override local binary, confirmed real multi-page pagination (16s read against the account's existing `AGENT_CONFIGURATION` entities) and the duplicate-match warning surfacing correctly after fix #3 above. Full before/after output in [this comment](https://github.com/newrelic/terraform-provider-newrelic/pull/3171#issuecomment-5359848769). Test resources were destroyed afterward — nothing left in the account. (That live test predates the v2.93.2 bump and used the temporary replace-directive commit; the fix itself is unchanged by the version bump.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/newrelic/terraform-provider-newrelic/v3
 
 go 1.26.0
 
+replace github.com/newrelic/newrelic-client-go/v2 => github.com/newrelic/newrelic-client-go/v2 v2.93.1-0.20260810175905-93cb87f7dc94
+
 require (
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.26.1
 	github.com/mitchellh/go-homedir v1.1.0
--- a/go.mod
+++ b/go.mod
@@ -2,13 +2,11 @@ module github.com/newrelic/terraform-provider-newrelic/v3
 
 go 1.26.0
 
-replace github.com/newrelic/newrelic-client-go/v2 => github.com/newrelic/newrelic-client-go/v2 v2.93.1-0.20260810175905-93cb87f7dc94
-
 require (
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.26.1
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/newrelic/go-agent/v3 v3.30.0
-	github.com/newrelic/newrelic-client-go/v2 v2.93.1
+	github.com/newrelic/newrelic-client-go/v2 v2.93.2
 	github.com/stretchr/testify v1.9.0
 	golang.org/x/exp v0.0.0-20240325151524-a685a6edb6d8
 	gopkg.in/yaml.v3 v3.0.1
--- a/go.sum
+++ b/go.sum
@@ -167,8 +167,8 @@ github.com/mitchellh/reflectwalk v1.0.2 h1:G2LzWKi524PWgd3mLHV8Y5k7s6XUvT0Gef6zx
 github.com/mitchellh/reflectwalk v1.0.2/go.mod h1:mSTlrgnPZtwu0c4WaC2kGObEpuNDbx0jmZXqmk4esnw=
 github.com/newrelic/go-agent/v3 v3.30.0 h1:ZXHCT/Cot4iIPwcegCZURuRQOsfmGA6wilW+S3bfBjY=
 github.com/newrelic/go-agent/v3 v3.30.0/go.mod h1:9utrgxlSryNqRrTvII2XBL+0lpofXbqXApvVWPpbzUg=
-github.com/newrelic/newrelic-client-go/v2 v2.93.1 h1:CaIc3K9wFg57dwKCpAo1PaydnXePB069wGTbS5z7Hec=
-github.com/newrelic/newrelic-client-go/v2 v2.93.1/go.mod h1:kROngr/zeNyxbdOI3zPK4Al+ze32XjLEVUDzjxsKW9g=
+github.com/newrelic/newrelic-client-go/v2 v2.93.1-0.20260810175905-93cb87f7dc94 h1:LQlQrXBREfuwIN/dfFsBeLmfJhQg2wYL5+mrBLUEGUo=
+github.com/newrelic/newrelic-client-go/v2 v2.93.1-0.20260810175905-93cb87f7dc94/go.mod h1:kROngr/zeNyxbdOI3zPK4Al+ze32XjLEVUDzjxsKW9g=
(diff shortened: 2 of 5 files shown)


ShaneLillieNR:
## Live test results

Ran a live test against a real NerdGraph account per [Development & Debugging in the New Relic Terraform Provider](https://newrelic.atlassian.net/wiki/spaces/VIRTUOSO/pages/4316004786), using the local dev-override binary compiled with newrelic-client-go pointed at [#1442](https://github.com/newrelic/newrelic-client-go/pull/1442)'s commit.

**Setup**: created two `newrelic_fleet_configuration` resources with the *same* name, then looked one up via `data "newrelic_fleet_configuration"` by that name — exercising both the `GetEntitySearch` cursor signature change and the new pagination/duplicate-match handling in `dsFleetReadByName`.

```
newrelic_fleet_configuration.cursor_fix_test_a: Creating...
newrelic_fleet_configuration.cursor_fix_test_b: Creating...
newrelic_fleet_configuration.cursor_fix_test_b: Creation complete after 2s [id=...fd12]
newrelic_fleet_configuration.cursor_fix_test_a: Creation complete after 3s [id=...fd12]
data.newrelic_fleet_configuration.by_name_live_test: Reading...
data.newrelic_fleet_configuration.by_name_live_test: Still reading... [00m10s elapsed]
data.newrelic_fleet_configuration.by_name_live_test: Read complete after 16s [id=...fd12]

Apply complete! Resources: 2 added, 0 changed, 0 destroyed.
```

The 16s "Still reading..." reflects paging through a real, non-trivial number of `AGENT_CONFIGURATION` entities in the account — confirming the cursor fix actually advances pagination against live NerdGraph, not just in the mocked unit tests.

**Bug caught by this live test**: the first pass showed no warning for the duplicate name, which unit tests (scoped to the pagination helper alone) hadn't caught. Root cause: `dataSourceNewRelicFleetConfigurationRead` discarded the `diags` returned by `dsFleetReadByName` on its success path (`return nil` instead of `return diags`), so the "multiple matches" warning never reached Terraform. Fixed in the latest commit. Re-running against the same still-existing duplicate resources confirmed the warning now surfaces correctly:

```
Warning: found 2 fleet configurations named "tf-fleet-cursor-fix-live-test"; using the first match (...fd12)

  with data.newrelic_fleet_configuration.by_name_live_test,
  on newrelic.tf line 32, in data "newrelic_fleet_configuration" "by_name_live_test":
  32: data "newrelic_fleet_configuration" "by_name_live_test" {

configuration names are not guaranteed to be unique. All matching
configuration IDs:
...fd12,
...fd12.
Use configuration_id instead of name to select a specific configuration
deterministically.
```

Test resources were destroyed afterward (`Destroy complete! Resources: 2 destroyed.`) — nothing left in the account.

This is a good live signal for newrelic/newrelic-client-go#1442: the cursor fix works correctly against a real account with real pagination.

ShaneLillieNR:
newrelic/newrelic-client-go#1442 has merged and shipped in [v2.93.2](https://github.com/newrelic/newrelic-client-go/releases/tag/v2.93.2). Updated `go.mod` to point at that real release and dropped the temporary `replace` directive.

Confirmed clean with no workarounds needed:
- `make compile` — builds clean, no `CompoundCondition` staleness (v2.93.2 has both the cursor fix and the `Description`/`TitleTemplate` fields)
- `go vet -tags unit ./...` — clean
- `go test -tags unit ./newrelic/...` — full suite passes

Marking ready for review.

pranav-new-relic (approved):

PR #3196, opened 2026-09-03
feat(notebooks): add newrelic_notebook resource

pranav-new-relic:
## Summary

Adds `newrelic_notebook` resource and `newrelic_notebook` data source for managing New Relic Notebooks via the Blob Storage API.

**Depends on:** newrelic/newrelic-client-go#1428 (feat/notebooks-api-support). The `go.mod` pins to commit `4645d1de`; will be updated to the release version once that PR merges.

### How it works

Notebook content lives in the Blob Storage REST API (`blob-api.service.newrelic.com/v1/e`). NerdGraph handles the Read path (title, scope metadata). The resource accepts the notebook body as a JSON string — no typed widget schema enforced server-side.

### Content field

The `content_json` field accepts any valid declarative UI JSON string:

```hcl
resource "newrelic_notebook" "example" {
  title        = "APM Health"
  content_json = file("${path.module}/notebook.json")
}
```

JSON is normalized on read (sorted keys, 2-space indent) so key reordering and whitespace never surface as plan diffs.

### Validation

Fires before any API call:
- Envelope: `type = "declarative"`, `version = 1`, `content` is an array
- Widget props: `viz.markdown` must have no widget-level `props`; all other viz require `props.title`
- Error messages use positional path (`content[N].content[N] (viz.X)`) — JSON body not echoed

### What was verified

68 scenarios against production (account 3806526) with a locally-built binary. Covers: full CRUD, import, validation, API-introduced drift detection, data source, all 19 viz types, gap props, mode switch, NGEP indexing (RetryContext, 4s), for_each, file(), unique-title constraint.

Full results: https://source.datanerd.us/gist/d32d4659b2e4f5f4eee462f19c9cbc9e

### Files changed

- `newrelic/resource_newrelic_notebook.go` — schema + CRUD + Import
- `newrelic/structures_newrelic_notebook.go` — normalization, validation, DiffSuppressFunc, helpers
- `newrelic/data_source_newrelic_notebook.go` — data source
- `newrelic/structures_newrelic_notebook_test.go` — unit tests
- `newrelic/resource_newrelic_notebook_test.go` — acceptance tests
- `newrelic/data_source_newrelic_notebook_test.go` — acceptance tests
- `newrelic/provider_newrelic.go` — registration
- `integration_test_utilities/integration_test_mappings.yaml` — registration
- `CHANGELOG.md` — unreleased entry
- `website/docs/r/notebook.html.markdown` — resource docs
- `website/docs/d/notebook.html.markdown` — data source docs

### DIFF
--- a/newrelic/data_source_newrelic_notebook.go
+++ b/newrelic/data_source_newrelic_notebook.go
@@ -1,0 +1,106 @@
+package newrelic
+
+import (
+	"context"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceNewRelicNotebook() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceNewRelicNotebookRead,
+		Schema: map[string]*schema.Schema{
+			"guid": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The unique entity identifier (GUID) of the notebook.",
+			},
+			"fetch_content": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				Description: "When true, the full notebook body is fetched from the Blob Storage API " +
+					"and stored in the `content` attribute. When false (default), only NerdGraph " +
+					"metadata (title, organization_id, blob_id) is retrieved, which is faster " +
+					"and avoids an extra API call.",
+			},
+
+			// Computed attributes populated on every read.
+			"title": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The title of the notebook.",
+			},
+			"organization_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The New Relic organization ID the notebook belongs to.",
+			},
+			"blob_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The blob identifier of the current notebook content.",
+			},
+
+			// Populated only when fetch_content = true.
+			"content": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Description: "The notebook body as a normalized JSON string. Only populated when " +
+					"`fetch_content` is `true`; empty string otherwise.",
+			},
+		},
+	}
+}
+
+func dataSourceNewRelicNotebookRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	providerConfig := meta.(*ProviderConfig)
+	client := providerConfig.NewClient
+
+	guid := d.Get("guid").(string)
+	fetchContent := d.Get("fetch_content").(bool)
+
+	log.Printf("[INFO] Reading New Relic notebook data source: guid=%s fetch_content=%v", guid, fetchContent)
+
+	nb, err := client.Notebooks.GetNotebookWithContext(ctx, guid)
+	if err != nil {
+		if strings.Contains(err.Error(), "not found") {
+			return diag.Errorf("notebook %s not found", guid)
+		}
+		return diag.FromErr(err)
+	}
+
+	d.SetId(guid)
+	_ = d.Set("title", nb.Name)
+	_ = d.Set("blob_id", nb.Content.ID)
+
+	orgID := nb.Scope.ID
+	_ = d.Set("organization_id", orgID)
+
+	if fetchContent {
+		if orgID == "" {
+			// Fall back to provider-level org resolution when scope is absent.
+			var resolveErr error
+			orgID, resolveErr = getOrganizationID(ctx, providerConfig, "")
+			if resolveErr != nil {
+				return diag.FromErr(resolveErr)
+			}
+		}
+		rawContent, contentErr := client.Notebooks.GetNotebookContentWithContext(ctx, orgID, guid)
+		if contentErr != nil {
+			return diag.FromErr(contentErr)
+		}
+		normalized, normErr := normalizeNotebookContent(string(rawContent))
+		if normErr != nil {
+			return diag.FromErr(normErr)
+		}
+		_ = d.Set("content", normalized)
+	} else {
+		_ = d.Set("content", "")
+	}
+
+	return nil
+}
--- a/newrelic/resource_newrelic_notebook.go
+++ b/newrelic/resource_newrelic_notebook.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
 func resourceNewRelicNotebook() *schema.Resource {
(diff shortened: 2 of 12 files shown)


newrelic/data_source_newrelic_notebook.go
```diff
@@ -0,0 +1,107 @@
...
+				Description: "The unique entity identifier (GUID) of the notebook.",
+			},
+			"fetch_content": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				Description: "When true, the full notebook body is fetched from the Blob Storage API " +
+					"and stored in the `content` attribute. When false (default), only NerdGraph " +
+					"metadata (title, organization_id, blob_id) is retrieved, which is faster " +
+					"and avoids an extra API call.",
```
pranav-new-relic, line 27: @pranav-new-relic everywhere such multiline strings have been added as comments both in the resource and in the data source, I don't really see why we have to do that. The description of an attribute of a resource or data source in Terraform only serves auxiliary purposes, such as allowing plugins in IDEs to give suggestions on the appropriate selection and schema, so this isn't going to be written to the UI - you might want to shorten such descriptions only in the schema only to make sure the file isn't too long.

newrelic/resource_newrelic_notebook.go
```diff
@@ -0,0 +1,397 @@
...
+					"appear in NerdGraph within the create timeout. Run terraform plan again " +
+					"to confirm the resource is in the expected state.",
+			})
+		}
+		log.Printf("[DEBUG] Waiting for notebook %s to be indexed in NerdGraph (retry in %s)...", resp.EntityGUID, backoff)
+		time.Sleep(backoff)
+		if backoff < 15*time.Second {
+			backoff *= 2
+		}
+	}
```
pranav-new-relic: @pranav-new-relic I do not believe we require this function, as this was introduced at a point in time when we believed it'd be useful to optimize the create by preventing a read call - however, it is inevitable now given NerdGraph has to give us metadata on the created notebook, which is why we end up creating the notebook. I'd say that it'd be useful to wrap the read function in a RetryContetx based function to prevent issues caused by late indexing on Entity Platform / NG, and that wouild act as a guardrail against flakey entity reads, and it would suffice just the read being called from create. In fact, we might want to even discard this if nothing "new" is coming in from the read, but I would want to still keep this as this would help check if Entity Platform has the notebook listed. We can think through this at a later point as optimization in a fast-follow, but please address this.

newrelic/resource_newrelic_notebook.go
```diff
@@ -0,0 +1,397 @@
...
+
+	titleChanged := d.HasChange("title")
+	contentChanged := d.HasChange("content") || d.HasChange("content_json")
+
+	log.Printf("[INFO] Updating New Relic notebook %s (title_changed=%v, content_changed=%v)", guid, titleChanged, contentChanged)
+
+	if !titleChanged && !contentChanged {
+		return resourceNewRelicNotebookRead(ctx, d, meta)
+	}
```
pranav-new-relic, line 190: if neither the title, nor the content have changed, why do a return of the read function instead of returning nil @pranav-new-relic ? Is there something else we'd like to cover?  If yes, please explain why; else, let us replace this accordingly.

newrelic/resource_newrelic_notebook.go
```diff
@@ -0,0 +1,397 @@
...
+		_ = d.Set("blob_id", freshBlobID)
+	}
+
+	diags := resourceNewRelicNotebookRead(ctx, d, meta)
+
+	// Re-assert the mutation's blob_id in case Read overwrote it with a stale
+	// NerdGraph value. This prevents an unnecessary Blob GET on the next plan.
+	if freshBlobID != "" {
+		_ = d.Set("blob_id", freshBlobID)
+	}
```
pranav-new-relic: I understand why we're doing the set of blob ID again based on the explanation here, but ideally, that would be handled by the read function already, wouldn't it? Please check and curb this redundancy as / if feasible.

newrelic/structures_newrelic_notebook.go
```diff
@@ -0,0 +1,260 @@
...
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetIndent("", "  ")
+	enc.SetEscapeHTML(false)
+	if err = enc.Encode(body); err != nil {
+		return "", nil, fmt.Errorf("could not re-serialize JSON: %w", err)
+	}
+	normalized = string(bytes.TrimRight(buf.Bytes(), "\n"))
+	return normalized, body, nil
+}
```
pranav-new-relic: @pranav-new-relic if you think about it: does it not seem like the function `parseAndNormalizeContent()` is very very close to the function `normalizeNotebookContent()`; isn't it? If you disagree, what is the difference here? Can you try and merge both of these functions and appropriately fix the usages of both of these functions, or if you think they should remain independent and untouched, why do you think that is so? Please take the necessary action here, or justify.

newrelic/resource_newrelic_notebook.go
```diff
@@ -0,0 +1,397 @@
...
+// notebookOrgID returns the organization ID from state if present, or resolves
+// it from provider credentials. Centralises the fallback logic used by Update
+// and Delete.
+func notebookOrgID(ctx context.Context, guid string, d *schema.ResourceData, pc *ProviderConfig) (string, error) {
+	if id, _ := d.Get("organization_id").(string); id != "" {
+		return id, nil
+	}
+	log.Printf("[DEBUG] organization_id not in state for notebook %s, resolving from provider credentials", guid)
+	return getOrganizationID(ctx, pc, "")
+}
```
pranav-new-relic: @pranav-new-relic IMHO functions like these and others in the current file - anything which is not the Create, Read, Update, Delete, Import functions and the schema of the resource (especially auxiliary functions like these) can go to the structures file (ideally a different file, but the helper file is fine for now). Sort the functions based on their relevance there.

pranav-new-relic:
Addressed

@pranav-new-relic in addition to the above two changes, we require redressal on the following concerns too.

1. The code appears to be too bulky in terms of understanding, and is not very well structured. This does not call for editing any of the contents of the code except for feedback, concerns and comments like the ones shared above, but I believe the code can be structured better in terms of variable naming, positioning of functions, following patterns usuall seen with other Terraform resources in the repo so the "code" continues to be bulky, but it is easier for a first time contributor to exactly map the working and flow of logic from one component to another. Kindly visualize the best approach for this and have it fixed.
2. I would like to call for another major behavioral change in the resource - we should probably not publicly disclose the existence of the `content` attribute, i.e. the HCL equivalent of the JSON content supported, and the key reason to this is the inclination of Product towards Declarative UI, which they would like to push for customer consumption for notebooks and dashboards (in future) management, so HCL should be an inevitable detour, but not a preferred option. To do this, I first want you to make sure that we have achieved parity on all changes made so far which cater to `content_json` also to `content` such as validation, etcetera - ensure gaps are covered, find anything that is hard to cover and report back, but once that is done, ensure all public references to `content` is all discarded from the docs. `content_json` shall be supported, HCL examples also can be discarded from the docs, we shall let customers know about this if they have a request at some point, but this isn't needed, given the inclination. In other words, `content` will exist and will fully be functional when a customer does tend to find it themselves and use it (or if we recommend) for which we need to get it fully working, but completely hide the existence of this in the public documentation, or even in this PR for that matter - in the description, etc.
3. The integration test seems to be failing, fix that too. Once done, iterate over all tests, and then give me a good heads up in a brief line or two about the status of mergeability of this PR.

newrelic/data_source_newrelic_notebook.go
pranav-new-relic: Done — all schema descriptions trimmed to single-sentence form across the resource and data source.

newrelic/resource_newrelic_notebook.go
pranav-new-relic: Done — the custom backoff loop is replaced with resource.RetryContext. NGEP indexing confirmed working in 4s on first retry during integration tests.

newrelic/resource_newrelic_notebook.go
pranav-new-relic: Fixed — Update now returns nil immediately when neither title nor content changed.

newrelic/resource_newrelic_notebook.go
pranav-new-relic: Removed — Read sets blob_id from NerdGraph when available. The re-assertion was redundant; removing it causes at most one extra Blob GET during the brief propagation window, which is acceptable. No drift observed across 68 integration tests.

newrelic/structures_newrelic_notebook.go
pranav-new-relic: Merged — normalizeNotebookContent now returns (string, interface{}, error). parseAndNormalizeContent is deleted. All callers that only need the string use _ for the body.

newrelic/resource_newrelic_notebook.go
pranav-new-relic: Done — notebookContentField, notebookOrgID, previousContentRaw, and isNotebookNotFoundError moved to structures_newrelic_notebook.go. The resource file now contains only schema + CRUD + Import.

nr-developer-toolkit (dismissed):

nr-developer-toolkit (approved):

PR #3202, opened 2026-09-07
Changes for entity synthesis resource enrichment

jashraf-hash:
Description
Adds a new opt-in variable, resource_name_enrichment_enabled, to the logs-integration module. When set to true, it passes RESOURCE_NAME_ENRICHMENT_ENABLED=true into the Function App's config, enabling a feature (shipped separately in the newrelic/oci-log-integration function image) that resolves OCIDs in forwarded logs to their OCI resource display name via Resource Search, and injects logging.oci.displayName into the log payload before it's sent to New Relic. Defaults to false, so existing deployments of this module are unaffected unless they explicitly opt in.

Fixes # (issue)

Type of change
 New feature (non-breaking change which adds functionality)
Checklist:
 My commit message follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
 My code is formatted to Go standards (N/A — this change is HCL/Terraform only, no Go code)
 I have performed a self-review of my own code
 I have made corresponding changes to the documentation
 I have added tests that prove my fix is effective or that my feature works (validated via live deployment — see below; no existing test suite covers this example module's .tf files)
 New and existing unit tests pass locally with my changes
How to test this change?
Deploy the logs-integration module with resource_name_enrichment_enabled = true set in the module block, using a Function image that supports the corresponding RESOURCE_NAME_ENRICHMENT_ENABLED env var.
In the OCI Console, check the deployed Function App's Configuration tab and confirm RESOURCE_NAME_ENRICHMENT_ENABLED is present with value true.
Invoke the function with a log payload matching one of the enrichment feature's supported types (e.g. a VCN flow log or DNS Private Resolver log) and confirm logging.oci.displayName appears in the forwarded payload.
Redeploy with the variable omitted (or set to false) and confirm the env var is false and no enrichment behavior occurs — existing forwarding is unaffected.


pbhadra0112:
The PR looks good.Just make sure its merged after https://github.com/newrelic/oci-log-integration/pull/63

pbhadra0112 (approved):

PR #3221, opened 2026-09-16
feat(scorecards-teams): add newrelic_team, newrelic_scorecard, newrelic_scorecard_rule, and NGEP settings resources

pranav-new-relic:
## Summary

Adds Terraform resources for managing New Relic Teams and Scorecards via the NGEP `entityManagement` GraphQL surface. Depends on newrelic/newrelic-client-go#1433.

### New resources

| Resource | Type | Notes |
|---|---|---|
| `newrelic_team` | CRUD | Full team lifecycle — name, description, tags, aliases, managers, entity ownership (non-authoritative), parent hierarchy |
| `newrelic_scorecard` | CRUD | Scorecard with progress levels, tags, rule attachment |
| `newrelic_scorecard_rule` | CRUD | NRQL-based scoring rule with `runInterval` + `enabled` (deprecated `schedule` field intentionally omitted) |
| `newrelic_teams_hierarchy_level` | Import+Update | Rename only — levels are auto-provisioned by NGEP |
| `newrelic_teams_organization_settings` | Import+Update | Singleton org-level discovery settings |

### Design decisions

- **Non-authoritative entity ownership**: `newrelic_team` tracks only statically-declared entity GUIDs. Tag-rule auto-assigned entities are ignored by Terraform to avoid fighting NGEP's discovery engine.
- **System tag preservation**: all update paths merge user-provided tags with the entity's existing `nr.*` system tags before sending to the API — prevents `nr.hierarchy.level`, `nr.team`, and `nr.teamGuid` from being wiped by unrelated updates.
- **`progress_levels` ForceNew**: the NGEP API rejects update mutations on progress levels; `ForceNew` on each inner field triggers destroy+recreate whenever levels change.
- **`patchTeamField`**: raw NerdGraph helper for fields that must be explicitly cleared (description, aliases, parent) — bypasses Go's `omitempty` for deliberate empty-value sends.

### Testing

Comprehensive manual test suite (26 scenarios) run against a live NGEP org:
- Full CRUD for all 5 resources
- Import + idempotency for all resources
- Drift detection via Chrome DevTools UI interaction:
  - Rule `enabled` toggled from UI → plan detects and restores
  - Tags added to scorecard/team from UI → plan detects and removes
  - Externally-added entity (static) → plan detects and removes
  - Tag-rule auto-assigned entity → **no diff** (non-authoritative confirmed)
- `progress_levels` ForceNew: change a color → destroy+recreate confirmed
- Bulk rule attachment: add 3 rules then remove 1 → delta-only API calls
- Description clear (empty string) → API confirms via GraphQL query
- Team → scorecard ownership relationship: works end-to-end

4 bugs found and fixed during testing (see commit history).

### Files changed

- `newrelic/resource_newrelic_team.go` — full CRUD
- `newrelic/resource_newrelic_scorecard.go` — full CRUD
- `newrelic/resource_newrelic_scorecard_rule.go` — full CRUD
- `newrelic/resource_newrelic_teams_hierarchy_level.go` — import+update
- `newrelic/resource_newrelic_teams_organization_settings.go` — import+update
- `newrelic/helpers_newrelic_team.go` — team collection sync, ownership split, user GUID lookup
- `newrelic/helpers_newrelic_ngep.go` — shared NGEP utilities (tag merge, indexing gate, collection pagination)
- `newrelic/structures_newrelic_team.go` — expand/flatten + CustomizeDiff
- `newrelic/structures_newrelic_scorecard.go` — expand/flatten
- `newrelic/structures_newrelic_ngep.go` — shared tag expand/flatten
- `newrelic/provider_newrelic.go` — resource registration
- `website/docs/r/` — 5 new doc files

## Test plan

- [ ] `terraform apply` creates all 5 resources against NGEP org
- [ ] `terraform plan` after apply shows no changes
- [ ] UI drift scenarios: toggle rule enabled, add tag, change description → plan detects each
- [ ] `terraform destroy` removes all 5 resources cleanly
- [ ] `terraform import` + plan shows no changes for each resource type


newrelic/resource_newrelic_scorecard_rule.go
```diff
@@ -0,0 +1,274 @@
...
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "A description of the rule.",
+			},
+			"enabled": {
+				Type:        schema.TypeBool,
+				Required:    true,
+				Description: "Whether this rule is active. Disabled rules are not evaluated.",
```
pranav-new-relic, line 113: @pranav-new-relic I would like you to perform a very trivial check here, though I think it should not be a real concern: can you check if the provider seems to accept values other than true or false against this attribute despite it being defined as `schema.TypeBool` and if so, can you please have adequate validation introduced to prevent this, if this is a real problem you identify upon testing?

newrelic/resource_newrelic_scorecard_rule.go
```diff
@@ -0,0 +1,274 @@
...
+				MaxItems: 1,
+				Description: "NRQL query configuration for this rule. The query must be FACET-ed " +
+					"and use if(latest(...), 1, 0) AS 'score' in the SELECT clause.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"query": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "The NRQL query. Must be FACET-ed and alias the result as 'score'.",
+						},
```
pranav-new-relic, line 130: Since `query` is mandatory (and so is the parent, `nrql_engine`) @pranav-new-relic you might want to introduce validation here similar to the attribute `name` so this is not left as a blank string by the customer, and `terraform plan` returns an appropriate schema validation error with the help of the validation placed.

newrelic/resource_newrelic_scorecard_rule.go
```diff
@@ -0,0 +1,274 @@
...
+						"accounts": {
+							Type:        schema.TypeSet,
+							Required:    true,
+							Description: "Account IDs where this rule runs.",
+							Elem:        &schema.Schema{Type: schema.TypeInt},
+						},
```
pranav-new-relic, line 136: Similar to the comment posted above, I believe there is a good scope of validation here to prevent the set from comprising values of accounts <= 0. Please add a `ValidateFunc` accordingly for both `accounts` and `join_accounts` - both of these attributes here, or add a `CustomizeDiff` if it looks like `ValidateFunc` does not support this usecase yet. You seem to have done something similar with the `impact_weight` attribute below.

newrelic/resource_newrelic_scorecard_rule.go
```diff
@@ -0,0 +1,274 @@
...
+					},
+				},
+			},
+			"impact_weight": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Description:  "Weight of this rule's impact on the overall scorecard score.",
+				ValidateFunc: validation.IntAtLeast(0),
+			},
```
pranav-new-relic, line 154: TODO: Check with the Scorecards team on the valid set of values `impact_weight` should be confined to, update validation on this attribute accordingly after we have an update, and ensure that the range of values is documented in the markdown file describing this resource as well.

newrelic/resource_newrelic_scorecard_rule.go
```diff
@@ -0,0 +1,274 @@
...
+				Description: "Tags in 'key:value1,value2' format.",
+				Elem:        &schema.Schema{Type: schema.TypeString},
+			},
+			"organization_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				ForceNew:    true,
+				Description: "The NGEP organization UUID. Auto-fetched if omitted.",
+			},
```
pranav-new-relic, line 206: @pranav-new-relic I would like to believe that the customer should not be able to manage the `organization_id` by themselves at all, which uproots the entire situation of them being able to change the `organization_id` to something else, and then causing a replacement on the resource owing to a `ForceNew` in the schema. I would like you to review the code of this entire resource including the schema and dependent files again to ensure that all of such references which could be supporting the idea of the customer supplying the `organization_id` are fully eliminated from the code, `organization_id` remains strictly a `Computed` attribute, and is only to be computed via the API call to fetch the ID of the organization that one of the functions is already using in the resource,

newrelic/resource_newrelic_scorecard_rule.go
```diff
@@ -0,0 +1,274 @@
...
+		Enabled: d.Get("enabled").(bool),
+		Scope: scorecards.EntityManagementScopedReferenceInput{
+			ID:   orgID,
+			Type: scorecards.EntityManagementEntityScopeTypes.ORGANIZATION,
+		},
+	}
+	if v, ok := d.GetOk("description"); ok {
+		input.Description = v.(string)
+	}
+	if v, ok := d.GetOk("nrql_engine"); ok {
```
pranav-new-relic: `nrql_engine` is required, so this might not require a `d.GetOk()` as we know it definitely has to be present in the configuration. Just some scope for reduction of code. You might want to apply this logic to mandatory attributes and group them in the first set of checks in the code here, and then move on to the other group, optional attributes - where you will definitely need a `d.GetOk()` like it is present here. Not a required change, just some scope for optimisation in places like these across the code.

newrelic/structures_newrelic_scorecard.go
```diff
@@ -0,0 +1,142 @@
...
+		out = append(out, v.(int))
+	}
+	return out
+}
+
+// expandNRQLEngineCreate maps the nrql_engine block to the Create input type.
+func expandNRQLEngineCreate(raw []interface{}) *scorecards.EntityManagementNRQLRuleEngineCreateInput {
+	if len(raw) == 0 || raw[0] == nil {
+		return nil
+	}
```
pranav-new-relic, line 98: Since `nrql_engine` is `Required` at the schema level, I do not see why we would want to return a nil here. The function should be structured better to return an error if `nrql_engine` is not received from the `d.GetOk()` in the configuration.

newrelic/structures_newrelic_scorecard.go
```diff
@@ -0,0 +1,142 @@
...
+		Query: m["query"].(string),
+	}
+	// accounts is TypeSet — use .(*schema.Set).List() not .([]interface{})
+	if s, ok := m["accounts"].(*schema.Set); ok {
+		engine.Accounts = expandIntListFromInterface(s.List())
+	}
+	// join_accounts is TypeSet — use .(*schema.Set).List() not .([]interface{})
+	if s, ok := m["join_accounts"].(*schema.Set); ok && s.Len() > 0 {
+		engine.JoinAccounts = expandIntListFromInterface(s.List())
+	}
```
pranav-new-relic: I believe that the assertion (type casting) of `join_accounts` and `accounts` can be done directly to a list as opposed to converting it into a set and then into a list. I understand that the rationale behind this could be to first extract unique values in the list of IDs supplied in either array (for which it is converted into a set) and then convert it back into a list, so we only have uniques present, but I think that's not a good use of time at run-time. We might instead want to outsource this logic of having only unique values using a simple set type data structure on `accounts` and `join_accounts` in `nrql_engine` to `CustomizeDiff` at the plan level; in other words, the plan time validation ensure that the `CustomizeDiff` is triggered, which checks if there are no repetitions of account IDs in either array, `accounts` and `join_accounts` and an error is flagged if there is such a situation. Plus, we can also have this plan time validation ensure that there are no common accounts in `accounts` and `join_accounts`, i.e. the intersection of both of these sets (of arrays) is none. With that done, I think we can save the run time by a few milliseconds by removing the interface -> set -> list conversion to just an interface -> list conversion at create time (in read, a reverse construction is anyway inevitable for diff comparison).

newrelic/structures_newrelic_scorecard.go
```diff
@@ -0,0 +1,142 @@
...
+	// accounts is TypeSet — use .(*schema.Set).List() not .([]interface{})
+	if s, ok := m["accounts"].(*schema.Set); ok {
+		engine.Accounts = expandIntListFromInterface(s.List())
+	}
+	if s, ok := m["join_accounts"].(*schema.Set); ok && s.Len() > 0 {
+		engine.JoinAccounts = expandIntListFromInterface(s.List())
+	}
+	return engine
+}
```
pranav-new-relic, line 138: Please apply the feedback I've shared above with `expandNRQLEngineCreate()` with this function `expandNRQLEngineUpdate()` as well, both of the points. Additionally, I do not see why both of these need to exist as individual functions. You can have one common function that returns both input types, which would be mutually exclusive - if the function is called with the "create" mode, you return the create input struct while the other would be empty or atleast structurally empty with no real values, and vice versa in the update mode. You would compute all values in the function prior to deciding which struct they would go into, and when the decision is made based on the argument the function was called with, you can return the right datatypes accordingly and consume them wherever needed. This would be a significant reduction to duplication. I would appreciate any other approaches you could have in mind to reduce duplication too and if you would like to tread that path, but this is one of the ways to do it - my suggestion.

newrelic/structures_newrelic_ngep.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+		key := strings.TrimSpace(parts[0])
+		vals := strings.Split(strings.TrimSpace(parts[1]), ",")
+		for i := range vals {
+			vals[i] = strings.TrimSpace(vals[i])
+		}
+		out = append(out, scorecards.EntityManagementTagInput{Key: key, Values: vals})
+	}
+	return out
+}
```
pranav-new-relic, line 38: @pranav-new-relic I understand that this structure for tags may have been opted to ensure parity with other resources (as comma separated, colon separated key value pairs in an array), but that might not be really good UX - the comma separated listing was inevitable with older resources with `newrelic_entity_tags`, given the then structure of the NerdGraph mutations. We might want to restructure the handling of tags in the scorecard and scorecard rule resources to be an interface which can comprise multiple key-value pairs (as a nested interface) with the key as a singular value for a string, and the values as an array of strings. I would like you to make this change to the schema in both resources, alter the structure wherever applicable in functions like these and more across the resource suite, fix the expand and flatten functions accordingly and fix the documentation too after having tested this. Ensure that this pivot does not affect the drift/sorting and does not introduce similar other repercussions.

newrelic/resource_newrelic_scorecard_rule.go
```diff
@@ -0,0 +1,274 @@
...
+	// to zero/empty) is reflected in state. Guarding with != 0 / != "" would
+	// leave stale non-zero values in state when the server returns 0/"".
+	_ = d.Set("impact_weight", rule.ImpactWeight)
+	_ = d.Set("progress_level", rule.ProgressLevel)
+	_ = d.Set("run_interval", rule.RunInterval)
+
+	_ = d.Set("nrql_engine", flattenNRQLEngine(rule.NRQLEngine))
+
+	return nil
+}
```
pranav-new-relic, line 311: It is nice that we are finding the "ghost" condition, but I'm not sure we're dealing it in the right way. I would think that we should have create/update not call read so we remove this ghost condition, and instead add it in read, similar to the entitysearch we're doing with the teams/scorecard resource to wait until the entity is indexed (which might be different from the ghost condition prevention, but it can still do the job IMHO) and only then succeed. In this case, we don't have to do a read after create/update at all, and the read doesn't have to have this ghost condition - it can just check if the entity exists or not, and if it doesn't, disband it with a `d.SetId("")`.

newrelic/resource_newrelic_scorecard_rule.go
pranav-new-relic: Addressed: `schema.TypeBool` in SDK v2 is validated strictly by the SDK — only `true`/`false`/`1`/`0`/`yes`/`no` are accepted; any other value raises a schema validation error before Create/Update is invoked. Added an explanatory comment in the schema to document this.

newrelic/resource_newrelic_scorecard_rule.go
pranav-new-relic: Addressed: added `ValidateFunc: validation.StringIsNotEmpty` to `query` inside `nrql_engine`. An empty query is now caught at plan time.

newrelic/resource_newrelic_scorecard_rule.go
pranav-new-relic: Addressed: `accounts` and `join_accounts` changed from TypeSet → TypeList. Added `customizeScorecardRuleDiff` validating at plan time: (1) no IDs ≤ 0, (2) no duplicates within accounts, (3) no duplicates within join_accounts, (4) no intersection between the two lists.

newrelic/resource_newrelic_scorecard_rule.go
pranav-new-relic: Acknowledged: added a TODO comment on `impact_weight` noting the valid upper bound needs confirmation from the Scorecards team before the ValidateFunc and documentation are updated.

newrelic/resource_newrelic_scorecard_rule.go
pranav-new-relic: Addressed across all resources: `organization_id` is now strictly `Computed` on all five resources (scorecard_rule, scorecard, team, hierarchy_level, org_settings). `Optional: true` and `ForceNew: true` removed. All Create functions now call `getOrganizationID(ctx, providerConfig, "")` — customers can never supply this field.

newrelic/resource_newrelic_scorecard_rule.go
pranav-new-relic: Addressed: Required attributes (nrql_engine, name, enabled) now use `d.Get()` directly in Create. Optional attributes continue to use `d.GetOk()`. The two groups are clearly separated in the code.

newrelic/structures_newrelic_scorecard.go
pranav-new-relic: Addressed: introduced `extractNRQLEngineParams` shared helper with a comment explaining why the nil path is unreachable (schema validates Required fields before Create is invoked). Both expand functions are now thin wrappers with no silent nil returns.

newrelic/structures_newrelic_scorecard.go
pranav-new-relic: Addressed: changed to TypeList + `customizeScorecardRuleDiff` for plan-time uniqueness/intersection validation. Expand functions now use `.([]interface{})` directly — no set conversion at runtime.

newrelic/structures_newrelic_scorecard.go
pranav-new-relic: Addressed: introduced `nrqlEngineParams` struct and `extractNRQLEngineParams` helper that parses the block once. `expandNRQLEngineCreate` and `expandNRQLEngineUpdate` are now single-line thin wrappers — no duplication.

newrelic/structures_newrelic_ngep.go
pranav-new-relic: Addressed across all three resources: `tags` restructured from flat `key:value` strings to nested blocks with `key` (string) and `values` ([]string). `expandNGEPTags`/`flattenNGEPTags` updated accordingly. Documentation examples updated. Drift detection confirmed working with the new format.

newrelic/resource_newrelic_scorecard_rule.go
pranav-new-relic: Addressed across all three resources: `isNGEPGhostNotFound` removed from all Read functions. Read now calls `d.SetId("")` on not-found (standard pattern). `waitForNGEPEntityIndexed` added to `scorecard_rule` Create (already existed in scorecard and team Create) so Read is only called after the entity is guaranteed visible.

PR #3230, opened 2026-09-17
feat(oci): add forwarder observability metrics tier and tenancy/compartment name attributes

voorepreethi:
## Summary
- Adds a `metrics_tier` variable (`none`/`basic`/`advanced`, default `none`) to the OCI logs-integration module, wired to `FORWARDER_METRICS_TIER`, so the log forwarder function can emit `forwarder.*` self-observability metrics at an opt-in level of detail.
- Adds `TENANCY_NAME` and `COMPARTMENT_NAME` env vars so those metrics can be tagged with human-readable names instead of raw OCIDs. Falls back to the tenancy's own name when deploying into the tenancy root compartment, since OCI's `GetCompartment` API can't resolve the root compartment by OCID (only `GetTenancy` can).
- Corresponding function code lives in `newrelic/oci-log-integration` (not yet merged/published).

## Test plan
- [x] `terraform validate`
- [x] `terraform plan`/`apply`/`destroy` against a real OCI tenancy: confirmed the module computes the expected `FORWARDER_METRICS_TIER`, `TENANCY_NAME`, `COMPARTMENT_NAME` values and can create/destroy a function app+function cleanly
- [x] Root-compartment edge case (`compartment_ocid == tenancy_ocid`): confirmed `COMPARTMENT_NAME` correctly falls back to the tenancy name with no errors
- [x] `metrics_tier` variable validation: confirmed an invalid value (e.g. `"bogus"`) is rejected before touching any infrastructure
- [x] End-to-end: invoked the deployed forwarder function with these env vars set and confirmed correct `compartment_name`/`tenancy_name` attributes on `forwarder.invocations` in NRDB
- [x] Backward compatible: `metrics_tier` defaults to `"none"`, so existing deployments of this module are unaffected unless they opt in


examples/modules/cloud-integrations/oci/logs-integration/locals.tf
```diff
@@ -32,6 +32,12 @@ locals {
   time_out_in_seconds           = 300
   image_url                     = "${var.region}.ocir.io/idptojlonu4e/newrelic-logs-integration/oci-log-forwarder:${var.image_version}"
 
+  # The tenancy's root compartment shares its OCID with the tenancy itself, but OCI's
+  # Identity service can't return it from GetCompartment (data.oci_identity_compartment) --
+  # only GetTenancy can. Detect that case so we read the name from the right data source.
+  is_root_compartment = var.compartment_ocid == var.tenancy_ocid
+  compartment_name    = local.is_root_compartment ? data.oci_identity_tenancy.current_tenancy.name : data.oci_identity_compartment.current_compartment[0].name
```
jashraf-hash, line 39: Wy this condition?If root compartment,then tenancy name otherwise current compartment name

voorepreethi: The root compartment's OCID is identical to the tenancy's OCID in OCI — but the GetCompartment API can't actually resolve it (it's not a fetchable Compartment resource, only GetTenancy can return its name). Without this check, terraform apply would fail for anyone deploying into the tenancy root (a common choice for org-wide logging setups).

is_root_compartment (var.compartment_ocid == var.tenancy_ocid) detects that case:

root compartment → use data.oci_identity_tenancy.current_tenancy.name (via GetTenancy)
regular sub-compartment → use data.oci_identity_compartment.current_compartment[0].name (via GetCompartment, which is only created — count = local.is_root_compartment ? 0 : 1 — when we're not in the root case, so it's never called on an OCID it can't resolve)
This name feeds COMPARTMENT_NAME on the function, so forwarder.* observability metrics get a human-readable tag instead of a raw OCID. Both branches were tested via terraform plan against a real tenancy (normal compartment and root-compartment cases) with no errors.

examples/modules/cloud-integrations/oci/logs-integration/main.tf
```diff
@@ -2,11 +2,14 @@
...
-    "NEW_RELIC_REGION"  = var.new_relic_region
-    "SECRET_OCID"       = var.secret_ocid
-    "CLIENT_TTL"        = local.client_ttl
+    "VAULT_REGION"           = var.region
+    "DEBUG_ENABLED"          = var.debug_enabled
+    "NEW_RELIC_REGION"       = var.new_relic_region
+    "SECRET_OCID"            = var.secret_ocid
+    "CLIENT_TTL"             = local.client_ttl
+    "FORWARDER_METRICS_TIER" = var.metrics_tier
+    "TENANCY_NAME"           = data.oci_identity_tenancy.current_tenancy.name
```
jashraf-hash, line 11: why these last two parameters are needed?

voorepreethi: These are to include in the custom metric payload to include in the Metric schema to query data based on tenancy and compartment

jashraf-hash (approved):

jashraf-hash:
@voorepreethi Unit tests failing.Please check.

